Issue #22, opened 2021-12-01
Investigate RPC proxy caching


valiafetisov:
### Goal
Working RPC proxy cache or a conclusion that it doesn't work for us

### Context
During the architecture discussion, we came up with the basic solution of putting infura provider endpoint behind caching proxy. This way, we can reduce its usage and scale to many simultaneous clients without making a significant change to our code and infrastructure. 

### Assets
Architecture schema and the discussion: https://github.com/sidestream-tech/auction-ui/issues/270#issuecomment-982476998

### Todo
- [x] Investigate whether a typical proxy (eg nginx is capable of doing this)
  - by googling/reading
  - or by trying it out
- [x] Document findings here or in a PR

valiafetisov:
The issue was addressed in the proof-of-concept PR https://github.com/sidestream-tech/unified-auctions-ui/pull/38

PR #38, opened 2022-01-17
RPC cache test

valiafetisov:
Adresses #22 

Validation strategy: 
1. Start frontend using `docker-compose`
2. Open frontend on the `/collateral` page
3. Insure Metamask is not connected
4. Check if auctions are fetched/ no errors are in the console

### DIFF
--- a/core/src/constants/NETWORKS.ts
+++ b/core/src/constants/NETWORKS.ts
@@ -14,7 +14,7 @@ const NETWORKS: Record<string, NetworkConfig> = {
         chainId: '0x2a',
         title: 'Kovan',
         gasPrice: 2000000000,
-        url: `https://kovan.infura.io/v3/${process.env.INFURA_PROJECT_ID}`,
+        url: `/json-rpc/kovan/${process.env.INFURA_PROJECT_ID}`,
         etherscanUrl: 'https://kovan.etherscan.io',
         uniswapV2CalleeDaiAddress: '0x5A40F810754f725DA93e2362775a0600468f7a83',
         uniswapV2LpTokenCalleeDaiAddress: '0xDeC8b9c2829583A89f7F182DEeD7C12112dfAeD0',
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -15,7 +15,7 @@ services:
     frontend:
         restart: unless-stopped
         expose:
-            - 3000
+            - 80
         env_file:
             - './frontend/.env'
         build:
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -21,7 +21,7 @@ services:
         build:
             context: .
             dockerfile: ./frontend/Dockerfile
-            environment:
+            args:
                 - INFURA_PROJECT_ID
                 - DEMO_MODE
                 - PRODUCTION_DOMAIN
--- a/reverse-proxy/production.conf
+++ b/reverse-proxy/production.conf
@@ -2,26 +2,32 @@ events {}
 
 http {
     client_max_body_size 256M;
-    proxy_cache_path /path/to/cache levels=1:2 keys_zone=infura:1m max_size=5g 
-                     inactive=1m use_temp_path=off;
+    # proxy_cache_path /cache levels=1:2 keys_zone=infura:1m max_size=5g 
+    #                  inactive=1m use_temp_path=off;
+
+    # upstream rpc {
+    #     server backend.service.consul;
+    # }
 
     server {
         # Frontend
         location / {
-            proxy_pass http://frontend:3000;
+            proxy_pass http://frontend:80;
         }
 
-        location ~* ^/infura/mainnet/(.*)$ {
-            proxy_pass https://mainnet.infura.io/v3/$1;
+        # RPC endpoint
+        location ~* ^/json-rpc/kovan/(.*)$ {
+            resolver 8.8.8.8;
+            proxy_pass https://kovan.infura.io/v3/$1;
             proxy_http_version 1.1;
             proxy_set_header Upgrade $http_upgrade;
             proxy_set_header Connection 'upgrade';
             proxy_set_header Host $host;
-            proxy_cache infura;
-            proxy_cache_use_stale error timeout http_500 http_502 http_503 http_504;
-            proxy_cache_methods POST;
-            proxy_cache_key "$request_uri|$request_body";
-            add_header X-Cached $upstream_cache_status;
+            # proxy_cache infura;
+            # proxy_cache_use_stale error timeout http_500 http_502 http_503 http_504;
+            # proxy_cache_methods POST;
+            # proxy_cache_key "$request_uri|$request_body";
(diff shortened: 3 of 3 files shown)


valiafetisov:
> So, the requests are `forbidden` for me

Nice! 

> Could it be that localhost needs to be allow-listed again?

For testing purposes only, you can use the key from the twitter bot – it is not restricted by origin. In the end, you may want to provide the key via env variable directly to the proxy (and don't expose it to the frontend)

BracketJohn:
> key from the twitter bot 

which environment? `localhost` still gives me the same problems 😢

valiafetisov:
> So, the requests are forbidden for me, even thouth the key is part of them

If the key is part of the correctly formatted request, then my guess is the same is yours: your actual request origin is not whitelisted. To overcome this problem, you may:
- for the testing purposes: temporarily use the key from any twitter bots, since none of them have restrictions
- if you confirm that caching works, you may:
  - transparently pass `origin` header of the request from the frontend eg `https://main.auction-ui.k8s.sidestream.tech` to the upstream request (that is actually made by proxy, which may skip or overwrite this header in the request)
  - instead of providing unrestricted key to the frontend and then passing it to proxy, you can provide it to the proxy directly

If you question was "is localhost allowed for the infura key stored under /localhost", the answer is yes. You can always validate that by running frontend on the main branch and seeing auctions being fetched

BracketJohn:
Notes

---

So, this works:
```
❯ export KEY=$(chamber read --quiet auction-ui/production.auction-ui.k8s.sidestream.tech/bot-twitter infura_project_id)
❯ curl "https://kovan.infura.io/v3/$KEY" -X POST  -H 'Accept: application/json' -H 'Accept-Language: en-US,en;q=0.5' -H 'Accept-Encoding: gzip, deflate' -H 'Content-Type: application/json' --data-raw '{"id":1,"jsonrpc":"2.0","params":[],"method":"eth_accounts"}'
{"jsonrpc":"2.0","id":1,"result":[]}%
```

BracketJohn:
Proxying and caching works now. To test it:
```
❯ docker-compose up reverse-proxy

# In new terminal
❯ export KEY=$(chamber read --quiet auction-ui/production.auction-ui.k8s.sidestream.tech/bot-twitter infura_project_id)

❯ curl localhost/json-rpc/kovan/$KEY --data-raw '{"id":1,"jsonrpc":"2.0","params":[],"method":"eth_accounts"}' -v
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> POST /json-rpc/kovan/5c293caa1e224323ab257c4123db5b37 HTTP/1.1
> Host: localhost
> User-Agent: curl/7.77.0
> Accept: */*
> Content-Length: 60
> Content-Type: application/x-www-form-urlencoded
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx/1.20.2
< Date: Wed, 26 Jan 2022 16:06:28 GMT
< Content-Type: application/json
< Content-Length: 36
< Connection: keep-alive
< Vary: Accept-Encoding
< Vary: Origin
< X-Cache-Status: MISS
<
* Connection #0 to host localhost left intact
{"jsonrpc":"2.0","id":1,"result":[]}%

❯ curl localhost/json-rpc/kovan/$KEY --data-raw '{"id":1,"jsonrpc":"2.0","params":[],"method":"eth_accounts"}' -v
*   Trying 127.0.0.1:80...
* Connected to localhost (127.0.0.1) port 80 (#0)
> POST /json-rpc/kovan/5c293caa1e224323ab257c4123db5b37 HTTP/1.1
> Host: localhost
> User-Agent: curl/7.77.0
> Accept: */*
> Content-Length: 60
> Content-Type: application/x-www-form-urlencoded
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Server: nginx/1.20.2
< Date: Wed, 26 Jan 2022 16:06:41 GMT
< Content-Type: application/json
< Content-Length: 36
< Connection: keep-alive
< Vary: Accept-Encoding
< Vary: Origin
< X-Cache-Status: HIT
<
* Connection #0 to host localhost left intact
{"jsonrpc":"2.0","id":1,"result":[]}%
```

Note how `X-Cache-Status` changes from `MISS` to `HIT`.

Here's what was required to make it work:
proxying:
  - the regex mapping of the client request to the proxy was not quite right 
  - regex is not required here, as the default behavior of nginx is already to map everything after the location that is matched onto the proxy
  - it may become required based on the actual requirements of this cache
caching:
  -  nginx does not cache content from responses that do not set a `Cache-Control` header, this is super implicitly stated in [their official caching guide](https://www.nginx.com/blog/nginx-caching-guide/#proxy_cache_methods)
  - some stackoverflow threads etc pp pointed me to overwrite this behavior via `proxy_cache_valid`

So now a PoC of proxying + caching exists. If you @valiafetisov can verify that this is the desired behavior (at least in principle), we should go back to #22 and define the exact caching requirements:
- what responses should be cached? (right now all 200er responses)
- for how long should responses be cached? (right now `min(inactive, proxy_cache_valid 200 10m)`
- do we want to allow manual cache invalidation? I.e., a force refresh?

valiafetisov:
Nice. In the end, as suggested above, we shouldn't expose unrestricted key to the frontend. Can we maybe set the key directly as env variable to nginx?

BracketJohn:
- We could use `envsubst` with a custom entrypoint (script) for docker to make this work. 
- We could hit some backend that is part of the application as another level of indirection, there we could securely have the key
- We can switch to kubernetes/ingress based development (hello Tilt!), which allows for more flexible templates/definitions of nginx

valiafetisov:
The docker image seems to actually support `envsubst` out of the box: https://github.com/docker-library/docs/tree/master/nginx#using-environment-variables-in-nginx-configuration-new-in-119

We just need to move config to the `/etc/nginx/templates/production.conf.template`

valiafetisov:
Let's change `10m` to `2m` so it's easier to test and maybe even try to deploy the branch? In the end, the cache should not be longer than ~20s since we just want to avoid duplicate requests from different clients

valiafetisov:
After the discussion today, since proof-of-concept works:
- [ ] let's make it testable
  - [ ] let's make it work with kubernetes
  - [ ] let's make sure that container logs indicate hit/miss
  - [ ] let's create a separate `RPC-caching-test` infura key with no restrictions
- [ ] let's never fetch auctions via metamask in another PR https://github.com/sidestream-tech/unified-auctions-ui/issues/65 (and merge it here first)
- [ ] let's deploy this branch and test it
- [ ] if it goes well, we can solve the problem of not exposing infura key to the frontend

BracketJohn:
Can we move this into new issue(s) and make it part of the next sprint? This reminds me of the open-sourcing todo list we someday defined somewhere and never correctly executed 😢

valiafetisov:
Created the issue https://github.com/sidestream-tech/unified-auctions-ui/issues/66

PR #40, opened 2022-01-18
Dashboard supported column

### LINKED ISSUE #26: Add "supported" indication to the dashboard collateral table
### Goal

Modify `/dashboard` page to include new column which tells which collateral is supported

### Context

Not all collateral types are supported on every test chain. For example, some `UNIV2*` collaterals, although being present in the collateral config are not actually fetched on `kovan` network. Since https://github.com/sidestream-tech/auction-ui/pull/352 is merged this information became dynamic. We want to add a new column with this information. Title: `Supported`, value: :white_check_mark: / :x: 

### Tasks
- [x] add new column to the dashboard table
- [x] use `isCollateralSupported` function to check each collateral
- [x] display values

fegbert:
closes #26

**current state**

`Failed to find` rows skip fetching of `step`, `cut` and `market value`, and these fields stay grayed out.

--- 

Relevant comments from PR in old repo PR:

> I think you would benefit from actually having one loop, instead of two since if collateral is not supported, we don't have to fetch step and cut parameters for it at all, saving us some quota 🙂

Moved everything into one loop in `fetchStepAndCut` and renamed this function to `fetchCoreValues`

 > In this project we use containers to provide data. Please provide it via a prop and update story to provide example data 🙂

Now provided via prop and stories are updated 🙂


frontend/components/CollateralTable.stories.js
```diff
@@ -7,11 +7,12 @@ const collaterals = generateFakeCollaterals();
...
-        collaterals,
+        collaterals: collaterals.collaterals,
+        onChainCollaterals: collaterals.onChain,
     }),
 };
 
 storiesOf('CollateralTable', module).add('Default', () => ({
     ...common,
-    template: '<CollateralTable :collaterals="collaterals" />',
+    template: '<CollateralTable :collaterals="collaterals" :on-chain-collaterals="onChainCollaterals" />',
```
valiafetisov, line 17: Why do you think the component should combine those two things? Couldn't it be done inside the store getter?

valiafetisov:
@BracketJohn why is the push pipeline failing here and why is it triggered? Drone file in this branch doesn't have a step which contains push trigger without a `branch: - main` restriction: https://github.com/sidestream-tech/unified-auctions-ui/blob/c1a3eac0785db78c7c206426882bff27610b843e/.drone.yml

BracketJohn:
As mentioned this morning, I thought that this was just a one PR kind of thing, but now it seems that this happens on every new PR of the unified-auctions. I'll have a look and let you know!

valiafetisov (approved):

florianstoecker (changes requested):
I personally cannot make sense of the "Found"-Column.

Which information does this column give you? 
As I have to ask this question, users will also do. I propose to have an explanatory sentence in the text section above to explain what the column means, or give it a better naming.

valiafetisov:
Ok! Sorry @fegbert:
- Let's call this column `Token` and display token address there (using `<FormatAddress shorten type="address">`) 
- When the address is not found, let's display red `Not found`

florianstoecker:
> Ok! Sorry @fegbert:
> 
> * Let's call this column `Token` and display token address there (using `<FormatAddress shorten type="address">`)
> * When the address is not found, let's display red `Not found`

Can we also add a short explainer sentence to the section above? I propose something like:
"Not all collateral types are supported on every test-chain. An indication for this can be found in the `Token` column"

valiafetisov:
> Not all collateral types are supported on every test-chain

That's the problem [we discussed in the old PR](https://github.com/sidestream-tech/auction-ui/pull/378#pullrequestreview-851700195). Word `supported` implies that ones that are found are "supported", but in fact "supported" is a combination of fields: found token, found clipper and calc contracts, availability of uniswap liquidity pair, etc. So your suggestion will actually create more confusion than clarity.

I think we can take it easy and treat this page as our internal tool, since none of the users actually need to see it, if everything functioning properly.

fegbert:
**updated preview**

frontend/helpers/generateFakeCollateral.js
```diff
@@ -4,13 +4,18 @@ import COLLATERALS from 'auctions-core/src/constants/COLLATERALS';
...
             symbol: collateralObject.symbol,
-            marketUnitPrice: parseFloat(faker.finance.amount()),
-            secondsBetweenPriceDrops: faker.datatype.number(120),
+            marketUnitPrice: new BigNumber(faker.finance.amount()),
+            secondsBetweenPriceDrops: new BigNumber(faker.datatype.number(120)),
             priceDropRatio: new BigNumber(faker.datatype.number({ min: 0.5, max: 1, precision: 0.0001 })),
+            uniswap: {
+                type: 'token',
+                route: faker.finance.ethereumAddress(),
+            },
```
valiafetisov, line 18: Ok, I got the confusion now:
- Token address has nothing to do with uniswap or the route
- Even if it does, you're reusing collateral object structure for something completely different, which is confusing
- I suggest you just create a new key like `tokenAddress` where you store this information
```suggestion
            tokenAddress: faker.finance.ethereumAddress(),
```

frontend/store/collaterals.ts
```diff
@@ -33,13 +34,27 @@ export const mutations = {
...
+    async fetchCoreValues({ commit, rootGetters }: ActionContext<State, State>) {
         const network = rootGetters['network/getMakerNetwork'];
         if (!network) {
             return;
         }
-        dispatch('setup');
         for (const collateral of Object.values(COLLATERALS)) {
+            const tokenAddress = await getTokenAddressByNetworkAndSymbol(network, collateral.symbol).catch(() => {
+                return undefined;
+            });
```
valiafetisov, line 45: The same problem as before, please don't silently catch errors, display them

frontend/store/collaterals.ts
```diff
@@ -56,15 +71,21 @@ export const actions = {
                 marketUnitPrice,
                 secondsBetweenPriceDrops: calcParameters.secondsBetweenPriceDrops,
                 priceDropRatio: calcParameters.priceDropRatio,
+                uniswap: {
+                    type: 'token',
+                    route: tokenAddress,
+                },
```
valiafetisov, line 77: As specified above, please don't reuse the same `uniswap` object as used in the collateral config, that is very confusing

frontend/helpers/generateFakeCollateral.js
```diff
@@ -4,13 +4,18 @@ import COLLATERALS from 'auctions-core/src/constants/COLLATERALS';
 
 export const generateFakeCollaterals = function () {
     const collaterals = [];
```
valiafetisov, line 6: > Please don't forget to update corresponding type

```suggestion
    const collaterals: CollateralRow[] = [];
```

This will help enforcing the correct type

frontend/components/CollateralTable.vue
```diff
@@ -22,7 +22,8 @@
         >
             <FormatCurrency v-if="isValidBigNumber(marketUnitPrice)" :value="marketUnitPrice" currency="DAI" />
             <div v-else>
-                <div v-if="isLoading(record)" class="flex items-center">
+                <div v-if="record.uniswap && !record.uniswap.route" />
```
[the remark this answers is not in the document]
fegbert: I left in the skipping of the loading of the `marketUnitPrice`, `step` and `cut` values if no Token Address has been found like we had when the `Found` column was false.

Removed it now 🙂

frontend/helpers/generateFakeCollateral.js
fegbert: I see. I let myself get confused by the "token" type and thought they were the same thing. Sorry for the confusion 😅

valiafetisov (approved):

PR #42, opened 2022-01-18
Support server-side auth

### LINKED ISSUE #35: Support server-side auth
### Goal
Extend our `core/src/maker` init script to support server-side wallet

### Context
Currently we only support execution of read methods server-side (ie fetching auctions inside `bot-twitter`), but we want to improve it to also execute write methods (such as authorizations or bidding). This require providing a wallet information to dai.js. For an example, see [create-auctions.js#L229](https://github.com/makerdao/auction-demo-keeper/blob/bca09252af029448ff67823b0b7825ad8c6feff8/scripts/create-auctions.js#L229)

### Todo
- [x] Create dev wallet with enough ETH on a test network, add private/public key pair to secrets / bitwarden
- [x] Modify `getMaker` function (or add another one) in order to pass account info to dai.js
- [ ] Test by creating a function that checks auth (and calls it if auth is missing) inside `bot-twitter` on auction start
- [ ] Bonus: add testchain test

benediktkaiser:
closes #35

### DIFF
--- a/core/src/maker.ts
+++ b/core/src/maker.ts
@@ -12,10 +12,10 @@ let globalNetwork: string;
 const createMaker = async function (network: string, privateKey?: string): Promise<typeof Maker> {
     console.info(`creating maker object with "${network}" network`);
     const networkConfig = getNetworkConfigByType(network);
-    const addressOverrides = await fetchContractsAddressesByNetwork(network);
+    const addressOverrides = await fetchContractsAddressesByNetwork(network); // IT GETS SUCK HERE
     globalNetwork = network;
 
-    globalMaker = await Maker.create('http', {
+    const makerOptions = {
         plugins: [
             [McdPlugin, { prefetch: false }],
             [LiquidationPlugin, { vulcanize: false }],
@@ -27,7 +27,6 @@ const createMaker = async function (network: string, privateKey?: string): Promi
             url: networkConfig.url,
             type: 'HTTP',
         },
-        privateKey,
         web3: {
             confirmedBlockCount: 5,
             statusTimerDelay: 24 * 60 * 60 * 1000,
@@ -38,7 +37,21 @@ const createMaker = async function (network: string, privateKey?: string): Promi
         },
         log: false,
         multicall: true,
-    });
+    };
+
+    if (privateKey) {
+        // and then here
+        globalMaker = await Maker.create('http', {
+            ...makerOptions,
+            privateKey: privateKey,
+        });
+        console.info(`created maker object with address: ${globalMaker.currentAddress()}`);
+    } else {
+        globalMaker = await Maker.create('http', {
+            ...makerOptions,
+        });
+    }
+
     if (typeof window !== 'undefined') {
         (window as any).maker = globalMaker;
     }
--- a/core/src/maker.ts
+++ b/core/src/maker.ts
@@ -9,11 +9,12 @@ import { fetchContractsAddressesByNetwork } from './contracts';
 let globalMaker: typeof Maker;
 let globalNetwork: string;
 
-const createMaker = async function (network: string): Promise<typeof Maker> {
+const createMaker = async function (network: string, privateKey?: string): Promise<typeof Maker> {
     console.info(`creating maker object with "${network}" network`);
     const networkConfig = getNetworkConfigByType(network);
     const addressOverrides = await fetchContractsAddressesByNetwork(network);
     globalNetwork = network;
+
     globalMaker = await Maker.create('http', {
         plugins: [
             [McdPlugin, { prefetch: false }],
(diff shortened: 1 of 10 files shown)


benediktkaiser:
I added the private wallet key that we will be using for development purposes in `auction-ui/dev` with the key `PRIVATE_WALLET_KEY`. It can also be found on Bitwarden. 

@florianstoecker will add some ETH funds on the test networks that we can test the server side wallet functions.

valiafetisov:
> I added the private wallet key

Please note that we changed how we organise secrets in the last sprint. Now they are organised in `project-name/host/service` format; in other words, the correct place to which you need to add secret is `auction-ui/localhost/bot-twitter`. I would also recommend to always start your environment variable name with the most important word, in other words name it `WALLET_PRIVATE_KEY` instead. If we would have `WALLET_ADDRESS` in the future (and we may already add it there, since we need this information in the future) they will be conceptually grouped together.

benediktkaiser:
> * [ ]  Please note that we changed how we organise secrets in the last sprint. Now they are organised in `project-name/host/service` format; in other words, the correct place to which you need to add secret is `auction-ui/localhost/bot-twitter`

Thank you I will add it there now!


> * [ ]  I would also recommend to always start your environment variable name with the most important word, in other words name it `WALLET_PRIVATE_KEY` instead. If we would have `WALLET_ADDRESS` in the future (and you should already add it there) they will be conceptually grouped together

Makes sense! I will do that!

benediktkaiser:
I have changed the naming of the env variable! 

I wanted to ask how to best test if the `private_key` is working correctly. I had a look into the tests of DAI.js, but noticed that they have commented out all of their tests that involve a private wallet key. 

https://github.com/makerdao/dai.js/blob/dev/packages/dai/test/Maker.spec.js#L32

@valiafetisov do you have an idea for what kind of function I could write to test the implementation? Until now I have only been able to `console.log` the maker object and it includes the private key, however an actual test to see if the wallet can actually be used would be good to have!

benediktkaiser:
> * [ ]  Please note that we changed how we organise secrets in the last sprint. Now they are organised in `project-name/host/service` format; in other words, the correct place to which you need to add secret is `auction-ui/localhost/bot-twitter`

Also quick note to this: I do not have enough permission to write to these secrets. Who could I ask to give me permissions?

valiafetisov:
> what kind of function I could write to test the implementation?

They issue specifies which function you should call: authorization. More specifically, it suggests to check auth status (ie `getWalletAuthorizationStatus`) and then execute it (ie via `authorizeWallet`).

benediktkaiser:
I have incorporated the authorisation check into the loop, however I am getting no feedback from the function. I also added console.log at multiple breakpoints, but the code is not being executed. 

Do you maybe see a flaw in my approach?

core/src/maker.ts
```diff
@@ -26,6 +27,7 @@ const createMaker = async function (network: string): Promise<typeof Maker> {
             url: networkConfig.url,
             type: 'HTTP',
         },
+        privateKey,
```
valiafetisov, line 30: As I suggested, and you pointed out in the comment above, this may not work and before debugging our code, please validate that passing `privateKey` is actually properly handled by the dai.js library. You may find this in the dai.js docs, in the code itself or in the test. It may be that you also need to pass something else or at least call `maker.useAccountWithAddress(address)`

benediktkaiser: I had assumed as you sent me the following code snippet, but I will look further into it!

benediktkaiser:
Aside from trying to pass the `privateKey` I am having issues executing other functions in the bot in general. 

```js
(await getNewAuctions()).map(notify);
console.log("TEST")
await checkWalletAuthorization();
```

I would expect the following code to result in "TEST" being printed in the console, however this does not happen. This is a test on `bot-twitter/src/index.ts`. So I am also confused why simple console logs are not being printed correctly.

valiafetisov:
Then again, the first thing that you should investigate is: "what happens when I pass `privateKey` to `Maker.create`?".

benediktkaiser:
> Then again, the first thing that you should investigate is: "what happens when I pass `privateKey` to `Maker.create`?".

I found references to the `privateKey` in the Maker Docs as well. Therefore I still assume that this is the right way to do this.

valiafetisov:
> I found references to the privateKey in the Maker Docs as well

Please document them here or in the issue. As documentation can be outdated, you should also have a look at the source code of the dai.js version we're using and also document/outline what's going on there

benediktkaiser:
> > I found references to the privateKey in the Maker Docs as well
> 
> Please document them here or in the issue. As documentation can be outdated, you should also have a look at the source code of the dai.js version we're using and also document/outline what's going on there

I thought I had added the link. (https://docs.makerdao.com/build/dai.js/getting-started#create-a-vault). As edited in the previous comment, when searching through Dai.JS you can see that it is used in their tests and source:

https://github.com/makerdao/dai.js/search?p=1&q=privateKey

benediktkaiser:
As discussed in the daily on the 25.1.22, this PR we no longer be required as we are working on removing DaiJS from our project anyway. Therefore proposing a solution to use server side auth with DaiJS is no longer required.

PR #46, opened 2022-01-18
Use ethers for transactions instead of dai.js

### LINKED ISSUE #44: Remove dai.js dependency for wallet connection, fee estimation, authorisation and execution
### Goal
Do not depend on dai.js for wallet connection, fee estimation, authorisation and execution, while preserving the same functionality.

### Context
Due to the limitations outlined in https://github.com/sidestream-tech/unified-auctions-ui/issues/34#issuecomment-1015225432, we decided to move away from dai.js, or at least try to see how feasible it is for our current code base. Therefore, the plan is to reimplement most of the functionality provided by dai.js inside our `/core` folder. Reference implementation can be found in [`auction-demo-keeper/src/transact.js`](https://github.com/makerdao/auction-demo-keeper/blob/main/src/transact.js) file. Otherwise, reading [dai.js source](https://github.com/makerdao/dai.js/tree/dev/packages/dai/src/eth) should also help.

Note: auction fetching will be reimplemented in another issue.

### Tasks
- [x] Reimplement fee estimation
- [x] Reimplement wallet connection
- [x] Reimplement authorisation and bidding execution, transaction tracking
- [x] Extensively test to validate that existing functionality is not affected by this refactoring

valiafetisov:
Closes #44


valiafetisov:
Benefits that already visible in this PR:
- Reset button is now properly tracked with notifications
- If gas estimation fails (meaning contract emulation doesn't work) the error is cached and not silently logged into the console (this can be tested via bidding on not yet profitable auction with `$nuxt.$store.dispatch('auctions/bid', { id: 'ETH-C:2' })`)

LukSteib (dismissed):
Testing done on `goerli` network
* connect wallet ✅ 
* get fee estimation (similar to prod deployment) ✅ 
* auth / de-auth ✅
* auction execution including transaction tracking ✅ (https://goerli.etherscan.io/tx/0x3ec5e44a7f3993005b682f0fa95777e66805f31f18049c321523a9591cbbac39)
* Transaction tracking for auction restart ✅ 
* gas estimation error cached ✅ 
```
Bidding error Error: Error: cannot estimate gas; transaction may fail or may require manual gas limit
````


Further observations while testing (although not affecting this review)
* auction fetching related error
   * Console log
```
fetch auction error InternalError {message: '[object ProgressEvent]', code: -32603}
```
-> but since related to auction fetching I assume not related to the scope of this PR

* there doesn't seem to be much liquidity for `ETH` in uniswap pool on `goerli`. Auction execution highly affects unsiwap prices and makes testing in the future hard.

File: core/src/tracker.ts [hunk 1]
```diff
@@ -1,69 +1,91 @@
...
+const trackTransaction = async function (
+    transactionPromise: Promise<ethers.ContractTransaction>,
+    notifier?: Notifier,
+    confirmTransaction = true
+): Promise<string> {
     const messageId = uuidv4();
-    const transactionManager = maker.service('transactionManager');
+
+    // Display initial message
     if (notifier) {
```
BracketJohn, line 17: Instead of checking the `notifier` at every point you use it, you could also set it up to always work in the beginning of the function. Looking at the types, something like:
```ts
const notifier = notifier || (messageType, messageContent: MessageContent) => {}
```

should allow you to use the notifier care free throughout this function (:

core/src/tracker.ts
```diff
@@ -1,69 +1,91 @@
...
+                key: messageId,
+                duration: DEFAULT_NOTIFICATION_DURATION,
+            });
+        }
+        throw new Error(error);
+    }
+
+    try {
+        // Wait for mining
+        const minedTransactionReceipt = await transaction.wait(1);
```
BracketJohn, line 49: Is `wait(1)` enough? I remember that sometimes a transaction will not make it into the next block (e.g., if the estimated gas price was off).

frontend/store/fees.ts
```diff
@@ -31,8 +31,12 @@ export const actions = {
         if (!network) {
             return;
         }
-        const fees = await getApproximateTransactionFees(network);
-        commit('setFees', fees);
+        try {
+            const fees = await getApproximateTransactionFees(network);
+            commit('setFees', fees);
+        } catch (error) {
+            console.warn(`Fees calculation error: ${error.message}`);
```
BracketJohn, line 38: nit/q: Why is this just a warning? Aren't the fees quite essential to the system?

core/src/addresses.ts
```diff
@@ -0,0 +1,68 @@
+import type { Contract } from 'ethers';
+import { ethers } from 'ethers';
+import memoizee from 'memoizee';
+import COLLATERALS, { getAllCollateralTypes } from './constants/COLLATERALS';
+import getProvider from './provider';
+import CHAINLOG from './abis/CHAINLOG.json';
+import { CHAINLOG_ADDRESS } from './constants/NETWORKS';
+
+const CHAINLOG_CACHE = 24 * 60 * 60 * 1000;
```
BracketJohn, line 9: > CHAINLOG_CACHE

Maybe instead `CHAINLOG_CACHE_EXPIRY_MS`?

Also, let's make this configurable, it will be helpful to avoid memoization for debugging of deployments (:

core/src/params.ts
```diff
@@ -5,7 +5,7 @@ import BigNumber from './bignumber';
 import getProvider from './provider';
 import MCD_CLIP_CALC from './abis/MCD_CLIP_CALC.json';
 import { RAY_NUMBER_OF_DIGITS } from './constants/UNITS';
-import { fetchContractsAddressesByNetwork, getSupportedCollateralTypes } from './contracts';
+import { fetchContractsAddressesByNetwork, getSupportedCollateralTypes } from './addresses';
 
 const PARAMS_CACHE = 24 * 60 * 60 * 1000;
```
BracketJohn, line 10: as above (also maybe out of scope)

core/src/tracker.ts
```diff
@@ -1,69 +1,91 @@
 import type { Notifier } from './types';
-import getMaker from './maker';
+import { ethers } from 'ethers';
 import { v4 as uuidv4 } from 'uuid';
+import truncateText from './helpers/truncateText';
 
-const trackTransaction = async function (transactionPromise: Promise<any>, notifier?: Notifier): Promise<string> {
-    const maker = await getMaker();
+const DEFAULT_NOTIFICATION_DURATION = 3;
+const NUMBER_OF_BLOCKS_TO_CONFIRM = 5;
```
BracketJohn, line 7: as above - I'm a bit surprised to see that all of these values are hard-coded. I appreciate that the context is a library rather than an app, but still we should offer ways to customize behavior, right?

Another point on this I raised earlier: With environment variables scattered around all modules it becomes very hard to see what can be configured in the application - how about using the `privateRuntimeConfig` that nuxt offers us, see https://nuxtjs.org/docs/configuration-glossary/configuration-runtime-config/#privateruntimeconfig to define these options in a centralized place?

BracketJohn (changes requested):

core/src/tracker.ts
valiafetisov: There are two `.wait()` calls in order to better notify the user: this one is used to tell the user that transaction succeeded, while the next one below is validating 5 blocks on top 🙂

frontend/store/fees.ts
valiafetisov: This is function is called on the interval, populating the store. If it fails ones (and this free endpoints sometimes fail)
- We don't want to show notification to the user since it's not something they initiated or can influence
- If it fails once in a while it doesn't matter: the previous values will stay within the store

core/src/addresses.ts
valiafetisov: This is the file that was only renamed, not sure why github shows it as a completely new one. I suggest to do this change in some other refactoring/renaming PR.

core/src/params.ts
valiafetisov: yes, let's not change it in this PR, since it's unrelated functionality

core/src/tracker.ts
valiafetisov: > I'm a bit surprised to see that all of these values are hard-coded

I think that was a standard for us to set those variables at the top of the file as a constant. In case we actually need to configure it, env variables or $config wouldn't be the best approach, since `/core` is not part of the nuxt. I outlined the approach in a new issue https://github.com/sidestream-tech/unified-auctions-ui/issues/59, please comment there if you think this approach is viable. And let's address it in a separate PR!

BracketJohn: Sounds good to me :+1:

BracketJohn (approved):

PR #47, opened 2022-01-19
Added staging header

### LINKED ISSUE #41: Add staging banner
**Goal:**
Have a permanent banner that is indicating to the user that he is currently using staging environment

**Context:**
In Release v0.5.1 we introduced automated deployments to our staging environment. (staging URL: https://main.auction-ui.k8s.sidestream.tech).
While publishing release notes we discussed whether to share staging URL publicly as it would be a cool feature for the community to check in and see changes already during our sprint. However, finally decided against it due to the fact that a user on staging can also interact with `mainnet` and hence bears the risk of loosing funds. This might be especially a problem if there a un-discovered problems on `staging`. Finally it is still a testing environment. 
In order to mitigate the risk and increase awareness for users being on staging we want to introduce a permanent banner that signals staging environment usage to the user.

**Tasks:**
- [x] Implement a banner to signal staging environment usage
   - [x] color: light red
   - [x] placement: At top of the page
   - [x] text: `This is a staging environment. Please use for testing purposes only.`
   - [x] display banner both in expert mode and explanatory mode
- [x] use frontend env variable to show banner

benediktkaiser:
closes #41 

Screenshot 2022-01-19 at 11 26 39

Screenshot 2022-01-19 at 11 26 21

The only caveat that I found was that due to the Header now being larger, we run into an issue with the SplitLayout view as it needs a defined header hight (which isn't correct anymore with the new header). One way to fix this would be to pass `isStagingEnvironment` all the way to the SplitLayout, however as this is only for the staging, I would argue that this might not be needed. What do you guys think? You can also see the double scrollbar on the second screenshot.


valiafetisov:
> we run into an issue with the SplitLayout view as it needs a defined header hight

I suggest to add this banner as a permanent [antd message](https://antdv.com/components/message). You can basically call a warning message with a 0 duration

benediktkaiser:
> > we run into an issue with the SplitLayout view as it needs a defined header hight
> 
> I suggest to add this banner as a permanent [antd message](https://antdv.com/components/message). You can basically call a warning message with a 0 duration. But it may be annoying 🙂

I would avoid this, as it would mean that all of our other notifications would stack below it, which might be a bit inconvenient. Passing the staging environment is always an option if the two scrollbars really bother us :D

frontend/README.md
```diff
@@ -34,7 +34,8 @@ Help on both things is given in the linked resources above.
...
--   `CONTACT_EMAIL`: (optional) Required in order to display contact link in the footer. This email should be able to accept and manage bug reports and other contact requests.
+- `INFURA_PROJECT_ID`: (required) [infura](https://infura.io/) project id (can be found in: dashboard -> etherium -> create new project -> settings -> keys)
+- `DEMO_MODE`: (optional) When set to true the page will only show a "Coming soon" screen. Can be used for production while the page is not ready yet.
+- `PRODUCTION_DOMAIN`: (optional) Required in order to enable [plausible.io statistics](https://github.com/moritzsternemann/vue-plausible#configuration). In addition to adding it here, the domain (e.g. `auctions.makerdao.network`) should also be registered within [plausible dashboard](https://plausible.io/).
+- `CONTACT_EMAIL`: (optional) Required in order to display contact link in the footer. This email should be able to accept and manage bug reports and other contact requests.
+- `STAGING_ENVIRONMENT`: (optional) When enabled a banner will warn users that they are currently viewing the staging environment, which may be unstable to use.
```
valiafetisov, line 41: This should rather be `IS_STAGING_ENVIRONMENT` or `STAGING_BANNER_SHOWN`

valiafetisov:
I think, two scrollbars on the staging is unacceptable. Therefore, if you want to extend the header heigh, you would need to fix this problem. That's why I propose a different solution 🙂

benediktkaiser:
> I think, two scrollbars on the staging is unacceptable. Therefore, if you want to extend the header heigh, you would need to fix this problem. That's why I propose a different solution 🙂

I implemented a fix for this! 

Screenshot 2022-01-19 at 12 13 11

frontend/containers/AuctionsContainer.vue
```diff
@@ -80,6 +81,9 @@ export default Vue.extend({
         hasAcceptedTerms(): boolean {
             return this.$store.getters['preferences/getAcceptedTerms'];
         },
+        isStagingEnvironment() {
+            return process.env.IS_STAGING_ENVIRONMENT;
+        },
```
valiafetisov, line 86: You don't have to pass it all the way from the container, especially if don't add a story to display behaviour of this prop

frontend/layouts/unified.vue
```diff
@@ -1,5 +1,6 @@
 <template>
     <div :class="isDarkMode && 'dark bg-gray-900'">
+        <StagingHeader v-if="isStagingEnvironment" />
         <Header
```
valiafetisov, line 4: I also suggest you to add it to the Header component instead, so it will be displayed everywhere where there is a header (in the future)

frontend/containers/AuctionsContainer.vue
benediktkaiser: So I can just add this computed property into the MainFlow component directly?

valiafetisov: Yes, I think it's ok

benediktkaiser:
I added it to the Header. I also introduced a new Story for the header to view the header with the alert. I had to introduce some small fixes to take into account the new Header height.

LukSteib (dismissed):

frontend/components/layout/StagingHeader.vue
```diff
@@ -0,0 +1,13 @@
+<template>
```
valiafetisov, line 1: Can you please rename this component to `StagingBanner`? Since placing it near actual `Header` suggests that it's a different version of it

valiafetisov (approved):

valiafetisov:
I've added `IS_STAGING_ENVIRONMENT` to the `auction-ui/main.auction-ui.k8s.sidestream.tech/frontend` secret.

But docker image needs to be updated to actually take it into account:
https://github.com/sidestream-tech/unified-auctions-ui/blob/1ded552d0fa5247351d669ac30d826145b9c232d/frontend/Dockerfile#L38-L44

As well as .drone config that provides the env variables:
https://github.com/sidestream-tech/unified-auctions-ui/blob/1ded552d0fa5247351d669ac30d826145b9c232d/.drone.yml#L16-L22

frontend/nuxt.config.js
```diff
@@ -13,6 +13,7 @@ export default {
         PRODUCTION_DOMAIN: process.env.PRODUCTION_DOMAIN,
         CONTACT_EMAIL: process.env.CONTACT_EMAIL || undefined,
         GITHUB_URL: pkg.repository.url,
+        IS_STAGING_ENVIRONMENT: process.env.IS_STAGING_ENVIRONMENT || false,
```
valiafetisov, line 16: You also probably want to convert this variable, since normal env variables are always comes as strings:
```suggestion
        IS_STAGING_ENVIRONMENT: Boolean(process.env.IS_STAGING_ENVIRONMENT) || false,
```

valiafetisov:
Can you please update docker image and .drone config as suggested in [this comment](https://github.com/sidestream-tech/unified-auctions-ui/pull/47#issuecomment-1021384506)?

valiafetisov (approved):

PR #51, opened 2022-01-19
Time till profitability indicator

### LINKED ISSUE #43: Implement indicator on when auctions turn profitable
**Goal:**
The user is provided with an indicator on when the auction at hand is becoming profitable

**Context:**
We are already calculating and showing time till next price drop and the actual next price (done via `step` `cut` parameters). We also have the current price on Uniswap for the collateral. 
Based on this information we should be able to project when an auction is turning profitable. 
With this issue we want to add a new row to the `transaction` page that gives exactly that indicator.

**Tasks:**
- [x] Include a new row `TimeTillProfitablity` in the tables on auction page and transaction page
   - [x] include new row below existing `Market difference` row
   - [x] naming of row: `Time till profitability`
- [x] Show downwards counting timer (similar to existing `Auction Ends` timer) 
- [x] Implement logic to compute `TimeTillProfitability`
   - [x] use `step` `cut` parameters
   - [x] extrapolate price changes and calculate time till `auction price` < `price on unsiwap` for the first time
   - [x] display computed time in tables

fegbert:
closes #43 

**Auction**

image

**Swap Profit**

image

---

Q: I'm currently hiding the row if the auction is profitable. Is this fine, or do you have another idea? 🙂


LukSteib (commented):
I am not able to fetch auctions when checking out this PR and running `npm run dev`.
Console provides me with the following two errors: 

```
geturl.js?9361:30 POST https://kovan.infura.io/v3/undefined 401
```
-> I would exclude the known infura `origin` problem since `localhost:3000` is specified there. Anything secret (ie. infura key) related?


```
auctions.ts?b334:132 fetch auction error Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.5.0)
    at Logger.makeError (index.js?ef0f:185:1)
    at Logger.throwError (index.js?ef0f:194:1)
    at JsonRpcProvider.eval (json-rpc-provider.js?9d4a:357:1)
```

fegbert:
Have you executed the secret command in the new repo yet?

```sh
secret export auction-ui/dev --format dotenv > ./frontend/.env
```

frontend/helpers/tillDuration.ts
```diff
@@ -25,3 +26,17 @@ export const formatInterval = function (startDate: Date, endDate: Date): string
...
     }
 };
+
+export const timeTillProfitable = function (auction: AuctionTransaction) {
+    let steps = 0;
+    let currentPrice = auction.approximateUnitPrice.toNumber();
+
+    const currentStepProgress = auction.secondsBetweenPriceDrops!! - auction.secondsTillNextPriceDrop!!;
+
+    while (currentPrice > auction.marketUnitPrice!!.toNumber()) {
```
benediktkaiser, line 36: I was just going through your PR and saw that you were calculating the price with a loop. I remember this was also how I initially approached it, until I figured out it could be done in a more effective way with math. Maybe have a look at this file to see how I calculated prices (you could also reuse some of the functions there or maybe move this entire function into that file as it relates to it).

https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/price.ts

valiafetisov:
- [ ] I suggest to rename `Time Till Profitability` to `Estimated Profitability Time` – since it's unclear for the user that it's not a hard number and it may significantly change
- [ ] If profitability date is later than auction endDate, we may show `Likely will not be profitable`

frontend/helpers/tillDuration.ts
```diff
@@ -25,3 +26,20 @@ export const formatInterval = function (startDate: Date, endDate: Date): string
         });
     }
 };
+
+export const calculateTimeTillProfitable = function (auction: AuctionTransaction) {
```
valiafetisov, line 30: - [ ] I suggest to rename this function to `calculateTransactionProfitDate` which returns:
  - `undefined` in case it's not possible to compute
  - `Date` in the future when it will become profitable
  - `Date` in the past if it is already profitable

- [ ] Please move this function to the `core/price.ts` 
- [ ] Please add `transactionProfitDate` to each auction
- [ ] On the frontend, you only need to use this date and don't need to compute anything

frontend/helpers/tillDuration.ts
```diff
@@ -25,3 +26,20 @@ export const formatInterval = function (startDate: Date, endDate: Date): string
...
+        auction.secondsTillNextPriceDrop === undefined ||
+        auction.marketUnitPrice === undefined ||
+        auction.priceDropRatio === undefined
+    ) {
+        return 0;
+    }
+
+    const currentStepProgress = auction.secondsBetweenPriceDrops - auction.secondsTillNextPriceDrop;
+    const stepsTillProfitable = Math.round(
+        auction.approximateUnitPrice.dividedBy(auction.priceDropRatio).minus(auction.marketUnitPrice).toNumber()
```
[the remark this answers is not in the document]
valiafetisov: As suggested in the daily: let's first implement it via a loop and optimise later.

core/src/price.ts
```diff
@@ -49,3 +49,39 @@ export const calculateTransactionProfit = function (auction: Auction): BigNumber
...
+    let steps = 0;
+    let currentValue = auction.approximateUnitPrice;
+    let secondsTillProfitable = 0;
+
+    if (isAlreadyProfitable) {
+        while (currentValue.isLessThan(auction.marketUnitPrice)) {
+            steps -= 1;
+            currentValue = currentValue.multipliedBy(new BigNumber(1).minus(auction.priceDropRatio).plus(1));
+        }
+        secondsTillProfitable = auction.secondsBetweenPriceDrops * steps + auction.secondsTillNextPriceDrop;
```
valiafetisov, line 75: If the auction is already profitable, the profitability indicator is no longer viable information

frontend/components/transaction/SwapTransactionTable.vue
```diff
@@ -40,6 +40,15 @@
                 <span v-else class="opacity-50">Unknown</span>
             </div>
         </div>
+        <div class="flex w-full justify-between">
+            <div>Estimated Profitability Time</div>
+            <div>
+                <template v-if="auctionTransaction.isActive && auctionTransaction.transactionProfitDate">
```
valiafetisov, line 46: Why `v-if` couldn't be the part of the original component? I think it would make sense to encapsulate all related logic in one place

frontend/components/utils/TimeTillProfitable.stories.js
```diff
@@ -0,0 +1,19 @@
...
+        return {
+            auction,
+        };
+    },
+};
+
+storiesOf('Utils/TimeTillProfitable', module).add('Default', () => ({
+    ...common,
+    template: '<TimeTillProfitable :auction="auction" />',
+}));
```
valiafetisov, line 19: Can you please cover all edge cases via stories?
- inactive / finished auctions
- never profitable
- already profitable
- etc

core/src/price.ts
```diff
@@ -49,3 +49,39 @@ export const calculateTransactionProfit = function (auction: Auction): BigNumber
...
+        auction.priceDropRatio === undefined
+    ) {
+        return undefined;
+    }
+
+    const isAlreadyProfitable = auction.approximateUnitPrice.isLessThan(auction.marketUnitPrice);
+    const date = new Date();
+
+    let steps = 0;
+    let currentValue = auction.approximateUnitPrice;
```
valiafetisov, line 67: Since `auction.approximateUnitPrice` is a bignumber and bignumber is an object, [`=` operator doesn't copy the value](https://www.digitalocean.com/community/tutorials/copying-objects-in-javascript), but only copy the reference to that object. Which means modifying `currentValue` will still modify `auction.approximateUnitPrice`.

Big number library actually have a special method for that: https://mikemcl.github.io/bignumber.js/#clone

core/src/price.ts
```diff
@@ -49,3 +49,39 @@ export const calculateTransactionProfit = function (auction: Auction): BigNumber
...
+        while (currentValue.isGreaterThan(auction.marketUnitPrice)) {
+            steps += 1;
+            currentValue = currentValue.multipliedBy(auction.priceDropRatio);
+        }
+        const secondsSinceLastPriceDrop = auction.secondsBetweenPriceDrops - auction.secondsTillNextPriceDrop;
+        secondsTillProfitable = auction.secondsBetweenPriceDrops * steps - secondsSinceLastPriceDrop;
+    }
+
+    date.setSeconds(date.getSeconds() + secondsTillProfitable);
+    return date;
```
valiafetisov, line 86: - please rename `date` to something less generic or just avoid creating this extra variable
- please don't use `setSeconds` since it's actually should accept values [only between 0 and 59](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setSeconds#parameters)
- please use `addSeconds` instead https://date-fns.org/v2.28.0/docs/addSeconds

core/src/price.ts
valiafetisov: Ah, true. I was also wondering how would you solve this proposal 🙂
Maybe you just return some random date in the past instead of undefined or just undefined + handle it in the component.

core/src/price.ts
valiafetisov: Yes, you're right! `new BigNumber` seems to be the correct way. Sorry for the wrong method suggestion

frontend/components/utils/TimeTillProfitable.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+    <div>
+        <div v-if="auction.isActive && auction.transactionProfitDate">
```
valiafetisov, line 3: Let's check `auction.isActive` as well as `isFinished` inside `calculateTransactionProfitDate` instead

core/src/price.ts
valiafetisov: My main suggestion here: we don't have to display exact time if the auction is already profitable and therefore we don't have to calculate it.

frontend/components/utils/TimeTillProfitable.stories.js
valiafetisov: finished auction is not yet covered!

core/src/price.ts
fegbert: > Maybe you just return some random date in the past instead of undefined 😅

Was thinking this too, but I guess this would be very confusing when looking at this function alone 😅

> But yeah, maybe undefined + handling this it in the component would be a better solution

Doing this now 👍

valiafetisov (approved):
Code approved. Waiting for someone to review `TimeTillProfitable` stories (:

LukSteib (approved):

florianstoecker:
> Q: I'm currently hiding the row if the auction is profitable. Is this fine, or do you have another idea? 🙂

I would agree to this point that Felix has stated initially. I think it would not make sense to additionally state that the auction is profitable, let's hide the column once the auction is profitable.
All other things look good to me! :)

LukSteib:
>  I think it would not make sense to additionally state that the auction is profitable, let's hide the column once the auction is profitable.

From UX point of view I am not a fan of hiding and unhiding information based on certain states. Keeping the table row stable and just displaying that the auction is profitable (which should also be reflected at this point by the other figures below that row) to me is just more consistent and provides the user always with the same table no matter which state the auction has. 

From legal perspective of course I am unsure. Will double check but since we are just displaying information and not pushing a user to participate I don't see a problem here

florianstoecker (approved):

PR #52, opened 2022-01-19
Minor UI changes related to feedback

### LINKED ISSUE #39: Adress UI related release feedback
**Goal:**
Adjust UI to improve overall UX and reduce confusion

**Context:**
After our Release `v0.5.1` we received some feedback from the community. 
1) It was asked why certain collateral types on `/dashboard` page show an error -> this is intended and due to the fact that `kovan` network (or `goerli`) were selected and the collateral types there are not supported by uniswap. We show this on hover. We can assume that the user asking the question in the first place did not recognise that hovering over `error` will provide popup with explanation to his question. Let's change that 

2) 
We are in a phase now where there are a decent amount of auctions happening on `mainnet` and it seems that for now rather experienced and auction interested people are using our UI. Hence we decided to switch our default network for all relevant pages from `kovan` to `mainnet` 

**Tasks:**
1)
- [x] extend the error indication in the `/dashboard` table with small `...` button that makes it clear from UX perspective that one can hover to see more information
   - [x] apply same hover effect on this button as currently applied to error 

2) 
- [x] switch default network from `kovan` to `mainnet`

fegbert:
closes #39 

**button**

> extend the error indication in the /dashboard table with small ... button that makes it clear from UX perspective that one can hover to see more information


frontend/components/CollateralTable.vue
```diff
@@ -30,7 +30,8 @@
                     :content="marketUnitPrice"
                     trigger="hover"
                 >
-                    <span class="text-red-500">error</span>
+                    <span class="mr-1 text-red-500">error</span>
+                    <span class="Error">...</span>
```
valiafetisov, line 34: My original suggestion was to show `marketUnitPrice` error text instead of `error` text here, but trimmed, so the user will see beginning of the actual error text (in red) and will realise without any additional signifiers that hovering it will show full text

benediktkaiser: Screenshot 2022-01-26 at 16 27 01

frontend/store/network.ts
```diff
@@ -7,7 +7,7 @@ import {
 } from 'auctions-core/src/constants/NETWORKS';
 import getWallet from '~/lib/wallet';
 
-const DEFAUT_NETWORK = 'kovan';
+const DEFAUT_NETWORK = 'mainnet';
```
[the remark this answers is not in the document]
benediktkaiser: Added it to the secrets!

frontend/nuxt.config.js
```diff
@@ -14,6 +14,7 @@ export default {
         CONTACT_EMAIL: process.env.CONTACT_EMAIL || undefined,
         GITHUB_URL: pkg.repository.url,
         STAGING_BANNER_URL: process.env.STAGING_BANNER_URL || undefined,
+        DEFAULT_ETHEREUM_NETWORK: process.env.DEFAULT_ETHEREUM_NETWORK || 'mainnet',
```
valiafetisov, line 17: Please don't forget to change docker image and .drone build script

frontend/components/CollateralTable.vue
```diff
@@ -50,7 +52,8 @@
                 :content="secondsBetweenPriceDrops"
                 trigger="hover"
             >
-                <span class="text-red-500">error</span>
+                <span class="mr-1 text-red-500">error</span>
+                <span class="Error">...</span>
```
valiafetisov, line 56: Can this also display truncated error message?

valiafetisov: (you may no longer need `Error` class then)

valiafetisov (approved):

PR #60, opened 2022-01-26
Stop computation of new values after auction finishes

### LINKED ISSUE #24: Stop computation of new approximateUnitPrice once auction is finished
**Goal:**
Stop computation of price drop when auction is finished

**Context:**
In the past we have implemented a logic based on the parameters (`step` and `cut`) to compute price drops of an ongoing auction on our own and not rely on fetching the current auction price from the chain. See the following PR where this functionality was implemented: 
Retrieve Step/Cut sidestream-tech/auction-ui#279 
Include Step/Cut in design sidestream-tech/auction-ui#289

We figured that we keep on updating the Auction Price based on the `step` `cut` logic even after the auction is finished. This might cause confusion for user looking at a finished auction and seeing an auction price that is not similar to the actual price the auction was taken. 
See discrepancy in screenshot (our UI) and the respective finish price on makerburn for that exact auction (below)

**Tasks:**
- [ ] Quickly make yourself familiar with the `step` `cut` based logic to update price of an auction (see linked PRs above)
- [ ] Ensure that we do not update prices of auctions based on this logic once an auction is finished

**Assets:**
Screenshot of our UI
Screenshot 2022-01-06 at 12 46 33

Makerburn auction stats:
https://makerburn.com/#/liquidations/eth/271

benediktkaiser:
closes #24

### DIFF
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -42,7 +42,7 @@ const enrichAuctionWithActualNumbers = async function (
     network?: string
 ): Promise<Auction> {
     const maker = await getMaker(network);
-    if (!auction.isActive) {
+    if (!auction.isActive || auction.isFinished) {
         return {
             ...auction,
             unitPrice: new BigNumber(0),
@@ -65,7 +65,7 @@ const enrichAuctionWithActualNumbers = async function (
 };
 
 const enrichAuctionWithMarketValues = async function (auction: Auction, network: string): Promise<Auction> {
-    if (!auction.isActive || !auction.approximateUnitPrice) {
+    if (!auction.isActive || !auction.approximateUnitPrice || auction.isFinished) {
         return auction;
     }
     try {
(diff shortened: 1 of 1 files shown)


valiafetisov (approved):

florianstoecker (approved):
Looks good to me! The prices do not drop further after I've executed two auctions.

benediktkaiser:
> One question to this approach: Do we really take the last auction price somewhere from the chain, or do we take the last one computed by ourselves? Because I imagine there is this edge case when someone takes the auction close to the price drop computation. So while the auction is being taken & in the state of confirming, the computation won't stop if we take our computed last price.

We calculate the new price ourselves every second and every 30 seconds it is re-pulled from the chain. However our calculations should always be exactly what the chain also has, as we execute the same math as on the chain. 

I don't know if this answered your question :D

valiafetisov:
> Do we really take the last auction price somewhere from the chain, or do we take the last one computed by ourselves?

At the moment we don't fetch data about finished auctions, so we just display the latest data we have. This will be improved in the future 🙂

PR #63, opened 2022-01-26
Improved Currency rounding

### LINKED ISSUE #58: Ensure to show one non zero decimal place
**Context:**
During the market downturn we could see a lot of liquidations happening on mainnet for different collateral types for the first time. 
We figured (unfortunately not taken a screenshot) that for one of the LP tokens (`UniV2*`) the `Auction amount` displayed was `0.00`. Nevertheless it was a valid auctions with opportunity to make profit. 
My suspicion: The number was very low and we only show 2 decimal places. Let's have a look into that and apply a solution to this edge case.

**Tasks:**
- [ ] Come up with solution to edge case outlined above (e.g. in case of `0.00` auction amount increasing the number of decimal places displayed up to first non 0 digit like for example `0.00003`)

benediktkaiser:
closes #58 

This PR extends our `FormatCurrency` component to take into account when very small numbers are used. 
We used to always cut of after the second decimal point which often made small numbers be displayed as 0.

`0.000454786546754578645678546874546789 DAI` will be displayed as `0.00045 DAI`
`0.0000000000001 DAI` will be displayed as `1e-13 DAI`
_(Stories exist to test both of these numbers)_

This is to ensure:

A.) We can display even very small sums
B.) It does not break our UI


LukSteib (changes requested):
- [ ] In storybook the changes you applied are not reflected (see screenshot below)

image

benediktkaiser:
This actually broken when I merged main into this PR, so I will fix it later!

LukSteib (dismissed):

frontend/components/utils/FormatCurrency.vue
```diff
@@ -34,6 +36,19 @@ export default Vue.extend({
         },
     },
     computed: {
+        formattedNumber(): number {
+            if (!this.value.toFixed()) {
+                return 0;
+            }
```
valiafetisov, line 42: Why would you want to return unformatted `0` from the `formattedNumber`?

frontend/components/utils/FormatCurrency.vue
```diff
@@ -1,10 +1,12 @@
...
-            v-else-if="value"
-            >NaN</span
-        ><span class="uppercase"> {{ currency }}</span>
-    </span>
+    <span
+        >{{ sign
+        }}<span v-if="isValidNumber"
+            ><animated-number v-if="value > 0.01" :value="value" :decimal-places="decimals" /><span v-else>{{
+                formattedNumber
+            }}</span></span
```
valiafetisov, line 7: Why would you want to disable animation and reimplement formatting if you can just properly determine number of decimals. Not even here, but inside `AnimatedNumber` itself?

frontend/components/utils/FormatCurrency.vue
```diff
@@ -34,6 +36,19 @@ export default Vue.extend({
         },
     },
     computed: {
+        formattedNumber(): number {
```
valiafetisov, line 39: If this is called `formattedNumber`, it should return string, not number, otherwise you can end up displaying some very long number

frontend/components/utils/FormatCurrency.vue
```diff
@@ -34,6 +36,19 @@ export default Vue.extend({
...
+            if (!this.value.toFixed()) {
+                return 0;
+            }
+
+            if (this.value < 0.01) {
+                const decimalCutOff = Math.abs(Math.floor(Math.log10(Number(this.value)))) + 1;
+                return parseFloat(
+                    parseFloat(this.value.toString()).toFixed(decimalCutOff === Infinity ? 0 : decimalCutOff)
+                );
```
valiafetisov, line 48: As stated above and in general, please avoid numerous conversions back and forth, the number/bignumber should ideally be treated as a bignumber and converted to a string only once right before we need to display it

frontend/components/utils/FormatCurrency.stories.js
```diff
@@ -42,4 +42,12 @@ storiesOf('Utils/FormatCurrency', module)
...
         ...common,
         template: '<FormatCurrency currency="dai" />',
+    }))
+    .add('Number under 0.01', () => ({
+        ...common,
+        template: '<FormatCurrency :value="0.000454786546754578645678546874546789" currency="dai" />',
+    }))
+    .add('Number under 0.000001', () => ({
+        ...common,
+        template: '<FormatCurrency :value="0.0000000000001" currency="dai" />',
```
valiafetisov, line 52: Please provide big number input as well

frontend/components/utils/FormatCurrency.vue
benediktkaiser: Because I do not just change the decimal point. When the number becomes too small it gets formatted like this: 1e-13 DAI. 

This kind of formatting sadly does not work with the Animated Number library. 

However I can do it for numbers that do not get formatted in this special way.

frontend/components/utils/FormatCurrency.vue
benediktkaiser: Whats a good name I could use otherwise?

frontend/components/utils/FormatCurrency.vue
benediktkaiser: Correct me If I am wrong, but as I am using normal Javascript functions to calculate this, I don't see any way around. Especially one like above where I determine the number of decimal points I need:
```js
 const decimalCutOff = Math.abs(Math.floor(Math.log10(Number(this.value)))) + 1;
```
I don't see any way to do this with BigNumbers. 

As for the line you highlighted: I only found this setting in the [BigNumber config ](https://mikemcl.github.io/bignumber.js/#exponential-at) where I can change this rounding. However I don't think its smart to change the config of BigNumber, which is why I went with traditional javascript methods.

frontend/components/utils/FormatCurrency.stories.js
benediktkaiser: I added two stories to show how it works with BigNumbers.

frontend/components/utils/FormatCurrency.vue
benediktkaiser: I made it return `this.value` instead of 0.

frontend/components/utils/FormatCurrency.vue
valiafetisov: > When the number becomes too small it gets formatted like this: 1e-13 DAI

We should not display those kind of numbers, as we can not expect from users to understand mathematical notation. Instead, we should just change the number of decimal points or even better: show something like `less than 0.01 DAI`

frontend/components/utils/FormatCurrency.vue
valiafetisov: Sorry that it's not clear, I implied not to rename the function, but to always return string

frontend/components/utils/FormatCurrency.vue
valiafetisov: See my comment above to why you may not even need it

frontend/components/utils/FormatCurrency.vue
benediktkaiser: Do you think 0.01 is a good cut of point for this? I would maybe display the decimal places until `0.00001` or maybe even later.

frontend/components/utils/FormatCurrency.vue
```diff
@@ -1,10 +1,13 @@
...
-    <span>
-        {{ sign }}<animated-number v-if="isValidNumber" :value="value" :decimal-places="decimals" /><span
-            v-else-if="value"
-            >NaN</span
-        ><span class="uppercase"> {{ currency }}</span>
-    </span>
+    <span
+        >{{ sign
+        }}<span v-if="isValidNumber"
+            ><animated-number v-if="value > 0.00001 || value === 0" :value="value" :decimal-places="decimals" /><span
```
valiafetisov, line 5: Can you please move `value > 0.00001 || value === 0` into a computed property and reuse it in the `decimals` (to not compute log and to avoid checking against infinity)? Otherwise, you can reuse `decimals` here as `decimals > 5` instead of creating a separate logic for the same thing

valiafetisov: We may also want to check `if (number > 0)` before counting decimals and applying log10

valiafetisov: One last thing: we may want to do it inside the animated number component (taking into account decimals prop as minimum), so this logic is applied everywhere

benediktkaiser: I understand moving `value > 0.00001 || value === 0` into a computed property, but I do not understand why I should use it in the `decimals` calculation to replace the check against Infinity.

valiafetisov: Then, let's do it the other way around and use `decimals` inside the `isValidNumber`?

valiafetisov:
> One last thing: we may want to do it inside the animated number component (taking into account decimals prop as minimum), so this logic is applied everywhere

- [ ] as `animated-number` is also used in the `FormatMarketValue`, we should better compute it inside `AnimatedNumber`
- [ ] make sure that  `{{ sign }}` works well with `under` and create a story for it

benediktkaiser:
> * [ ]  make sure that  `{{ sign }}` works well with `under` and create a story for it

I already have two stories for this. (One with big Number and one with a normal one See `Big Number under 0.000001` and `Number under 0.000001`)

frontend/components/utils/FormatCurrency.vue
benediktkaiser: I got to be honest. I still don't see the point in this. I would be up for a quick call if that makes the explanation easier.

benediktkaiser:
I moved the computation to `animated-number`. I also added new stories for `FormatMarketValue`. These include:
- Zero (Market value and price is the same)
- No input
- BigNumber

frontend/components/utils/AnimatedNumber.vue
```diff
@@ -30,6 +32,16 @@ export default Vue.extend({
         },
     },
     computed: {
+        displayFullNumber(): boolean {
+            return this.decimals <= 5;
```
valiafetisov, line 36: Let's put 5 on top of the file as `DECIMAL_PLACES_MAX`

frontend/components/utils/AnimatedNumber.vue
```diff
@@ -30,6 +32,16 @@ export default Vue.extend({
...
+        },
+        decimals(): number {
+            if (this.isValidNumber) {
+                const decimalCutOff = Math.abs(Math.floor(Math.log10(Number(this.value))));
+                return decimalCutOff === Infinity ? this.decimalPlaces || 2 : decimalCutOff;
+            }
+            return this.decimalPlaces || 2;
```
valiafetisov, line 43: And `2` as `DECIMAL_PLACES_DEFAULT`

frontend/components/utils/AnimatedNumber.vue
```diff
@@ -1,10 +1,12 @@
...
-        :format-value="formatValue"
-        :duration="duration"
-    />
+    <span v-if="isValidNumber"
+        ><animated-number
+            v-if="displayFullNumber"
+            :value="formatValue(value)"
+            :format-value="formatValue"
+            :duration="duration"
+        /><span v-else> under 0.00001 </span></span
```
valiafetisov, line 8: - Can you please pass `0.00001` as a value to `animated-number`, so it will be properly animated (since we're doing it inside `AnimatedNumber` I would expect it to keep being animated)
- Can you please add negative value as a story (so we can see how it works with the `-`)

benediktkaiser: > * Can you please add negative value as a story (so we can see how it works with the `-`)

I added support for negative numbers and added stories for them as well!

frontend/components/utils/AnimatedNumber.vue
```diff
@@ -26,10 +31,22 @@ export default Vue.extend({
...
         decimalPlaces: {
             type: Number,
-            default: 2,
+            default: DECIMAL_PLACES_DEFAULT,
         },
     },
     computed: {
+        displayFullNumber(): boolean {
+            return this.decimals <= DECIMAL_PLACES_MAX || this.value < 0;
+        },
```
valiafetisov, line 40: This means, if the `value` is less than 0, let's say `-0.1` we will `displayFullNumber`, which will, according to the template, display `0.00001`

valiafetisov (commented):
I think I will take this one over to not stress you out with more comments 🙂

valiafetisov (approved):

LukSteib (changes requested):
- [ ] Looking at the stories `'Big Number 0.0004547'` and `'Number 0.000454786546'` are both displayed as  `0.00 DAI`. Since they are not `< 0.00001` I would expect `0.00045` in these cases. Pls explain if I am not getting the new logic applied here.

valiafetisov:
> Looking at the stories

Please check again, since the fix was pushed right before you've done a review 😞

LukSteib (approved):

PR #64, opened 2022-01-26
Remove dai.js dependency

### LINKED ISSUE #31: Support new collateral types
### Goal
Support new collateral types like `MATIC`, `WSTETH` and others.

### Context
- if https://github.com/sidestream-tech/auction-ui/issues/328 is implemented
  - update collateral config and test that all collaterals are tradable on uniswap
- if we don't implement https://github.com/sidestream-tech/auction-ui/issues/328
  - update dai.js version, which [was updated very recently](https://github.com/makerdao/dai.js/commit/1328fc42b7152297372f53f6e19254f7af132b04), with new collateral types
  **Note**: updating dai.js version may break kovan support as its no longer officially supported.

### Tasks 
- [x] determine full list of collateral types that needs to be added, document it here
- [x] add new collateral types to the `/constants/COLLATERALS.ts` config
- [ ] extensively test
  - [ ] create new auction types on kovan
  - [ ] test bidding for new and old auction types
  - [ ] test auth/deauth for new collateral types
  - [ ] validate that new collateral types have icons

valiafetisov:
- Removes dai.js dependency
- Also closes #31 

MATIC auction is fetched, but there is no liquidity (maybe it's the time when we learn how to liquidity on a testchain?):

Screenshot 2022-01-26 at 13 40 01


valiafetisov:
No more dai.js dependencies starting from https://github.com/sidestream-tech/unified-auctions-ui/pull/64/commits/33cf8a98bef40185f678c7b920fe8eeb4d461dfd

Everything seems to work based on successful execution of the ETH-A auction https://kovan.etherscan.io/tx/0x32ec2987b8e58076504c602da15c87e5ead4439187dc30be12597baa0e601871

frontend/store/wallet.ts
```diff
@@ -61,19 +60,13 @@ export const actions = {
...
     },
-    async connect({ commit, rootGetters, dispatch }: ActionContext<State, State>, walletType: string): Promise<void> {
+    async connect({ commit, dispatch }: ActionContext<State, State>, walletType: string): Promise<void> {
         dispatch('authorizations/reset', undefined, { root: true });
         commit('setIsLoading', true);
         try {
             const wallet = getWallet(walletType);
             await wallet.connect();
             commit('setAddress', wallet.address);
-            if (wallet.address) {
```
BracketJohn, line 71: What changed now, that wallet will always have an address?

core/src/contracts.ts
```diff
@@ -12,6 +12,11 @@ export const getClipperNameByCollateralType = function (collateralType: string):
     return `MCD_CLIP_${suffix}`;
 };
 
+export const getJoinNameByCollateralType = function (collateralType: string): string {
```
BracketJohn, line 15: q: what's the join name of a collateral type?

BracketJohn (changes requested):

frontend/store/wallet.ts
valiafetisov: No, the call within `if` is no longer needed, since the way we "attach" wallet to app changed

LukSteib (approved):
Approve part of auction fetching (and all other steps that formerly relied on `dai.js`) -> Tested full circle

What is still outstanding:
Find a way to properly test new collateral types (e.g. `MATIC`) on testchain.
This however should not block merge of this PR. We need to take the topic of testing with us for future considerations

core/src/contracts.ts
valiafetisov: It's a helper to construct contract name `MCD_JOIN_*`. I guess it will be better to call it `getJoinContractNameByCollateralType`, but since it's already approved I will change it in another PR

PR #68, opened 2022-01-27
Bid with DAI tooltip

### LINKED ISSUE #67: Adjust `Purchase with DAI` tooltip text
**Goal:**
Point users to the correct UI as long as we haven't reached feature parity with [https://liquidations.makerdao.com/](Liquidation Portal).

**Context:**
In our discussion to align with DUX (maintainers of existing liquidation portal) we decided to improve guidance for users to the suitable UI depending on the current use case. Either participate without Dai in auctions (ie. flashloan based participation that we offer) and to bid with Dai (feature that we haven't implemented yet but liq portal has).
* Liq portal will show a banner that informs user on our UI being the one to participate via flashloan without capital requirements
* We point users that want to bid with Dai to liq portal

**Tasks:**
- [x] Adjust text of our `Purchase with DAI` button to `Bid with DAI` (reasoning: Bidding implies that the auction can still be won by somebody else)
   - [x] Adjust the explainer text above accordingly and write `Bid with DAI` for first option instead 
- [x] Adjust text of current tooltip for this button to `This website doesn't support to bid on the auction with your own DAI yet. You can do so on the [Liquidation Portal]<https://liquidations.makerdao.com/>`

valiafetisov:
Closes #67 

- The text is a bit different than proposed
- I've also added the link to the explanation itself, since it's more user-friendly than figuring out to hover over an inactive button 🙂 

Screenshot 2022-01-27 at 16 41 34

### DIFF
--- a/frontend/components/Auction.vue
+++ b/frontend/components/Auction.vue
@@ -146,8 +146,10 @@
                     There are two ways to participate in an auction:
                     <ul class="list-disc list-outside pl-5">
                         <li>
-                            Purchase with DAI: This allows the participant to manually bid DAI on the auctioned
-                            collateral and redeem the auctioned collateral. Partial bids are also allowed.
+                            Bid with DAI: This allows the participant to manually bid DAI on the auctioned collateral
+                            and redeem the auctioned collateral. This transaction type is not yet supported. In the
+                            meantime, you can use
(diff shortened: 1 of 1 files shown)


LukSteib (approved):
Approved - like your additions / adjustments 👍

PR #74, opened 2022-01-31
Added server side Auth

benediktkaiser:
If a private Wallet key is set in the `.env` of `/bot-twitter`, the system will automatically set a signer and authenticate with the key. On success it will console.info the wallet address of the authenticated wallet:

```js
> auction-twitter-bot@1.0.0 dev
> nuxt


WARN  twitter: skipping twitter setup due to development mode             15:49:47

ℹ Authenticated with wallet "***********" 
```

For testing purposes: Our Wallet private key can be found in the secrets.

### DIFF
--- a/bot-twitter/package.json
+++ b/bot-twitter/package.json
@@ -13,6 +13,7 @@
     "start": "nuxt start",
     "generate": "nuxt generate",
     "lint": "eslint --ext \".js,.ts\" --ignore-path .gitignore . --max-warnings=0",
+    "debug": "node --inspect node_modules/.bin/nuxt",
     "test": "jest"
   },
   "repository": {
--- a/bot-twitter/src/authorizations.ts
+++ b/bot-twitter/src/authorizations.ts
@@ -1,0 +1,13 @@
+import getSigner from 'auctions-core/src/signer';
+import { authorizeWallet, getWalletAuthorizationStatus } from 'auctions-core/src/authorizations';
+
+export async function setupWallet(network: string) {
+    const signer = getSigner(network);
+    const walletAddress = await signer.getAddress();
+    const isAuth = await getWalletAuthorizationStatus(network, walletAddress);
+
+    if (!isAuth) {
+        console.info('Wallet has not been authorized yet. Attempting authorization now.');
+        await authorizeWallet(network, false);
+    }
+}
--- a/bot-twitter/src/authorizations.ts
+++ b/bot-twitter/src/authorizations.ts
@@ -10,6 +10,6 @@ export async function setupWallet(network: string) {
         return;
     }
     console.info(`Wallet "${walletAddress}" has not been authorized yet. Attempting authorization now`);
-    await authorizeWallet(network, false);
-    console.info(`Wallet "${walletAddress}" successfully authorized`);
+    const transactionHash = await authorizeWallet(network, false);
+    console.info(`Wallet "${walletAddress}" successfully authorized via "${transactionHash}" transaction`);
 }
--- a/bot-twitter/src/authorizations.ts
+++ b/bot-twitter/src/authorizations.ts
@@ -2,12 +2,14 @@ import getSigner from 'auctions-core/src/signer';
 import { authorizeWallet, getWalletAuthorizationStatus } from 'auctions-core/src/authorizations';
 
 export async function setupWallet(network: string) {
-    const signer = getSigner(network);
+    const signer = await getSigner(network);
     const walletAddress = await signer.getAddress();
     const isAuth = await getWalletAuthorizationStatus(network, walletAddress);
-
-    if (!isAuth) {
-        console.info('Wallet has not been authorized yet. Attempting authorization now.');
-        await authorizeWallet(network, false);
+    if (isAuth) {
+        console.info(`Wallet "${walletAddress}" has already been authorized`);
+        return;
     }
+    console.info(`Wallet "${walletAddress}" has not been authorized yet. Attempting authorization now`);
+    await authorizeWallet(network, false);
+    console.info(`Wallet "${walletAddress}" successfully authorized`);
 }
--- a/bot-twitter/src/index.ts
+++ b/bot-twitter/src/index.ts
@@ -2,6 +2,7 @@ import { getNetworkConfigByType } from 'auctions-core/src/constants/NETWORKS';
 import { setSigner } from 'auctions-core/src/signer';
 import { getNewAuctions } from './auctions';
 import notify from './notify';
+import { setupWallet } from './authorizations';
 
 const NETWORK = process.env.ETHEREUM_NETWORK || 'kovan';
 const DEFAULT_REFETCH_INTERVAL = 60 * 1000;
@@ -26,6 +27,7 @@ const loop = async function (): Promise<void> {
 const setup = async function (): Promise<void> {
     if (process.env.WALLET_PRIVATE_KEY) {
         await setSigner(NETWORK, process.env.WALLET_PRIVATE_KEY);
+        await setupWallet(NETWORK);
     }
     await loop();
 };
--- a/bot-twitter/src/index.ts
+++ b/bot-twitter/src/index.ts
@@ -1,16 +1,27 @@
+import NETWORKS from 'auctions-core/src/constants/NETWORKS';
+import { setSigner } from 'auctions-core/src/signer';
 import { getNewAuctions } from './auctions';
 import notify from './notify';
 
+const NETWORK = process.env.ETHEREUM_NETWORK || 'kovan';
 const DEFAULT_REFETCH_INTERVAL = 60 * 1000;
 const REFETCH_INTERVAL = parseInt(process.env.REFETCH_INTERVAL ?? '') || DEFAULT_REFETCH_INTERVAL;
 let refetchIntervalId: ReturnType<typeof setInterval> | undefined;
+
+if (!NETWORKS[NETWORK]) {
+    throw new Error(`network "${NETWORK}" is not supported`);
+}
+
+if (process.env.WALLET_PRIVATE_KEY) {
+    setSigner(NETWORK, process.env.WALLET_PRIVATE_KEY).then(() => {});
+}
 
 const loop = async function (): Promise<void> {
     if (refetchIntervalId) {
         clearInterval(refetchIntervalId);
     }
     try {
-        (await getNewAuctions()).map(notify);
+        (await getNewAuctions(NETWORK)).map(notify);
     } catch (error) {
         console.error('loop: error', error);
     } finally {
@@ -18,4 +29,4 @@ const loop = async function (): Promise<void> {
     }
 };
 
-loop();
+loop().then(() => {});
--- a/bot-twitter/src/index.ts
+++ b/bot-twitter/src/index.ts
@@ -1,4 +1,4 @@
-import NETWORKS from 'auctions-core/src/constants/NETWORKS';
+import { getNetworkConfigByType } from 'auctions-core/src/constants/NETWORKS';
 import { setSigner } from 'auctions-core/src/signer';
 import { getNewAuctions } from './auctions';
 import notify from './notify';
@@ -8,13 +8,7 @@ const DEFAULT_REFETCH_INTERVAL = 60 * 1000;
 const REFETCH_INTERVAL = parseInt(process.env.REFETCH_INTERVAL ?? '') || DEFAULT_REFETCH_INTERVAL;
 let refetchIntervalId: ReturnType<typeof setInterval> | undefined;
 
-if (!NETWORKS[NETWORK]) {
-    throw new Error(`network "${NETWORK}" is not supported`);
-}
-
-if (process.env.WALLET_PRIVATE_KEY) {
-    setSigner(NETWORK, process.env.WALLET_PRIVATE_KEY).then(() => {});
-}
+getNetworkConfigByType(NETWORK);
 
 const loop = async function (): Promise<void> {
     if (refetchIntervalId) {
@@ -29,4 +23,13 @@ const loop = async function (): Promise<void> {
     }
 };
 
-loop().then(() => {});
+const setup = async function (): Promise<void> {
+    if (process.env.WALLET_PRIVATE_KEY) {
+        await setSigner(NETWORK, process.env.WALLET_PRIVATE_KEY);
+    }
+    await loop();
+};
+
+setup().catch(error => {
+    throw error;
+});
--- a/bot-twitter/src/index.ts
+++ b/bot-twitter/src/index.ts
@@ -1,5 +1,5 @@
 import { getNetworkConfigByType } from 'auctions-core/src/constants/NETWORKS';
-import { setSigner } from 'auctions-core/src/signer';
+import { createSigner } from 'auctions-core/src/signer';
 import { getNewAuctions } from './auctions';
 import notify from './notify';
 import { setupWallet } from './authorizations';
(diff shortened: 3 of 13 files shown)


core/src/signer.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const setSigner = async function (network: string, privateKey: string) {
+    const provider = getProvider(network);
+
+    try {
+        signers[network] = new ethers.Wallet(privateKey, provider);
+        const address = await signers[network].getAddress();
+        console.info(`Authenticated with wallet "${address}"`);
+    } catch (error) {
+        console.error(`Could not authenticate with the private key: ${error}`);
```
valiafetisov, line 26: Should we instead re-throw the error here? How else would the parent function calling `setSigner` know that it failed?

core/src/signer.ts
```diff
@@ -0,0 +1,30 @@
+import { ethers } from 'ethers';
+import NETWORKS from './constants/NETWORKS';
+import getProvider from './provider';
+
+const signers: Record<string, ethers.Wallet> = {};
+
+const getSigner = function (network: string): ethers.Wallet | undefined {
+    if (!NETWORKS[network]) {
+        throw new Error(`The network "${network}" is not supported yet!`);
+    }
```
valiafetisov, line 10: This error and the check should be better stored in the `NETWORKS.ts`
```suggestion
    getNetworkConfigByType(network);
```
– will produce the same result

core/src/signer.ts
```diff
@@ -0,0 +1,30 @@
...
+
+    if (!signers[network]) {
+        return undefined;
+    }
```
valiafetisov, line 14: This should throw an error instead. You don't want to check `getSigner` against undefined every time you used it.

bot-twitter/src/index.ts
```diff
@@ -1,21 +1,32 @@
...
+        (await getNewAuctions(NETWORK)).map(notify);
     } catch (error) {
         console.error('loop: error', error);
     } finally {
         refetchIntervalId = setTimeout(loop, REFETCH_INTERVAL);
     }
 };
 
-loop();
+loop().then(() => {});
```
valiafetisov, line 32: What's the purpose of the `.then(() => {})`?

bot-twitter/src/index.ts
```diff
@@ -1,21 +1,32 @@
...
 const DEFAULT_REFETCH_INTERVAL = 60 * 1000;
 const REFETCH_INTERVAL = parseInt(process.env.REFETCH_INTERVAL ?? '') || DEFAULT_REFETCH_INTERVAL;
 let refetchIntervalId: ReturnType<typeof setInterval> | undefined;
 
+if (!NETWORKS[NETWORK]) {
+    throw new Error(`network "${NETWORK}" is not supported`);
+}
+
+if (process.env.WALLET_PRIVATE_KEY) {
+    setSigner(NETWORK, process.env.WALLET_PRIVATE_KEY).then(() => {});
```
valiafetisov, line 16: - same as above
- if setSigner is async the error it throws will not be handled anywhere, so you need to await it to produce some kind of effect (ie stop program from executing). I suggest to define `setup` function and call seSigner there, right before initiating the loop. This way, the loop will not start if setSigner throws an error

valiafetisov:
Please note that the original issue (which is not linked here) suggest to execute `authorizeWallet`. For this, you may need to use newly created `getSigner` somewhere inside `executeTransaction` and change how metamask wallet works as well (is use setSigner instead of setProvider there)

benediktkaiser:
So I am running into the following issue: 

- I start the bot
- 5 minutes later I get the following error: 

```js
 ERROR  could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.5.0)                                                                                                                         12:12:46

  at Logger.makeError (/unified-auctions-ui/core/node_modules/@ethersproject/logger/lib/index.js:199:21)
  at Logger.throwError (/unified-auctions-ui/core/node_modules/@ethersproject/logger/lib/index.js:208:20)
  at JsonRpcProvider.<anonymous> (/unified-auctions-ui/core/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:517:54)
  at step (/unified-auctions-ui/core/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:48:23)
  at Object.throw (/unified-auctions-ui/core/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:29:53)
  at rejected (/unified-auctions-ui/core/node_modules/@ethersproject/providers/lib/json-rpc-provider.js:21:65)
```

Could this be because I do not use `setProvider()` in the Twitter bot? I assumed it would not be an issue as when `getProvider()` is called, a new provider is created if none exists:

```js
if (!providers[network]) {
    const networkUrl = NETWORKS[network].url;
    providers[network] = new ethers.providers.JsonRpcProvider(networkUrl);
}
```

Is this line of thinking correct or am I making a mistake? Thank you in advance for the help!

valiafetisov:
Another hint: to speedup your development, you may want to set lower timeout to the `JsonRpcProvider`:
- https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider
- https://docs.ethers.io/v5/api/utils/web/#ConnectionInfo

benediktkaiser:
> According to [this comment](https://github.com/ethers-io/ethers.js/issues/1316#issuecomment-887760999), I wonder if you have _proper_ infura key set in the `bot-twitter/.env`? (note that keys on frontend and in the bot are different, use `secret` to get correct one)

I checked again and I am using the correct secrets (unless our key in the secrets is outdated).

valiafetisov:
Then, the other suggestion found on the same linked issue is to `await provider.ready()` before using the provider (you can make getProvider async and validate it there, for example) https://docs.ethers.io/v5/api/providers/provider/#Provider-ready

benediktkaiser:
> Then, the other suggestion found on the same linked issue is to `await provider.ready()` before using the provider (you can make getProvider async and validate it there, for example) https://docs.ethers.io/v5/api/providers/provider/#Provider-ready

I tried using `.ready()`, but that function does not seem to exist.

I followed the issue even futhur and identified that it gets stuck while trying to call the contracts:

```js
const contractNames: string[] = await contract.list();
```
[`/core/src/addresses.ts#21`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/addresses.ts#L21)

Could the issue have to do with something else or is this still the provider?

valiafetisov:
> to retrieve the contract

I don't think there is any retrieval happening. I am also not sure why are we awaiting `new ethers.Contract` since it's a [not async](https://docs.ethers.io/v5/api/contract/contract/#Contract--creating). `contract.list()` since to be the first call, when the provider is being used.

benediktkaiser:
> Another hint: to speedup your development, you may want to set lower timeout to the `JsonRpcProvider`:
> 
> * https://docs.ethers.io/v5/api/providers/jsonrpc-provider/#JsonRpcProvider
> * https://docs.ethers.io/v5/api/utils/web/#ConnectionInfo

Another question about this:

When setting the provider I console logged `providers[network].connection`. It gave me back the url property of the object. However when I tried to set a custom timeout with `providers[network].connection.timeout = X` I got the following error:

` ERROR  Cannot add property timeout, object is not extensible                                                                                                                                                             13:45:33`

How can I set the timeout?

valiafetisov:
According to the linked docs `ethers.providers.JsonRpcProvider` can accept [the object](https://docs.ethers.io/v5/api/utils/web/#ConnectionInfo) instead of string with the url.

benediktkaiser:
When debugging etherJS with `provider.on('debug', console.info);`, I got the following error message:

```js
ℹ {                                                                                                                                                                                                                       13:59:27
  action: 'response',
  error: Error: timeout (requestBody="{\"method\":\"net_version\",\"params\":[],\"id\":45,\"jsonrpc\":\"2.0\"}", requestMethod="POST", timeout=1000, url="https://mainnet.infura.io/v3/***********", code=TIMEOUT, version=web/5.5.0)
      at Logger.makeError (/unified-auctions-ui/core/node_modules/@ethersproject/logger/lib/index.js:199:21)
      at Timeout._onTimeout (/unified-auctions-ui/core/node_modules/@ethersproject/web/lib/index.js:149:35)
      at listOnTimeout (node:internal/timers:557:17)
      at processTimers (node:internal/timers:500:7) {
    reason: 'timeout',
    code: 'TIMEOUT',
    requestBody: '{"method":"net_version","params":[],"id":45,"jsonrpc":"2.0"}',
    requestMethod: 'POST',
    timeout: 1000,
    url: 'https://mainnet.infura.io/v3/***********'
  },
  request: {
    method: 'net_version',
    params: [],
    id: 45,
    jsonrpc: '2.0'
  },
  provider: JsonRpcProvider {
    _isProvider: true,
    _events: [
      [Event]
    ],
    _emitted: {
      block: -2
    },
    formatter: Formatter {
      formats: [Object]
    },
    anyNetwork: false,
    _networkPromise: Promise {
      <rejected> [Error]
    },
    _maxInternalBlockNumber: -1024,
    _lastBlockNumber: -2,
    _pollingInterval: 4000,
    _fastQueryDate: 0,
    connection: {
      url: 'https://mainnet.infura.io/v3/***********',
      timeout: 1000
    },
    _nextId: 46,
    _eventLoopCache: {
      detectNetwork: null,
      eth_chainId: null
    }
  }
}
```

I don't know if this helps in any way.

valiafetisov:
@benediktkaiser the problem was the infura key stored in `auction-ui/localhost/bot-twitter`. Using production key worked, I though you've tested it in the very beginning.

@florianstoecker:
- please create a new infura key just for the development
- please notify us of keys that are reaching the 100% quota in the future (I know infura sends a lot of emails about it)!

valiafetisov (approved):

florianstoecker:
> * please create a new infura key just for the development

Done. It's in bitwarden under "Dev Key Infura"

> * please notify us of keys that are reaching the 100% quota in the future (I know infura sends a lot of emails about it)!

Our limits were not reached since the 22nd (where we've upgraded). They weren't even close. But I'll notify once we're getting closer. The only thing I could imagine where the limits were reached: We have two separate keys for the bot already - Twitter Bot Staging & Twitter Bot Production. We've limited our staging bot to 100K requests a day. If this key was used here, then this could be the reason why it ran out of requests.

benediktkaiser:
> @benediktkaiser the problem was the infura key stored in `auction-ui/localhost/bot-twitter`. Using production key worked, I though you've tested it in the very beginning 😅

I tested it with the key stored in `auction-ui/localhost/bot-twitter`. I don't even know if I have access to the production keys. 😅

florianstoecker:
> * please do a frontend functional review that no wallet-related functionality is broken

Did following actions successfully on kovan:
- connected my wallet
- restarted 1 aave, 1 balancer, 2 ETH auctions
- Authorized DAI transactions
- Authorized ETH-A transactions
- [Executed an ETH-A auction](https://kovan.etherscan.io/tx/0x878fb1a0399407132bffafb6433ac10ac94f185d9be1f65f600895d9ae6d0755) 


One link auction did not restart on the first attempt & I've got following error shown in the console:
```Auction redo error: Error: transaction failed (transactionHash="0x9fffa5f09d7a34cef78b12e908305b5c7180ad66ce67cf5d048fa2023fce3ffe", transaction={"hash":"0x9fffa5f09d7a34cef78b12e908305b5c7180ad66ce67cf5d048fa2023fce3ffe","type":2,"accessList":null,"blockHash":null,"blockNumber":null,"transactionIndex":null,"confirmations":0,"from":"0x4f3B7fb6204C614B28F0e1f50Dc1C99175631678","gasPrice":{"type":"BigNumber","hex":"0x77359400"},"maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x77359400"},"maxFeePerGas":{"type":"BigNumber","hex":"0x77359400"},"gasLimit":{"type":"BigNumber","hex":"0x022d9a"},"to":"0x1eB71cC879960606F8ab0E02b3668EEf92CE6D98","value":{"type":"BigNumber","hex":"0x00"},"nonce":1020,"data":"0xd843416d00000000000000000000000000000000000000000000000000000000000000c80000000000000000000000004f3b7fb6204c614b28f0e1f50dc1c99175631678","r":"0xd5e351c172ec2ec517470085b08b4e86cc6bf74e37bdd0e10dbba9e8de175bc0","s":"0x3de59b163aa0edeef348ed0cffc1f3c2ecc3de0efdda53db8d79aebe8e430bcd","v":1,"creates":null,"chainId":0}, receipt={"to":"0x1eB71cC879960606F8ab0E02b3668EEf92CE6D98","from":"0x4f3B7fb6204C614B28F0e1f50Dc1C99175631678","contractAddress":null,"transactionIndex":3,"gasUsed":{"type":"BigNumber","hex":"0x022d9a"},"logsBloom":"0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","blockHash":"0xebcaf2c0e55de53535e29fe7c8d35c95ac740d2d273cbff79cc18c22628c8baf","transactionHash":"0x9fffa5f09d7a34cef78b12e908305b5c7180ad66ce67cf5d048fa2023fce3ffe","logs":[],"blockNumber":29645633,"confirmations":1,"cumulativeGasUsed":{"type":"BigNumber","hex":"0x0795e1"},"effectiveGasPrice":{"type":"BigNumber","hex":"0x77359400"},"status":0,"type":2,"byzantium":true}, code=CALL_EXCEPTION, version=providers/5.4.5)```

**Comment by benediktkaiser** [comment]
> One link auction did not restart on the first attempt & I've got following error shown in the console:

We just discussed in the Daily that this is an error that we have run into outside of this PR. @florianstoecker will open a new issue in which we can investigate why this behaviour happens.

File: bot-twitter/package.json [hunk 6]
```diff
@@ -13,6 +13,7 @@
     "start": "nuxt start",
     "generate": "nuxt generate",
     "lint": "eslint --ext \".js,.ts\" --ignore-path .gitignore . --max-warnings=0",
+    "debug": "node --inspect node_modules/.bin/nuxt",
```
benediktkaiser, line 16: What does this debug do and what can we use it for in the future? (Just interested)

bot-twitter/src/authorizations.ts
```diff
@@ -0,0 +1,15 @@
...
+    const signer = await getSigner(network);
+    const walletAddress = await signer.getAddress();
+    const isAuth = await getWalletAuthorizationStatus(network, walletAddress);
+    if (isAuth) {
+        console.info(`Wallet "${walletAddress}" has already been authorized`);
+        return;
+    }
+    console.info(`Wallet "${walletAddress}" has not been authorized yet. Attempting authorization now`);
+    const transactionHash = await authorizeWallet(network, false);
+    console.info(`Wallet "${walletAddress}" successfully authorized via "${transactionHash}" transaction`);
```
benediktkaiser, line 14: What happens if the authorization fails?

benediktkaiser (commented):
Everything looks really good. Glad to see it finally working!

bot-twitter/package.json
valiafetisov: `node --inspect` starts a websockets endpoint, to which you can connect your chrome console and have access to the running program. More info here: https://nodejs.org/en/docs/guides/debugging-getting-started/

bot-twitter/src/authorizations.ts
valiafetisov: `authorizeWallet` will throw an error that will be propagated up to the `setup`, where it will throw unhandled error and halt bot execution. This is fine for this PR, although in the #7 I propose to do authorisation only when it's profitable to do so, therefore it will be part of the loop in the future

valiafetisov:
If the problem describe in https://github.com/sidestream-tech/unified-auctions-ui/issues/86 is unrelated to this PR, @florianstoecker would you then approve it?

florianstoecker (approved):

Issue #73, opened 2022-01-31
Add `wstETH` collateral support


valiafetisov:
### Goal
Participation in `wstETH` collateral auctions is possible via the UI

### Context
`wstETH` collateral requires it's own callee contract and uses Uniswap v3 exchange, therefore:
- new callee address should be passed to the clipper contract (along with possibly different callee data)
- the price of the new collateral should be calculated separately

This issue will also be used to refactor callee logic to streamline collateral onboarding.

### Tasks
- [ ] Refactor callee / collateral logic
  - [ ] Import callee addresses from a single file (proposed in https://github.com/makerdao/exchange-callees/pull/22)
  - [ ] Make a declarative configuration: price calculation function + callee data generation per callee
- [ ] Add new callee support
  - [ ] Deploy callee to test networks if needed
  - [ ] Add uniswap v3 library
  - [ ] Generate callee data
  - [ ] Calculate price
- [ ] Add new collateral config
- [ ] Manually validate that new collateral is supported

valiafetisov:
> Add new callee support

The callee in question is https://github.com/makerdao/exchange-callees/blob/master/src/WstETHCurveUniv3Callee.sol

But I can not find its address, since action-demo-keeper [doesn't seem to support this collateral](https://github.com/makerdao/auction-demo-keeper/blob/main/config/examples/mainnet-uniswapv3.json).

Till we get the address from the PR (by merging the PR and asking them to fill the json), I can try to deploy it myself, but it's not clear for me, what constructor parameters should be used. The contract accepts: 

- `address curvePool_`: supposedly `CurveContractV1.vy` contract from the official docs https://curve.readthedocs.io/ref-addresses.html, [but it can not be find on test chains](https://blockscan.com/address/0x845838DF265Dcd2c412A1Dc9e959c7d08537f8a2)
- `address uniV3Router_`: supposedly [`0xe592427a0aece92de3edee1f18e0157c05861564`](https://etherscan.io/address/0xe592427a0aece92de3edee1f18e0157c05861564#code) (across networks)
- `address daiJoin_`: `MCD_JOIN_DAI` from the https://chainlog.makerdao.com
- `address weth_`: supposedly `WETH` token address, also from the https://chainlog.makerdao.com

valiafetisov:
[...]

PR #77, opened 2022-02-01
Callee refactoring

valiafetisov:
Partially addresses #73 

In this PR all logic related to callees and uniswap was rearranged and splitted into different files. Now every callee is required (enforced by the type) to have two functions: one for calculating the price of the collateral, the other is for producing the parameters for the transaction. Therefore, both things should be tested again for each of two callees that we support


benediktkaiser (dismissed):
I had a flyover of the code. It all looks very clean to me. I like the approach you took to add different kinds of callee contracts. Should make collateral onboarding a lot easier!

I did not do a functional review.

LukSteib:
@valiafetisov can you merge `main` with the vuex fix into this branch pls

LukSteib (commented):
Successfully tested UniswapV2Callee on kovan:
https://kovan.etherscan.io/tx/0xda3cbb4037772da1e1d2a2c6dfc88190fcee3cc7d7f66b0f40d07a02f7b6d247

Still pending:
Test UniswapV2LpTokenCallee - might take some time due to necessity of:
* vault creation
* vault liquidation
* auction participation

LukSteib (changes requested):
- [x] For the UniV2 LP-tokens we are not able to determine the price which hinders execution. Problem exists on both kovan [1] and mainnet [2]. On `kovan` we should be able to determine the price for `UNIV2DAIETH-A` at least. On `mainnet`we shoud be able to determine prices for all of them 

[1]
image

[2]
image

valiafetisov:
Thanks for spotting it, I've imported the same file twice 😅. It works now, but I haven't checked if the displayed price is correct.

Screenshot 2022-02-04 at 16 18 47

LukSteib (approved):
Successfully executed a UniswapV2LpTokenCallee based tx on kovan: 

https://kovan.etherscan.io/tx/0xa35eb352b028c25a5719d411b1280cf13b84029b11645d12452885818369f316

PR #78, opened 2022-02-01
Callee contract table on dashboard

### LINKED ISSUE #75: Include callee contract reference to /dashboard
**Goal:**
Reference the `exchange-callee` contracts we are using to execute the flash loans with their respective addresses on the dashboard

**Context:**
Depending on the collateral type we are using different exchange callee contracts in order to execute auctions. For each network these contracts need to be deployed separately. The dashboard page is a good place to show the contracts we are using and include a link to the respective etherscan page of the contract. Similar to us displaying the token addresses for the collateral types

Since the callee references do not fit in the collateral table let's create a second one. With a dedicated heading and text block. And show a two column table with `Callee` and `address`. 

Example:
| Callee Contract  | Address |
| ------------- | ------------- |
| uniswapV2Callee  | 0x5A40F810754f725DA93e2362775a0600468f7a83 |
| uniswapV2LpTokenCallee  | 0xDeC8b9c2829583A89f7F182DEeD7C12112dfAeD0 |

**Tasks:**
- [x] Include new text block to dashboard page below collateral table
   - [x] Heading: Exchange Callee Contracts
   - [x] Text: see file in assets below
- [x] Implement table following the example above
   - [x] Check with @valiafetisov whether it makes sense to wait for implementation of this one since we might have a more structure way in our files to store callee addresses 
   - [x] Ensure that we dynamically display the correct callee names and addresses depending on the network and what we support

**Assets:**
Doc with text: https://docs.google.com/document/d/1DdvTn84XpS4kaTk-oh1ADiTh38SDvvPE2i9IygCgWWQ/edit?usp=sharing

fegbert:
closes #75 

**preview**

image


LukSteib (dismissed):

frontend/containers/DashboardContainer.vue
```diff
@@ -25,6 +30,13 @@ export default Vue.extend({
                 this.$store.dispatch('preferences/setExplanationsAction', newIsExplanationsShown);
             },
         },
+        callees(): CalleeAddresses {
+            const pageNetwork = this.$store?.getters['network/getPageNetwork'];
+            if (pageNetwork === 'stub') {
+                return {};
+            }
```
valiafetisov, line 37: As `stub` string is configurable and there can be more networks, let's instead use `getMakerNetwork` and check that it's not empty here

frontend/components/CalleeTable.vue
```diff
@@ -0,0 +1,50 @@
+<template>
+    <table class="Table">
+        <tr>
+            <th class="Heading">Callee Contract</th>
+            <th class="Heading">Address</th>
+        </tr>
+        <tr v-for="[callee, address] of Object.entries(callees)" :key="callee">
```
valiafetisov, line 7: What do we display if there are no callees (ie, empty object is provided)?
Please handle this properly and add a story for this case

fegbert: Added `not found` text as full row:

image

Is the text fine? 🙂

frontend/components/DashboardAuctionsView.vue
```diff
@@ -45,6 +54,10 @@ export default Vue.extend({
             type: Array as Vue.PropType<CollateralRow[]>,
             default: () => [],
         },
+        callees: {
+            type: Object as Vue.PropType<CalleeAddresses>,
+            default: () => {},
```
valiafetisov, line 59: This returns undefined. If you want to return an empty object:
```suggestion
            default: () => ({}),
```

frontend/components/CalleeTable.vue
```diff
@@ -0,0 +1,67 @@
...
+import FormatAddress from '~/components/utils/FormatAddress.vue';
+
+export default Vue.extend({
+    components: {
+        FormatAddress,
+    },
+    props: {
+        callees: {
+            type: Object as Vue.PropType<CalleeAddresses>,
+            default: () => {},
```
valiafetisov, line 36: Same here (:

frontend/components/CalleeTable.vue
```diff
@@ -0,0 +1,67 @@
...
+        },
+    },
+    computed: {
+        isCalleesEmpty() {
+            return Object.keys(this.callees).length === 0;
```
valiafetisov, line 41: Otherwise this will throw an error `Cannot convert undefined or null to object`

valiafetisov (approved):

LukSteib (approved):
Approved from my side
@valiafetisov thanks for pointing out the no data edge case - I definitely should have covered that one during my review 🙈

PR #82, opened 2022-02-02
Moved fee estimation and profitability calculation to `/core`

### LINKED ISSUE #80: Move fee estimation and profitability calculation to `/core`
### Goal:

In order to complete #45 we need to move some essential function from the frontend into the core folder. This allows us to use these function in other parts of the project such as the keeper. 

### Context:

Fee estimation is currently done inside the store, therefore it needs to be refactored to the /core folder and use memoizee as a caching solution instead of the store.

### Tasks:

- [ ] Move fee estimation to /core
- [ ] Move profitability calculation to /core

benediktkaiser:
closes #80 

After some trying around I decided to only `memoizee` the functions. Moving them out of the store would require a larger rewrite of how they are imported and merged into the Auction Transaction. 

If this is not okay, then I can sit down and refractor them too!


valiafetisov:
We have to move them to the `core`, since we want to use them in the bot-twitter, that's the whole point of this refactoring. After they are moved to the `core`, you will no longer need `store/fee` file and can directly enrich the auction with fees like all other enrichers.

And, as always, it's better to cache lower-level function that does the fetching (`getGasPrice`) instead.

benediktkaiser:
> We have to move them to the `core`, since we want to use them in the bot-twitter, that's the whole point of this refactoring. After they are moved to the `core`, you will no longer need `store/fee` file and can directly enrich the auction with fees like all other enrichers.

The functions to calculate the fee etc are already in the core. The store only temporary saves them for auction injecting, but all of the needed functions are already in the core. 

https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/fees.ts
https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/price.ts#L54

Or am I mistaken here?

valiafetisov:
> The functions to calculate the fee etc are already in the core

Perfect, then let's add cache and get rid of the store?

core/src/fees.ts
```diff
@@ -23,6 +26,12 @@ export const getApproximateTransactionFees = async function (network: string): P
     };
 };
 
+export const getApproximateTransactionFees = memoizee(_getApproximateTransactionFees, {
+    maxAge: TRANSACTION_FEE_CACHE,
+    promise: true,
+    length: 1,
+});
```
valiafetisov, line 33: As suggested already: can you please cache actual function that fetches gas price instead (and please use smaller cache, something like 5 seconds)?
As `getExchangeRateBySymbol` is already cached, as far as I remember.

benediktkaiser: > * As suggested [above](https://github.com/sidestream-tech/unified-auctions-ui/pull/82#issuecomment-1028081957): can you please cache actual function that fetches gas price instead (and please use smaller cache, something like 10 seconds)?

- [X] Cached the Gas price for 10 seconds
- [X] Removed cache on parent function

core/src/gas.ts
```diff
@@ -2,16 +2,24 @@ import { ethers } from 'ethers';
 import BigNumber from './bignumber';
 import { getNetworkConfigByType } from './constants/NETWORKS';
 import { ETH_NUMBER_OF_DIGITS } from './constants/UNITS';
+import memoizee from 'memoizee';
 
 const API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
 const TRANSACTION_SPEED = 'fast';
 
-const getCurrentGasPrice = async function (): Promise<BigNumber> {
+const GAS_CACHE = 10 * 60 * 1000;
```
valiafetisov, line 10: That is 10 minutes 😅

benediktkaiser: Oh god.. Hahaha

benediktkaiser: I set it to 10 seconds now!

valiafetisov (approved):

PR #85, opened 2022-02-04
Fixed VueX Fee bug

### LINKED ISSUE #84: Fix vuex/fees bug
### Goal
Fixed `[vuex] do not mutate vuex store state outside mutation handlers.`

### Context
After merging #80, we discovered a bug:

Screenshot 2022-02-03 at 20 44 23
Screenshot 2022-02-03 at 20 44 41

### Tasks
- [x] Investigate and document
- [x] Fix

benediktkaiser:
closes #84 

The first error message was due to us still calling:
```js
dispatch('fees/setup', null, { root: true });
```
https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/store/auctions.ts#L140

The other issue was that we were trying to mutate the store directly. Before we had a `.map()` function with which we made a new Object, which we then enriched with the restarting state. 

Once I removed the enriching with Fees we no longer made a new object resulting in us trying to mutate the store with `.foreach`, leading to the error. I changed the foreach to a map function which I then exported and everything is now working as intended!

### DIFF
--- a/frontend/store/auctions.ts
+++ b/frontend/store/auctions.ts
@@ -36,15 +36,18 @@ export const state = (): State => ({
 });
 
 export const getters = {
-    listAuctions(state: State) {
+    listAuctions(state: State): AuctionTransaction[] {
         return Object.values(state.auctionStorage);
     },
     listAuctionTransactions(state: State, getters: any, _rootState: any): AuctionTransaction[] {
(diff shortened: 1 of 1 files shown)


LukSteib (approved):
Approved due to the fact that I do not receive the `vuex` error in console any longer

valiafetisov (approved):

PR #88, opened 2022-02-07
Improve indication of restarting auctions

### LINKED ISSUE #28: Better indicate when needs to be restarted
**Goal**
Better indicate when an auction needs to be restarted.

**Context**
Currently, it is not that clear when an auction needs to be restarted and the columns `Market Difference` & `Auction Price` are showing _Unknown_. Let's discuss what can be done to better indicate that an auction needs to be restarted & explain the two columns showing unknown.

**Tasks**
- [x] When the auction is active/not finished, but not tradable, change `Market Difference` text from `Unknown` to `Not tradable`, but keep it gray (note that this should also be changed in the auction page)
- [x] Change `See details` button text to `Participate` when the auction is active
- [x] Change `See details` button text to `Restart auction` when the auction is inactive
- [x] Keep `See details` when the auction is finished
- [x] On the auction page: replace `Auction Price Total` with `Unknown` when the auction is inactive
   - [x] Add a tooltip(like in the screenshots below) here that says something like "Since the auction is not active, there is no total Auction Price for this auction." 
- [x] Add `Connect a Wallet` block with text above `Restart Auction` block in case no wallet is connected – so we have active call to action in the page at all times (right now there is no hint to the user why the restart button is inactive)

**Assets**
image

fegbert:
closes #28 

> When the auction is active/not finished, but not tradable, change `Market Difference` text from `Unknown` to `Not tradable`
image

> Change `See details` button text to `Participate` when the auction is active
image

> Change `See details` button text to `Restart auction` when the auction is inactive
image

> Keep `See details` when the auction is finished
image

> On the auction page: replace `Auction Price Total` with `Unknown` when the auction is inactive
>   - Add a tooltip(like in the screenshots below) here that says something like "Since the auction is not active, there is no total Auction Price for this auction."
image

> Add `Connect a Wallet` block with text above `Restart Auction` block in case no wallet is connected
image

Only showing while no wallet is connected & auction needs to be restarted.

### DIFF
--- a/frontend/components/Auction.vue
+++ b/frontend/components/Auction.vue
@@ -2,6 +2,13 @@
     <TextBlock :title="`Auction #${auctionId}`">
         <Alert v-if="errorText" class="my-3" :message="errorText" type="error" />
         <div v-if="auction">
+            <WalletBlock
+                v-if="!auction.isActive && !auction.isFinished && !walletAddress"
+                class="my-4"
+                :wallet-address="walletAddress"
+                :is-explanations-shown="isExplanationsShown"
+                @connectWallet="$emit('connect')"
+            />
             <RestartBlock
                 v-if="errorText === 'This auction is inactive and must be restarted'"
                 :transaction-fee="auction.restartTransactionFeeETH"
@@ -240,6 +247,7 @@ import Loading from '~/components/common/Loading.vue';
 import Explain from '~/components/utils/Explain.vue';
 import RestartBlock from '~/components/transaction/RestartBlock.vue';
 import TimeTillProfitable from '~/components/utils/TimeTillProfitable.vue';
+import WalletBlock from '~/components/transaction/WalletBlock.vue';
 
 export default Vue.extend({
     name: 'Auction',
@@ -258,6 +266,7 @@ export default Vue.extend({
         Tooltip,
         TimeTillProfitable,
         Popover,
+        WalletBlock,
     },
     props: {
         auction: {
--- a/frontend/components/Auction.vue
+++ b/frontend/components/Auction.vue
@@ -56,7 +56,7 @@
                                 <template v-if="auction.isActive && auction.marketUnitPriceToUnitPriceRatio">
                                     <format-market-value :value="auction.marketUnitPriceToUnitPriceRatio" />
                                 </template>
-                                <span v-else class="opacity-50">Unknown</span>
+                                <span v-else class="opacity-50">Not tradable</span>
                             </td>
                         </tr>
                         <tr>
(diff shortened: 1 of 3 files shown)


frontend/components/Auction.vue
```diff
@@ -2,6 +2,13 @@
     <TextBlock :title="`Auction #${auctionId}`">
         <Alert v-if="errorText" class="my-3" :message="errorText" type="error" />
         <div v-if="auction">
+            <WalletBlock
+                v-if="!auction.isActive && !auction.isFinished && !walletAddress"
+                class="my-4"
+                :wallet-address="walletAddress"
+                :is-explanations-shown="isExplanationsShown"
+                @connectWallet="$emit('connect')"
+            />
```
valiafetisov, line 11: Can you please move this inside `RestartBlock`? Since it's already have logic to display or not display itself and the wallet connection is practically part of the restart button

frontend/components/Auction.vue
```diff
@@ -56,7 +63,7 @@
                                 <template v-if="auction.isActive && auction.marketUnitPriceToUnitPriceRatio">
                                     <format-market-value :value="auction.marketUnitPriceToUnitPriceRatio" />
                                 </template>
-                                <span v-else class="opacity-50">Unknown</span>
+                                <span v-else class="opacity-50">Not tradable</span>
```
valiafetisov, line 66: Since it's `v-else`, it will also be shown if `!auction.isActive`, which is not correct. `Not tradable` should only be shown when there is no `marketUnitPriceToUnitPriceRatio`

frontend/components/AuctionsTable.vue
```diff
@@ -27,7 +27,7 @@
                 v-if="record.isActive && record.marketUnitPriceToUnitPriceRatio && !record.isFinished"
                 :value="marketUnitPriceToUnitPriceRatio"
             />
-            <span v-else class="opacity-50">Unknown</span>
+            <span v-else class="opacity-50">Not tradable</span>
```
valiafetisov, line 30: Same as above

valiafetisov (dismissed):

LukSteib (changes requested):
Was not in the scope of the initial issue but since we are working on improving UX for auctions that require restrart in this PR, I think it fits
- [x] Change explanation text below auction table for auctions that require restart since it is misleading for the user right now (see screenshot below)
   - [x] Keep first sentence
   - [ ] Replace rest simply with the following sentence: `This auction requires restart in order to determine prices properly.`
   - [x] ensure that the text as we have it right now is displayed for all active auctions

image

fegbert:
Should we also add another case for finished auctions? 🙂

LukSteib:
> Should we also add another case for finished auctions? 🙂

Good question. Let's not further increase the scope of this PR and get it merged. We can take text adjustment for finished auctions to the backlog (or cover in #90)

fegbert:
changed text:

image

LukSteib (approved):

valiafetisov (approved):

PR #90, opened 2022-02-07
Fetch finished auctions

### LINKED ISSUE #33: Display finished auction
### Goal
An auction should be available by its link even after it was finished

### Context
Currently, we only fetch and display current auctions – auctions that are active or ones that need a restart. That results in a problem that as soon as an auction is taken by someone, our link to it (`/collateral?network=kovan&auction=BAL-A%3A`) is broken and currently shows `This auction was not found`, which is misleading.

In this issue we need to first investigate different ways to fetch auctions by id (without a need cache them ourselves). Otherwise:
- Subscribe/fetch clipper events (`Kick`, `Take`, `Redo`)
- Enrich event data to the minimal `Auction` type (as an event only contains `(index_topic_1 uint256 id, uint256 top, uint256 tab, uint256 lot, index_topic_2 address usr, index_topic_3 address kpr, uint256 coin)`)
- Store auctions in a database (setup database service)
- Update frontend auction page to fetch the auction from db (if the auction is not found or maybe always)

### Todo
- [ ] Investigate, write down a conclusion
- [ ] Implement caching or direct fetching

benediktkaiser:
closes #33


benediktkaiser:
@valiafetisov I have a question. 

When it comes to calling the events. I followed the guidelines outlined [here](https://docs.ethers.io/v5/concepts/events/).

I got this far:
```js
const abi = ['event Take(uint256,uint256,uint256,address,bytes)'];
new Contract(address, abi, provider);
```

But I am not sure how to call and filter it now. Could you give me a pointer? Thanks!

valiafetisov:
[You already have an ABI for the clipper contract](https://github.com/sidestream-tech/unified-auctions-ui/blob/e50fa92f3a6e5a7c763f0793076ad55faca4b2f0/core/src/abis/MCD_CLIP.json#L190-L238). You just need to `getContract` and use its methods. I guess something like `contract.filters.Take(encodedId)` should work.

core/src/helpers/parseAuctionURL.ts
```diff
@@ -0,0 +1,12 @@
+import convertNumberTo32Bytes from './convertNumberTo32Bytes';
+
+export function parseAuctionURL(url: string): { collateralType: string; auctionID: string } {
+    const parts = url.split(':', 1);
+    const collateralType = parts[0];
+    const auctionID = convertNumberTo32Bytes(parts[1]);
```
valiafetisov, line 6: - Please rename `auctionID` to https://github.com/sidestream-tech/unified-auctions-ui/blob/e50fa92f3a6e5a7c763f0793076ad55faca4b2f0/core/src/types.ts#L6
- Please don't use `convertNumberTo32Bytes` within here, otherwise this function wouldn't be reusable
- Please validate parts (collateral type via getCollateralConfig and `auctionId` to be number)

core/src/helpers/convertNumberTo32Bytes.ts
```diff
@@ -1,6 +1,6 @@
 import { ethers } from 'ethers';
 
-const convertNumberTo32Bytes = function (number: number): string {
+const convertNumberTo32Bytes = function (number: number | string): string {
```
valiafetisov, line 3: Please don't change the type

core/src/auctions.ts
```diff
@@ -127,6 +130,20 @@ export const fetchAllAuctions = async function (network: string): Promise<Auctio
...
+    const addresses: Record<string, any> = await fetchContractsAddressesByNetwork(network);
+    const address: string = addresses[`MCD_CLIP_${collateralType}`];
+    const provider = await getProvider(network);
+
+    if (!address) {
+        throw new Error(`No Clip contract could be found for "${collateralType}"`);
+    }
+
+    const abi = ['event Take(uint256,uint256,uint256,address,bytes)'];
+    new Contract(address, abi, provider);
```
valiafetisov, line 144:
```suggestion
    const contractName = getClipperNameByCollateralType(auction.collateralType);
    const contract = await getContract(network, contractName);
```

core/src/auctions.ts
```diff
@@ -12,10 +12,13 @@ import {
...
     calculateTransactionProfitDate,
 } from './price';
-import { getSupportedCollateralTypes } from './addresses';
+import { fetchContractsAddressesByNetwork, getSupportedCollateralTypes } from './addresses';
 import { getClipperNameByCollateralType } from './contracts';
 import convertNumberTo32Bytes from './helpers/convertNumberTo32Bytes';
 import { enrichAuctionWithTransactionFees, getApproximateTransactionFees } from './fees';
+import getProvider from './provider';
+import { Contract } from 'ethers';
+import { parseAuctionURL } from './helpers/parseAuctionURL';
```
valiafetisov, line 21: If a file only exports one function, please make a default export

valiafetisov:
I guess, you just created an event filter. But now you need to fetch the events (which suppose to be 1) that match this filter.
```
let eventFilters = [contract.filters.Take(encodedId)];
let events = await contract.queryFilter(eventFilters);
```
https://docs.ethers.io/v5/api/contract/contract/#Contract-queryFilter

Please also note that your second `topics` seems invalid: `encodedId` should not be all 0

benediktkaiser:
```js
[
    {
        "blockNumber": 24649045,
        "blockHash": "0x2c32a806a13002acc78f079ae29699f646b142210cb3cee9f72a4ecef7a0a333",
        "transactionIndex": 2,
        "removed": false,
        "address": "0x7dD1Fb6b9aFdBA9F28DB89c81723b8c6B27A2Fbe",
        "data": "0x000000000000000000000000000000000000001c4446edb759ac4a7d3aa1d2ce000000000000000000000000000000000000001c4446edb759ac4a7d3aa1d2ce00000000000000000000000023084f676940b784e08bf994edc2026dec9056f20000000000000000000000001a4e76e9ad3c0aace758051ca056fa25467a1c1d000000000000000000000000000000000000000000000000011875ddd04cd8a3",
        "topics": [
            "0x05e309fd6ce72f2ab888a20056bb4210df08daed86f21f95053deb19964d86b1",
            "0x0000000000000000000000000000000000000000000000000000000000000002",
            "0x000000000000000000000000c140ce1be1c0eda2f06319d984c404251c59494e"
        ],
        "transactionHash": "0xc455dce21c3eacfb134a273a82ff70e46d9df23632e7f8d37bae7a3c1a332bab",
        "logIndex": 14,
        "event": "Take",
        "eventSignature": "Take(uint256,uint256,uint256,uint256,uint256,uint256,address)",
        "args": [
            {
                "type": "BigNumber",
                "hex": "0x02"
            },
            {
                "type": "BigNumber",
                "hex": "0x1c4446edb759ac4a7d3aa1d2ce"
            },
            {
                "type": "BigNumber",
                "hex": "0x1c4446edb759ac4a7d3aa1d2ce"
            },
            {
                "type": "BigNumber",
                "hex": "0x23084f676940b784e08bf994edc2026dec9056f2"
            },
            {
                "type": "BigNumber",
                "hex": "0x1a4e76e9ad3c0aace758051ca056fa25467a1c1d"
            },
            {
                "type": "BigNumber",
                "hex": "0x011875ddd04cd8a3"
            },
            "0xC140ce1be1c0edA2f06319d984c404251C59494e"
        ]
    },
    {
        "blockNumber": 24649069,
        "blockHash": "0x9e099dc440167095d7083c88a8a080d58221b5c34d2a10c1d8f1f224116382ae",
        "transactionIndex": 5,
        "removed": false,
        "address": "0x7dD1Fb6b9aFdBA9F28DB89c81723b8c6B27A2Fbe",
        "data": "0x000000000000000000000000000000000000001bfbea0f2dd6375e396aaf9530000000000000000000000000000000000000001bfbea0f2dd6375e396aaf95300000000000000000000000001a4e76e9ad3c0aace758051ca056fa25467a1c1d00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000027ce80fbd02ac5",
        "topics": [
            "0x05e309fd6ce72f2ab888a20056bb4210df08daed86f21f95053deb19964d86b1",
            "0x0000000000000000000000000000000000000000000000000000000000000002",
            "0x000000000000000000000000c140ce1be1c0eda2f06319d984c404251c59494e"
        ],
        "transactionHash": "0xa1db0ec6765e172ab11ed0668f81457337515a6d11dbf0fe86585fe708fd0a5c",
        "logIndex": 20,
        "event": "Take",
        "eventSignature": "Take(uint256,uint256,uint256,uint256,uint256,uint256,address)",
        "args": [
            {
                "type": "BigNumber",
                "hex": "0x02"
            },
            {
                "type": "BigNumber",
                "hex": "0x1bfbea0f2dd6375e396aaf9530"
            },
            {
                "type": "BigNumber",
                "hex": "0x1bfbea0f2dd6375e396aaf9530"
            },
            {
                "type": "BigNumber",
                "hex": "0x1a4e76e9ad3c0aace758051ca056fa25467a1c1d"
            },
            {
                "type": "BigNumber",
                "hex": "0x00"
            },
            {
                "type": "BigNumber",
                "hex": "0x27ce80fbd02ac5"
            },
            "0xC140ce1be1c0edA2f06319d984c404251C59494e"
        ]
    }
]
```

I got this back as a response. It looks good and now I just need to format it!

benediktkaiser:
The format of the Take is as follows: 

```js
 Take (index_topic_1 uint256 id, uint256 max, uint256 price, uint256 owe, uint256 tab, uint256 lot, index_topic_2 address usr)
```

I can use this to determine what the parameters of the Auction are.

valiafetisov:
So there were two take transactions with seemingly the same auction id with two minutes difference. I am not sure I understand what's happening there (maybe internal auction counter was somehow reset?):
https://kovan.etherscan.io/tx/0xa1db0ec6765e172ab11ed0668f81457337515a6d11dbf0fe86585fe708fd0a5c
https://kovan.etherscan.io/tx/0xc455dce21c3eacfb134a273a82ff70e46d9df23632e7f8d37bae7a3c1a332bab

Can you try to fetch old auctions for other collateral types? Or do it on the mainnet?

core/src/helpers/parseAuctionURL.ts
```diff
@@ -0,0 +1,17 @@
...
+    const parts = url.split(':');
+    const collateralType = parts[0];
+    const auctionId = parts[1];
+
+    // Check if we support the CollateralType
+    getCollateralConfigByType(collateralType);
+
+    return {
+        collateralType,
+        auctionId: parseInt(auctionId),
```
valiafetisov, line 13: Please check it's not `Number.isNaN()`, otherwise throw an error

frontend/store/auctions.ts
```diff
@@ -227,4 +228,8 @@ export const actions = {
         const network = rootGetters['network/getMakerNetwork'];
         await checkAllSupportedCollaterals(network);
     },
+    async fetchFinishedAuction({ rootGetters }: ActionContext<State, State>) {
+        const network = rootGetters['network/getMakerNetwork'];
+        await fetchFinishedAuction(network, 'ETH-A:2');
```
valiafetisov, line 233: Please use the actual auction id 🙂

benediktkaiser:
> So there were two take transactions with seemingly the same auction id with two minutes difference. I am not sure I understand what's happening there (maybe internal auction counter was somehow reset?): https://kovan.etherscan.io/tx/0xa1db0ec6765e172ab11ed0668f81457337515a6d11dbf0fe86585fe708fd0a5c https://kovan.etherscan.io/tx/0xc455dce21c3eacfb134a273a82ff70e46d9df23632e7f8d37bae7a3c1a332bab
> 
> Can you try to fetch old auctions for other collateral types? Or do it on the mainnet?

I tried this on my hardhat network and I don't get two responses there! I assume it was an issue with that one auction!

benediktkaiser:
I managed to get it to a stable state, however the numbers are sadly not correct at all. I will post two screenshots, the first being a screen of the UI after I executed the auction (so the correct data), and the second being what I received from the event. 

<details>
  <summary>Correct Action</summary>
  
Screenshot 2022-02-11 at 12 31 52

</details>

<details>
  <summary>Auction taken from event</summary>
  
Screenshot 2022-02-11 at 12 31 40

</details>

As you can see many details don't line up at all and `lot` is even set to 0 (which is why it says Auction Amount is under 0.00001). @valiafetisov am I doing some of the conversion wrong or is the data I am retrieving just incorrect?

valiafetisov:
Yeah, it may be that they emit state of the contract ["after the execution"](https://github.com/makerdao/dss/blob/c666ab1fdac4cb3dd8a8b4223f951a9773a64c55/src/clip.sol#L415) that's why `tab` and other numbers are zero. Have you tried calling `fetchAuctionByCollateralTypeAndAuctionId` at the block of the take event (or if it doesn't work, at the block before that)? [According to the ethers docs](https://docs.ethers.io/v5/api/contract/contract/#:~:text=report%20meaningful%20values-,overrides.blockTag,-%2D%20a%20block%20tag), you can pass block number as a `blockTag`. But as discussed before, this may not work with our infura key.

benediktkaiser:
> Yeah, it may be that they emit state of the contract ["after the execution"](https://github.com/makerdao/dss/blob/c666ab1fdac4cb3dd8a8b4223f951a9773a64c55/src/clip.sol#L415) that's why `tab` and other numbers are zero. Have you tried calling `fetchAuctionByCollateralTypeAndAuctionId` at the block of the take event (or if it doesn't work, at the block before that)? [According to the ethers docs](https://docs.ethers.io/v5/api/contract/contract/#:~:text=report%20meaningful%20values-,overrides.blockTag,-%2D%20a%20block%20tag), you can pass block number as a `blockTag`. But as discussed before, this may not work with our infura key.

I have not tested this yet. It would make sense to look into to see if it works!

valiafetisov:
> this may not work with our infura key.

That's the correct assumption: infura provides historical data only as a paid feature (250$/mo). Therefore, my suggestion here would be to change the issue to fetch all related `take` events to this auction id and display them nicely with links to etherscan. Eg, in the end it can look like this:
```
This auction was taken at ${takeEvents[2].date.toISOString()} via 2 transactions:
- Transaction 0x000...0002
- Transaction 0x000...0001
```

core/src/auctions.ts
```diff
@@ -131,61 +130,15 @@ export const fetchAllAuctions = async function (network: string): Promise<Auctio
...
-        authTransactionFeeETH: new BigNumber(0),
-        authTransactionFeeDAI: new BigNumber(0),
-        restartTransactionFeeETH: new BigNumber(0),
-        transactionProfitMinusFees: new BigNumber(0),
-        approximateUnitPrice: new BigNumber(0),
-    };
-};
-
-export const fetchFinishedAuction = async function (network: string, auctionURL: string): Promise<AuctionTransaction> {
+export const fetchFinishedAuction = async function (network: string, auctionURL: string): Promise<Array<Event>> {
```
valiafetisov, line 133: Let's call it `fetchTakeEvents`

frontend/store/auctions.ts
```diff
@@ -52,6 +55,9 @@ export const getters = {
     getAuctionById: (state: State) => (id: string) => {
         return state.auctionStorage[id];
     },
+    getEventsByID: (state: State) => (id: string) => {
+        return state.eventStorage[id];
+    },
```
valiafetisov, line 60: Let's call it `takeEventsStorage` since we may have more types of events in the future or account for them here:
```suggestion
    getTakeEventsByID: (state: State) => (id: string) => {
        return state.eventStorage[id];
    },
```

benediktkaiser:
## Finished Auction view: 

Screenshot 2022-03-08 at 09 46 21

## How to test:
For anyone testing the system I advise to view the following Auction as a test.
http://localhost:3000/collateral?network=kovan&auction=ETH-A%3A1

It has already been executed and contains multiple transactions in its Event history.

frontend/store/auctions.ts
```diff
@@ -22,7 +21,7 @@ let updateAuctionsPricesIntervalId: ReturnType<typeof setInterval> | undefined;
 
 interface State {
     auctionStorage: Record<string, AuctionTransaction>;
-    eventStorage: Record<string, Event[]>;
+    takeEventStorage: Record<string, Event[]>;
```
valiafetisov, line 24: What is the `Event` type, where did you get it? I suggest that you define our own type and store only things we need, ie transactionHash at the moment (and maybe a block number)

frontend/components/Auction.vue
```diff
@@ -305,7 +317,10 @@ export default Vue.extend({
     computed: {
         errorText(): string | null {
             if (!this.isAuctionsLoading && !this.auction) {
-                this.$emit('fetchFinishedAuction', this.auctionId);
+                this.$emit('fetchTakeEventsFromAuction', this.auctionId);
```
valiafetisov, line 320: I am not sure that it's optimal to emit something from the computed property. Are you sure it's emitted once and not every now and then?

frontend/components/Auction.vue
```diff
@@ -305,7 +317,10 @@ export default Vue.extend({
...
+                if (this.takeEvents) {
+                    return 'This auction is finished';
+                }
                 return 'This auction was not found';
```
valiafetisov, line 324: Does that mean in the moment when there is no `this.takeEvents` yet, `This auction was not found` will be displayed?

frontend/components/Auction.vue
```diff
@@ -233,6 +233,14 @@
                 </div>
             </TextBlock>
         </div>
+        <div v-if="takeEvents" class="mt-5">
+            <p class="mb-2">The auction was taken via {{ takeEvents.length }} transactions.</p>
```
valiafetisov, line 237:
```suggestion
            <p class="mb-2">The auction was taken via {{ takeEvents.length }} transactions:</p>
```

Please also handle `s` in the `transactions` in case `length === 1`

core/src/date.ts
```diff
@@ -4,13 +4,18 @@ import { getNetworkConfigByType } from './constants/NETWORKS';
...
 
-const fetchLatestBlockDate = async function (network: string): Promise<Date> {
+export const fetchDateByBlockNumber = async function (network: string, blockNumber: number): Promise<Date> {
     const provider = await getProvider(network);
-    const blockNumber = await provider.getBlockNumber();
     const block = await provider.getBlock(blockNumber);
     return new Date(block.timestamp * 1000);
 };
 
+const fetchLatestBlockDate = async function (network: string): Promise<Date> {
```
benediktkaiser, line 13: Do we still want this function, or should I remove it?

frontend/store/auctions.ts
```diff
@@ -237,12 +239,10 @@ export const actions = {
         const network = rootGetters['network/getMakerNetwork'];
         await checkAllSupportedCollaterals(network);
     },
-    async fetchFinishedAuction({ rootGetters, commit }: ActionContext<State, State>, url: string) {
+    async fetchTakeEventsFromAuction({ rootGetters, commit }: ActionContext<State, State>, url: string) {
```
valiafetisov, line 242:
```suggestion
    async fetchTakeEventsFromAuction({ rootGetters, commit }: ActionContext<State, State>, auctionId: string) {
```

frontend/store/auctions.ts
benediktkaiser: I get it directly from `ether.js`. 
https://docs.ethers.io/v5/api/contract/contract/#Contract-queryFilter

It is also the type that `contract.queryFilter` returns, which is why I would continue using it.

frontend/components/Auction.vue
benediktkaiser: Currently during all my tests, the loading animation is set until all the events have been retrieved. Therefore `This auction was not found` was not been displayed to me in my testing. 

However I understand where you are coming from. Would it make sense to introduce a new prop called `eventsLoading` and to check that?

frontend/components/Auction.vue
benediktkaiser: Sure!

frontend/store/auctions.ts
valiafetisov: As suggested, we may not want to put the whole `Event` object into the store 🙂

frontend/components/Auction.vue
valiafetisov: > Currently during all my tests, the loading animation is set until all the events have been retrieved

This is not in any way confirmed by the code I see. `isAuctionsLoading` is only set inside the action querying auctions on the page load. It is not set inside events querying. If you testing it on the hardhat I imaging the loading times are very short to notice it. `isEventsLoading` would be ok, a better solution would be `isAuctionLoadingStore: Record<string, boolean>`

core/src/date.ts
valiafetisov: `fetchLatestBlockDate` is used in other places, you just refactored it into two functions, which is ok to keep.

Also, [the UI proposal](https://github.com/sidestream-tech/unified-auctions-ui/pull/90#issuecomment-1048898985) actually have `at ${takeEvents[takeEvents.length - 1].date.toISOString()}` which you may want to add to the text as a bonus (but then you would need to fetch date for all events and change `Event` type to our custom `AuctionEvent` which contains the date)

LukSteib:
I have tested it for the latest auction tonight on mainnet and looks good (see screenshot below)
http://localhost:3000/collateral?network=mainnet&auction=ETH-A%3A427

image

What I still don't get is the difference in transactions displayed. Why do we sometimes display only one transaction (see example above) and sometimes up to 7 like in the example posted by @benediktkaiser? Isn't there one transaction that is associated to the take event and this is the one we should display?

valiafetisov:
> Isn't there one transaction that is associated to the take event and this is the one we should display?

I was also [initially confused](https://github.com/sidestream-tech/unified-auctions-ui/pull/90#issuecomment-1034626504), but there are several transactions due to the partial bidding. We can, of course, display only the last one, but that one may not match the auction data we show in the tweet, when the auction starts and in general be misleading.

LukSteib:
> but there are several transactions due to the partial bidding

Ah I see. Makes sense if the several transactions refer to the partial bidding.

core/src/date.ts
benediktkaiser: Added this dates!

benediktkaiser: I personally prefer displaying dates with `toDateString()` as I believe it is more readable. 
Is there a reason you wanted to use `.toISOString()` or can I use `toDateString()`?

frontend/store/auctions.ts
benediktkaiser: I made a smaller version of the object!

frontend/components/Auction.vue
benediktkaiser: I moved it to a watcher!

core/src/auctions.ts
```diff
@@ -128,6 +130,34 @@ export const fetchAllAuctions = async function (network: string): Promise<Auctio
...
+    const date = await fetchDateByBlockNumber(network, takeEvent.blockNumber);
+    return {
+        ...takeEvent,
+        date,
+    };
+};
+
+export const fetchTakeEvents = async function (network: string, auctionURL: string): Promise<Array<TakeEvent>> {
+    const { collateralType, auctionId } = parseAuctionURL(auctionURL);
+    const encodedAuctionId = convertNumberTo32Bytes(auctionId);
```
valiafetisov, line 143: I think there is a confusion between `auction.id`, `auction.auctionId` and `auctionId` that is revealed in this PR. I would suggest to rename type `auction.auctionId` to `auction.index` and then use those as equals: `auction.id` = `auctionId`, `auction.index` = `auctionIndex`. That would make many places less confusing:
```suggestion
    const { collateralType, auctionIndex } = parseAuctionURL(auctionId);
    const encodedAuctionIndex = convertNumberTo32Bytes(auctionIndex);
```
Would you think it makes sense to do it here or in another PR (merged before this one)?

core/src/auctions.ts
```diff
@@ -128,6 +130,34 @@ export const fetchAllAuctions = async function (network: string): Promise<Auctio
...
+const enrichTakeEventWithDate = async function (network: string, takeEvent: TakeEvent): Promise<TakeEvent> {
+    const date = await fetchDateByBlockNumber(network, takeEvent.blockNumber);
+    return {
+        ...takeEvent,
+        date,
+    };
+};
+
+export const fetchTakeEvents = async function (network: string, auctionURL: string): Promise<Array<TakeEvent>> {
+    const { collateralType, auctionId } = parseAuctionURL(auctionURL);
```
valiafetisov, line 142: So `parseAuctionURL` (which can later be renamed to `parseAuctionId`) can throw an error. What happens then?

Hint: please validate that opening url with invalid parameter `INVALID` shows usual "no action found"

frontend/store/auctions.ts
```diff
@@ -227,4 +247,13 @@ export const actions = {
...
         await checkAllSupportedCollaterals(network);
     },
+    async fetchTakeEventsFromAuction({ rootGetters, commit }: ActionContext<State, State>, auctionId: string) {
+        commit('setAreTakeEventsFetching', true);
+
+        const network = rootGetters['network/getMakerNetwork'];
+
+        const events = await fetchTakeEvents(network, auctionId);
+        commit('setTakeEvents', { id: auctionId, events });
+        commit('setAreTakeEventsFetching', false);
```
valiafetisov, line 257: Related to the comment above: this should be called within `finally`

core/src/types.ts
```diff
@@ -112,3 +112,9 @@ export declare interface MessageContent {
 export declare interface Notifier {
     (messageType: 'loading' | 'error' | 'success' | 'info', messageContent: MessageContent): void;
 }
+
+export declare interface TakeEvent {
+    transactionHash: string;
+    blockNumber: number;
+    date?: Date;
```
valiafetisov, line 119:
```suggestion
    transactionDate?: Date;
```

frontend/components/Auction.vue
```diff
@@ -233,12 +233,31 @@
...
+                The auction was taken via {{ takeEvents.length }} transaction<span v-if="takeEvents.length !== 1"
+                    >s</span
+                >
+                at {{ takeEvents[takeEvents.length - 1].date.toISOString() }}.
+            </p>
+            <ul class="list-disc list-inside">
+                <li v-for="(event, index) of takeEvents" :key="index">
+                    Transaction <FormatAddress :value="event.transactionHash" :shorten="true" /> (<TimeTill
+                        :date="event.date"
+                    />)
```
valiafetisov, line 247: I would imagine it can be something like `Transaction 0x000..000 executed 2h ago`
Also, since `transactionDate` can be undefined, I suppose this should be handled somehow

frontend/components/Auction.vue
```diff
@@ -331,6 +358,14 @@ export default Vue.extend({
...
             },
         },
+        fetchTakeEvents: {
+            immediate: true,
+            handler() {
+                if (!this.areAuctionsFetching && !this.auction) {
+                    this.$emit('fetchTakeEventsFromAuction', this.auctionId);
+                }
+            },
+        },
```
valiafetisov, line 368: This will fetch events no matter if auction is found or not, on the component load and never again, since there is no prop or data called `fetchTakeEvents`. Instead, you should watch for the specific prop:
```suggestion
        areAuctionsFetching(areAuctionsFetching) {
            if (!areAuctionsFetching && !this.auction) {
                this.$emit('fetchTakeEventsFromAuction', this.auctionId);
            }
        },
```

core/src/auctions.ts
benediktkaiser: I would do it in a separate PR. It seems to be a bigger refracting job.

core/src/auctions.ts
benediktkaiser: I added a try catch that console logs the error if an invalid id is passed. The UI will then only show `No Auction found`.

I can also confirm that Auctions that do not exist are displayed with `No Auction found` as well.

valiafetisov: I mean code-wise there is a difference between:
- auctions that doesn't exist, ie inexistent auction index [`ETH-A:9000`](http://localhost:3000/collateral?network=mainnet&auction=ETH-A%3A9000)
- invalid auction index [`ETH-A:INVALID`](http://localhost:3000/collateral?network=mainnet&auction=ETH-A%3AINVALID)
- inexistent collateral type [`INVALID:1`](http://localhost:3000/collateral?network=mainnet&auction=INVALID%3A1)
- invalid auction id [`INVALID`](http://localhost:3000/collateral?network=mainnet&auction=INVALID)

Please test all edge cases by opening frontend with those urls

core/src/auctions.ts
valiafetisov: Then please make another PR, so we can merge it before this PR is merged!

core/src/auctions.ts
benediktkaiser: > * auctions that doesn't exist, ie inexistent auction index [`ETH-A:9000`](http://localhost:3000/collateral?network=mainnet&auction=ETH-A%3A9000)

> * invalid auction index [`ETH-A:INVALID`](http://localhost:3000/collateral?network=mainnet&auction=ETH-A%3AINVALID)

> * inexistent collateral type [`INVALID:1`](http://localhost:3000/collateral?network=mainnet&auction=INVALID%3A1)

> * invalid auction id [`INVALID`](http://localhost:3000/collateral?network=mainnet&auction=INVALID)

All of them display `Auction Not Found`

core/src/auctions.ts
benediktkaiser: I just committed it in here directly!

frontend/components/AuctionEventsBlock.vue
```diff
@@ -0,0 +1,32 @@
...
+import FormatAddress from '~/components/utils/FormatAddress.vue';
+import TimeTill from '~/components/common/TimeTill.vue';
+
+export default Vue.extend({
+    name: 'AuctionEventsBlock',
+    components: { FormatAddress, TimeTill },
+    props: {
+        takeEvents: {
+            type: Array as Vue.PropType<TakeEvent[]>,
+            default: null,
```
valiafetisov, line 28: If the default is null, you need to add checks to the template, since `null.length` will throw
```suggestion
            required: true,
```

frontend/components/MainFlow.vue
```diff
@@ -70,7 +74,15 @@ export default Vue.extend({
             type: Array as Vue.PropType<AuctionTransaction[]>,
             default: () => [],
         },
-        isAuctionsLoading: {
+        takeEventStorage: {
+            type: Object as Vue.PropType<string, TakeEvent[]>,
+            default: () => {},
```
valiafetisov, line 79: This will result in `undefined`. Have you meant that?
```suggestion
            default: () => ({}),
```

frontend/helpers/generateFakeAuction.js
```diff
@@ -24,7 +24,7 @@ export const generateFakeAuction = function () {
     return {
         network: 'stub',
         id: `${collateralObject.ilk}:${auctionId}`,
-        auctionId,
+        index: auctionId,
```
valiafetisov, line 27:
```suggestion
        index,
```

frontend/components/AuctionEventsBlock.vue
```diff
@@ -0,0 +1,32 @@
+<template>
+    <div class="mt-5">
+        <p class="mb-2">
+            The auction was taken via {{ takeEvents.length }} transaction<span v-if="takeEvents.length !== 1">s</span>
+            at {{ takeEvents[takeEvents.length - 1].transactionDate.toISOString() }}.
+        </p>
+        <ul class="list-disc list-inside">
+            <li v-for="(event, index) of takeEvents" :key="index">
```
valiafetisov, line 8: Providing something unique from the data itself is always better than array index:
```suggestion
            <li v-for="event of takeEvents" :key="event.transactionHash">
```

core/src/auctions.ts
```diff
@@ -128,14 +130,42 @@ export const fetchAllAuctions = async function (network: string): Promise<Auctio
...
+const enrichTakeEventWithDate = async function (network: string, takeEvent: TakeEvent): Promise<TakeEvent> {
+    const date = await fetchDateByBlockNumber(network, takeEvent.blockNumber);
+    return {
+        ...takeEvent,
+        transactionDate: date,
+    };
+};
+
+export const fetchTakeEvents = async function (network: string, auctionURL: string): Promise<Array<TakeEvent>> {
+    const { collateralType, index } = parseAuctionId(auctionURL);
```
valiafetisov, line 142:
```suggestion
export const fetchTakeEvents = async function (network: string, auctionId: string): Promise<Array<TakeEvent>> {
    const { collateralType, index } = parseAuctionId(auctionId);
```

core/src/fetch.ts
```diff
@@ -45,17 +45,17 @@ export const fetchAuctionStatus = async function (
 const fetchAuctionByCollateralTypeAndAuctionId = async function (
     network: string,
     collateralType: string,
-    auctionId: number
+    auctionIndex: number
```
valiafetisov, line 48: the function should be called `fetchAuctionByCollateralTypeAndAuctionIndex`

LukSteib (changes requested):
- [ ] NIT: We might format this ISO based date string (e.g. `2022-03-07T21:27:31.000Z`) to a more user friendly representation (e.g. similar to how we do it for `Auction Start` in the `AuctionTable`)
(only if it's done easily and quick - should not block us from getting this one over the finishing line)

benediktkaiser:
> * NIT: We might format this ISO based date string (e.g. `2022-03-07T21:27:31.000Z`) to a more user friendly representation (e.g. similar to how we do it for `Auction Start` in the `AuctionTable`)
>   (only if it's done easily and quick - should not block us from getting this one over the finishing line)

There is no simple built in function that allows us to display the date with the time. I could make a new function for it if you want.

valiafetisov:
Also, right before displaying `This auction is finished` it displays `This auction is inactive and must be restarted` for some reason:

https://user-images.githubusercontent.com/3393626/157486355-28dc82eb-edb6-408b-b6bd-d6c32dd1002d.mov

valiafetisov:
- [ ] Auction table stopped displaying ids (let's rename it to `Index` and display `auction.index`)
Screenshot 2022-03-09 at 17 45 43

valiafetisov:
One other error that appeared in this PR ([confirmed that staging doesn't have it](https://main.auction-ui.k8s.sidestream.tech/collateral)):

- [ ] Frontend is broken with `No network found with name "undefined"` when network mismatch dialog is shown:

Screenshot 2022-03-09 at 17 49 52

The error is originated from `FormatAddress` components `etherscanURL` computed prop where we use `getNetworkConfigByType`. To solve, the `getNetworkConfigByType` call should be try/catched. I assume that the error is not there on staging since we haven't used `FormatAddress` on the main page before

valiafetisov (approved):

LukSteib (approved):

PR #92, opened 2022-02-09
WstETH collateral support

### LINKED ISSUE #73: Add `wstETH` collateral support

valiafetisov:
Closes #73 

Screenshot 2022-02-09 at 10 04 41

Note for the reviewer: only testable via hardhat

### DIFF
--- a/core/package-lock.json
+++ b/core/package-lock.json
@@ -22,6 +22,8 @@
         "@types/uuid": "^8.3.3",
         "@typescript-eslint/eslint-plugin": "^4.33.0",
         "@typescript-eslint/parser": "^4.33.0",
+        "@uniswap/sdk-core": "^3.0.1",
+        "@uniswap/v3-sdk": "^3.8.2",
         "dotenv": "^10.0.0",
         "eslint": "^7.8.0",
         "eslint-config-prettier": "^6.11.0",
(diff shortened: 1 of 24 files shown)


LukSteib (approved):
Was able to 
* auth `wstETH` transactions ✔️ 
* execute `wstETH` auction ✔️ 

---
Two questions:
* Do we make use of the dynamic routing functionality of Uniswap V3? 
* Do the transaction fees reflect gas prices and gas usage for tx execution properly on the hardhat fork?

valiafetisov:
> Do we make use of the dynamic routing functionality of Uniswap V3?

No, the exchange is done directly ETH->DAI, the callee itself also doesn't do any routing.

> Do the transaction fees reflect gas prices and gas usage for tx execution properly on the hardhat fork?

No, ethers.js can not figure it out itself and we so far only used one endpoint to determine mainnet gas price, which is unrelated here. Therefore, I configured the gas price to be very high: same way as on other test chains, but higher.

benediktkaiser (approved):

PR #105, opened 2022-02-15
Features page

### LINKED ISSUE #101: Implement a /features page
**Goal:**
Provide a page to users that describes all of the features of our tool(s) 

**Context:**
We agreed to create a feature page that lists all of the functionality of our tools in bullet lists. This should help a user that wants to inspect further what actions are faciliated via our UI (and which are not).
For that we have to create a dedicated subpage `/features` that allows to provide this overview

**Tasks:**
- [x] use existing landingblock with background image and heading `Feature overview` 
- [x] Use existing header without only with dark mode switch
   - [x] Maker Logo with text `Collateral auctions` that links to `/collateral` page (similar to /dashboard header)  
- [x] path: auctions.makerdao.network/features
- [x] Include the text (provided in gdoc below) and structure it nicely (correct spacing etc.)
   - [x] **note:** the comment on T&C at the end of the document is out of scope. Everything else should be included. 
- [x] Ensure responsivenes
- [x] Include a link to our footer called `Features` that links to this newly created page 

**Assets:**
Gdoc: https://docs.google.com/document/d/13pEGxUQlJ45kkCNCFATZRs8zw8mFJINmHx2MqAmqTGo/edit#

fegbert:
closes #101 

**preview**

image


frontend/components/FeatureList.vue
```diff
@@ -0,0 +1,280 @@
...
+import Vue from 'vue';
+import TextBlock from '~/components/common/TextBlock.vue';
+
+export default Vue.extend({
+    components: {
+        TextBlock,
+    },
+    data() {
+        return {
+            features: [
```
benediktkaiser, line 33: I think this is debatable and maybe you put some thought into this. 
As this feature list is pretty long I was wondering if it might make more sense to split it off into a separate file and to then export the const. Might make editing texts/UI easier in the future as we managed to separate it. 

What do you think?

frontend/components/FeatureList.vue
```diff
@@ -0,0 +1,280 @@
...
+                        {
+                            text: `The auction keeper is an open source tool provided by us for download`,
+                            indent: 0,
+                        },
+                        {
+                            text: `The keeper is not operated by us for or on behalf of the user`,
+                            indent: 0,
+                        },
+                        {
+                            text: `A user is able to download the code and run it himself`,
```
benediktkaiser, line 247: This is more a comment for @florianstoecker. 

I believe we should be using gender neutral language. Therefore I would replace `himself` with `themself`.

benediktkaiser: Had a quick talk with Flo about this. He agrees it is good to change this immediately!

benediktkaiser (changes requested):

frontend/components/FeatureList.vue
fegbert: I actually thought the same thing but doubted myself 😅 The main reason I used this file structure instead of just writing the text in the template is to improve readability of the component, so I agree that moving the text outside of it would increase that even more.

frontend/components/FeatureList.vue
fegbert: Very nice catch!

benediktkaiser (dismissed):

frontend/components/layout/Header.vue
```diff
@@ -72,7 +75,7 @@ export default Vue.extend({
         type: {
             type: String,
             default: 'default',
-            validator: type => ['default', 'unified'].includes(type),
+            validator: type => ['default', 'unified', 'features'].includes(type),
```
valiafetisov, line 78: Can you reuse `unified` type or is it in some way different?

frontend/components/FeatureList.vue
```diff
@@ -0,0 +1,54 @@
...
+                    <a v-if="feature.url" :href="feature.url"> {{ feature.url }} </a>
+                    <span v-else class="italic">Not yet implemented</span>
+                </div>
+                <h3 class="text-lg text-gray-700 font-semibold dark:text-gray-200 my-2">List of features:</h3>
+                <ul
+                    v-for="(point, index) in feature.features"
+                    :key="index"
+                    class="list-inside list-disc text-gray-700 dark:text-gray-200"
+                >
+                    <li :class="`Indent-${point.indent}`">{{ point.text }}</li>
```
valiafetisov, line 16: - This is not correct from the semantic point of view. The person with a screen reader would have a degraded possibility to understand the lists. But to implement it properly, you would need to 1) put `ul` into a separate component 2) [provide name of the component](https://github.com/sidestream-tech/unified-auctions-ui/blob/dd6882ad6cb84ee54d95b8e5bd933a31ec61a702/frontend/components/Auction.vue#L257) 3) call it recursively (you may want to restructure the FeatureList as well)
- Second reason to refactor this: class names constructed dynamically will be removed in production build https://stackoverflow.com/a/69687876

frontend/data/FeatureTexts.ts
```diff
@@ -0,0 +1,230 @@
...
+                indent: 1,
+            },
+            {
+                text: `For the following metrics computations are being made from the tool based on the fetched input parameters`,
+                indent: 1,
+            },
+            {
+                text: `Time till auction ends`,
+                indent: 2,
+            },
```
valiafetisov, line 28: For example: 
```suggestion
            {
                text: `The tool enables the user to see detailed metrics of collateral auctions`,
                items: [
                    {
                        text: `Basic auction information is fetched from the blockchain via the Infura API and
                        displayed in a structured way in the UI`,
                    },
                    {
                        text: `For the following metrics computations are being made from the tool based on the fetched input parameters`,
                        items: [
                            {
                                text: `Time till auction ends`,
                            },
                        ],
                    },
                ],
            },
```

frontend/components/FeatureList.vue
```diff
@@ -0,0 +1,54 @@
...
+<style scoped>
+.Indent-1 {
+    margin-left: 25px;
+    list-style-type: circle;
+}
+
+.Indent-2 {
+    margin-left: 50px;
+    list-style-type: square;
+}
```
valiafetisov, line 49: If you do this recursively, you can use default styling (which already suppose to change `list-style-type` and apply double `margin-left`) or use `ul > li > ul` css selector

frontend/components/layout/Header.vue
fegbert: using `minimal` from privacy page now 🙂

LukSteib (approved):

PR #106, opened 2022-02-16
Terms and Conditions consent with cookies

### LINKED ISSUE #79: Discuss and implement Terms and Condition consent implementation
**Goal:**
Define and implement the most appropriate user flow to give consent to our terms and conditions

**Context:**
As Is:
* The user is prompted with T&C consent popup when trying to connect wallet for the first time
* We currently store T&C consent in local storage for unlimited time once user has given consent

In the future we want to change that in order to being able to make adjustments to the T&S statement.
Discussed alternatives:
* A1: User has to agree to T&C on wallet connection for every new session 
* A2: User has to agree to T&C on wallet connection when we release a new version of T&S statement (implies to have T&C versioning)
* A3: User has to agree to T&C on wallet connection and we store consent with cookie for period of X days before consent expires

**Tasks:**
- [x] Discuss different possible approaches and decide
- [x] Implement approach to store user consent to T&C in session cookie
   - [x] see: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie
   - [x] meaning: With each new browser session the user is provided with T&C popup when connecting wallet to UI

fegbert:
closes #79  

> User has to agree to T&C on wallet connection for every new session
> With each new browser session the user is provided with T&C popup when connecting wallet to UI

expected behavior:

- when connecting wallet for the first time -> ask for consent
- after that -> don't ask again. Not on new tab or second browser window
- when closing & reopening browser / using incognito -> ask again. 

For setting cookies I am using [vue-cookies](https://www.npmjs.com/package/vue-cookies)

### DIFF
--- a/frontend/layouts/default.vue
+++ b/frontend/layouts/default.vue
@@ -7,7 +7,7 @@
             :dark-mode.sync="isDarkMode"
             :wallet-address="walletAddress"
             :is-wallet-loading="isWalletLoading"
-            :has-accepted-terms="hasAcceptedTerms()"
+            :has-accepted-terms="hasAcceptedTerms"
             :staging-banner-url="stagingBannerURL"
             @changeWalletType="changeWalletType"
             @openTermsModal="setTermsModal(true)"
@@ -90,12 +90,15 @@ export default Vue.extend({
         stagingBannerURL() {
             return process.env.STAGING_BANNER_URL;
         },
+        hasAcceptedTerms(): boolean {
+            return this.$store.getters['cookies/hasAcceptedTerms'];
+        },
     },
     methods: {
         ...mapActions('network', ['setPageNetwork', 'fixWalletNetwork']),
         ...mapActions('wallet', ['changeWalletType']),
         acceptTerms(): void {
-            this.$cookies.set('terms-and-conditions', 'accepted', 0);
+            this.$store.dispatch('cookies/acceptTerms');
             this.$store.commit('modals/setTermsModal', false);
             this.$store.commit('modals/setWalletModal', true);
         },
@@ -104,9 +107,6 @@ export default Vue.extend({
         },
         setWalletModal(open: boolean): void {
             this.$store.commit('modals/setWalletModal', open);
-        },
-        hasAcceptedTerms(): boolean {
-            return this.$cookies.get('terms-and-conditions') === 'accepted';
         },
     },
 });
--- a/frontend/layouts/default.vue
+++ b/frontend/layouts/default.vue
@@ -98,7 +98,7 @@ export default Vue.extend({
         ...mapActions('network', ['setPageNetwork', 'fixWalletNetwork']),
         ...mapActions('wallet', ['changeWalletType']),
         acceptTerms(): void {
-            this.$store.dispatch('cookies/acceptTerms');
+            this.$store.commit('cookies/acceptTerms');
             this.$store.commit('modals/setTermsModal', false);
             this.$store.commit('modals/setWalletModal', true);
         },
--- a/frontend/layouts/default.vue
+++ b/frontend/layouts/default.vue
@@ -7,7 +7,7 @@
             :dark-mode.sync="isDarkMode"
             :wallet-address="walletAddress"
             :is-wallet-loading="isWalletLoading"
-            :has-accepted-terms="hasAcceptedTerms"
+            :has-accepted-terms="hasAcceptedTerms()"
             :staging-banner-url="stagingBannerURL"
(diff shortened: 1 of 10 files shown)


florianstoecker:
I've also tested the behavior with (sadly) following outcome:

When completely closing the browser(chrome) & opening it again & then visiting the UI, which I was connected to before I closed the browser, my Wallet immediately connects to the UI without ever asking for consent.

fegbert:
> [ ] restarted browser, popup was triggered

Discussed in today's daily: If someone has the "reopen last tabs" feature enabled, chrome / chromium-based browsers seem to save session cookies, so this is the expected behavior here from a technical pov. For the legal side, @LukSteib thinks it is fine as well, so we can keep this in.

frontend/layouts/default.vue
```diff
@@ -108,6 +105,9 @@ export default Vue.extend({
         setWalletModal(open: boolean): void {
             this.$store.commit('modals/setWalletModal', open);
         },
+        hasAcceptedTerms(): boolean {
+            return this.$cookies.get('terms-and-conditions') === 'accepted';
+        },
```
valiafetisov, line 110: Why did you change it to be a method instead of a computed function? Making it a method means that it wouldn't be reactive.

valiafetisov (commented):
One other way to implement it would be to move `acceptedTerms` to a new store (eg `cookie`) and use the same `vuex-persist` but use js-cookie as a storage mechanism, they have an example of that: https://www.npmjs.com/package/vuex-persist#detailed

frontend/layouts/default.vue
fegbert: Setting the cookie didn't seem to automatically update the computed property, so I had to reload the for it to change. That's why I changed it to check on button press with a method.

Is there maybe a way to "force" an update?

fegbert:
> One other way to implement it would be to move acceptedTerms to a new store (eg cookie) and use the same vuex-persist but use js-cookie as a storage mechanism

Do you think this could have any advantages over using `vue-cookies`?

valiafetisov:
> Do you think this could have any advantages over using `vue-cookies`?

Yes, you wouldn't need to change any UI and reactivity would work as expected.

frontend/layouts/default.vue
valiafetisov: Forcing an update is not a correct solution, [the variable should be reactive in the first place for the update to happen](https://github.com/tofandel/vue-cookies#readme). See my comment below.

valiafetisov:
So, to be more specific, here we define how we preserve our `preferences` vuex store into the `localStorage`: 

https://github.com/sidestream-tech/unified-auctions-ui/blob/47fb9f41684431b1e48305a193bc9f9fa8861e4f/frontend/plugins/vuex-persist.client.js#L5-L8

In order to preserve `cookies` vuex store in cookies, you can add right below those lines something like:

```js
new VuexPersistence({
    restoreState: (key, storage) => Cookies.getJSON(key),
    saveState: (key, state, storage) => Cookies.set(key, state),
    modules: ['cookies'],
}).plugin(store);
```

And then treat `cookies` store like a regular store. Would that work?

valiafetisov (approved):

LukSteib (changes requested):
Chrome testing:
- [X] Popup triggered when first connecting to the site
- [X] Opened page in new tab, popup was not triggered
- [X] Opened Page incognito, popup was triggered
- [ ] restarted browser, popup was triggered -> popup is not triggered (AFAIK I do not have the chrome based open last tabs feature enabled)
- [ ] also I tried to check whether the cookie is set when using incognito mode but couldn't see it (see recording below). Don't now if this is the right approach to check whether the cookie is set properly



https://www.loom.com/share/90d6b3841bbb40f996a73f839149e5b3

LukSteib:
>  also I tried to check whether the cookie is set when using incognito mode but couldn't see it (see recording below). Don't now if this is the right approach to check whether the cookie is set properly

Ah - I see. We are taking the local storage approach similar to explanation preference. Have a feeling that this is not session based though. On browser restart terms acceptance is still set to true.

fegbert:
Hmm... not sure what's going on there.
We're not using the local storage for this anymore. There should be a cookie called `vuex`

image

Which has a value `acceptedTerms` which is set to true when accepting the terms & conditions.

image

LukSteib (approved):
Installing new dependencies before testing the new "feature" sometimes helps. 
Sorry for the confusion. 

Behavior now as expected and hence approved from my side!

PR #119, opened 2022-02-28
AccessDaiBlock Component

### LINKED ISSUE #115: Implement `AccessDaiBlock` component
### Goal

Implemented `AccessDaiBlock` component in the storybook

### Context

Similar to the `AuthorisationBlock` we would need to have component titles as `Allow access to DAI`, with one button (which is active or inactive depending on the `isDaiAccessGranted` and `disabled`)

### Assets

Component disabled:
MakerDAO@2x (1)

Component with `isDaiAccessGranted` true:
MakerDAO@2x (5)

The context:
MakerDAO - Bid with DAI (total)@2x

### Tasks
- [x] create new `AccessDaiBlock` inside `transaction` folder
- [x] please check that `isExplanationsShown` is taken into the account
- [x] please check that `isLoading` is taken into the account
- [x] create a story

fegbert:
closes #115

**default**

image

**disabled**

image

**loading**

image

**access granted**

image

**expert mode**

image

## Questions:

- In the mock, the `disabled` state looks different than the look we use for other buttons. Is this intentional & should I change it to look like the mock, or should I keep the style we are currently using?
- Should the button be disabled when access is already granted? If not, should it have an action / emit an event?

### DIFF
--- a/frontend/components/transaction/AccessDaiBlock.vue
+++ b/frontend/components/transaction/AccessDaiBlock.vue
@@ -1,0 +1,55 @@
+<template>
+    <div>
+        <TextBlock title="Allow access to DAI">
+            This action allows you to deposit DAI into the
+            <Explain text="VAT">
+                The
+                <a
+                    href="https://docs.makerdao.com/smart-contract-modules/core-module/vat-detailed-documentation#2-contract-details"
+                >
+                    VAT contract
+                </a>
+                is the core vault engine of the Maker Protocol and manages the central accounting invariants of DAI.
+                Depositing and interacting with the VAT is necessary in order to participate in auctions.
+            </Explain>
+            . The following transaction authorizes the wallet address to deposit into the VAT. It is a prerequisite to
+            participate in the auction.
+        </TextBlock>
+        <div class="flex flex-row-reverse mt-3">
+            <BaseButton
+                v-if="!isDaiAccessGranted"
+                type="primary"
+                :disabled="disabled"
+                @click="$emit('grantDaiAccess')"
+            >
+                Allow access to DAI
+            </BaseButton>
+            <BaseButton v-else>Access to Dai was granted</BaseButton>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import TextBlock from '~/components/common/TextBlock.vue';
+import BaseButton from '~/components/common/BaseButton.vue';
+import Explain from '~/components/utils/Explain.vue';
+
+export default Vue.extend({
+    components: {
+        TextBlock,
+        BaseButton,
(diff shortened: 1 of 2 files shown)


valiafetisov:
> Is this intentional & should I change it to look like the mock, or should I stay with the "normal" style (as seen in the preview above)?

It should be similar to the other auth steps, it's just a whimsical limitation, the looks on the screenshots are good!

> Should the button be disabled when access is already granted? If not, should it have an action / emit an event?

Good question. I think for now the state should disabled, similar to the other auth steps.

LukSteib:
> Good question. I think for now the state should disabled, similar to the other auth steps. We may add revert operation since this allow access [is more dangerous than other auth steps](https://user-images.githubusercontent.com/3393626/151387065-860a982e-2635-481c-bb40-c8444d37c4fb.png). @LukSteib, what do you think?

I would stick to the usual auth step for now and disable the button once auth was done. 
I see your point, though, of potentially providing an option to revert. However this has the downsides of 
a) tx fees for each revert and auth cycle
b) risk of a using reverting while still having dai deposited (which would require an other auth to withdraw ...) 
-> Hence we would definitely think about this flow more thoroughly

frontend/components/transaction/AccessDaiBlock.vue
```diff
@@ -0,0 +1,64 @@
...
+            default: false,
+        },
+        isLoading: {
+            type: Boolean,
+            default: false,
+        },
+        isDaiAccessGranted: {
+            type: Boolean,
+            required: false,
+        },
```
valiafetisov, line 57: The prop should either be required or the default value should be provided, I think this one is critical for the component to work:
```suggestion
        isDaiAccessGranted: {
            type: Boolean,
            required: true,
        },
```

Also, can you please add this lint rule and validate that it works?
https://eslint.vuejs.org/rules/require-default-prop.html

valiafetisov (approved):

frontend/components/transaction/AccessDaiBlock.vue
fegbert: This is actually already set (to warn), but it doesn't count `boolean` props.

> This rule requires default value to be set for each props that are not marked as `required` (except `Boolean` props).

I've looked for another rule to add the check for booleans but I have only found one which does the exact opposite (not allow default values) 🙁

valiafetisov (approved):

LukSteib (changes requested):
- [ ] Can you please use the text provided here https://github.com/sidestream-tech/unified-auctions-ui/issues/115#issuecomment-1054094572

LukSteib (approved):

valiafetisov (approved):

PR #122, opened 2022-02-28
BidBlock Component

### LINKED ISSUE #118: Implement `BidBlock` component
### Goal

Implemented `BidBlock` component in the storybook

### Context

Similar to other transaction blocks, we will need to have component `BidBlock` titled as `Bid`, with one button `Bid XXX.XX DAI for XXX.XX ${collateralSymbol}`.

### Assets

MakerDAO@2x (11)

### Tasks
- [x] create new `BidBlock` inside `transaction` folder
- [x] please check that `isExplanationsShown` is taken into the account
- [x] please check that `isLoading` is taken into the account
- [x] please check that `disabled` is taken into the account
- [x] create a story for all possible states

fegbert:
closes #118 

**default**

image

**disabled**

image

**loading**

image

**expert mode**

image

### DIFF
--- a/frontend/components/transaction/BidBlock.vue
+++ b/frontend/components/transaction/BidBlock.vue
@@ -1,0 +1,58 @@
+<template>
+    <div>
+        <TextBlock v-if="isExplanationsShown" title="Made a bid">
+            Auction participation incurs transaction fees (paid in ETH). Hence the connected wallet needs to hold
+            enough funds to cover these fees. Whenever a transaction that incurs such fees needs to be processed a
+            confirmation from within the wallet is necessary and will be prompted.
+        </TextBlock>
+        <div class="flex flex-row-reverse mt-3">
+            <BaseButton :disabled="disabled" :is-loading="isLoading" type="primary">
+                <span
+                    >Bid <format-currency :value="totalPrice" currency="DAI" /> for
+                    <format-currency :value="collateralAmount" :currency="collateralSymbol"
+                /></span>
+            </BaseButton>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import TextBlock from '~/components/common/TextBlock.vue';
+import BaseButton from '~/components/common/BaseButton.vue';
+import FormatCurrency from '~/components/utils/FormatCurrency.vue';
+
+export default Vue.extend({
+    components: {
+        TextBlock,
+        BaseButton,
+        FormatCurrency,
+    },
+    props: {
+        isExplanationsShown: {
+            type: Boolean,
+            default: true,
+        },
+        isLoading: {
+            type: Boolean,
+            default: false,
+        },
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+        totalPrice: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: undefined,
+        },
+        collateralAmount: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: undefined,
+        },
+        collateralSymbol: {
+            type: String,
+            default: '',
+        },
+    },
+});
+</script>
--- a/frontend/components/transaction/BidBlock.stories.js
+++ b/frontend/components/transaction/BidBlock.stories.js
@@ -1,0 +1,55 @@
+import { storiesOf } from '@storybook/vue';
+import BidBlock from './BidBlock';
+import { generateFakeAuction } from '~/helpers/generateFakeAuction';
+
+const fakeAuction = generateFakeAuction();
+
+const common = {
+    components: { BidBlock },
+    data() {
+        return {
+            auction: fakeAuction,
+        };
+    },
+};
+
+storiesOf('Transaction/BidBlock', module)
+    .add('Default', () => ({
+        ...common,
+        template: `
+    <BidBlock 
+        :totalPrice="auction.totalPrice"
+        :collateralAmount="auction.collateralAmount"
+        :collateralSymbol="auction.collateralSymbol"
+    />`,
+    }))
+    .add('Disabled', () => ({
+        ...common,
+        template: `
+    <BidBlock 
+        :totalPrice="auction.totalPrice"
+        :collateralAmount="auction.collateralAmount"
+        :collateralSymbol="auction.collateralSymbol"
+        disabled
+    />`,
+    }))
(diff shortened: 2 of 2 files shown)


frontend/components/transaction/BidBlock.vue
```diff
@@ -0,0 +1,58 @@
...
+        },
+        isLoading: {
+            type: Boolean,
+            default: false,
+        },
+        disabled: {
+            type: Boolean,
+            default: false,
+        },
+        totalPrice: {
```
valiafetisov, line 44: this should be `transactionAmountDAI`

frontend/components/transaction/BidBlock.vue
```diff
@@ -0,0 +1,58 @@
...
+        <TextBlock v-if="isExplanationsShown" title="Made a bid">
+            Auction participation incurs transaction fees (paid in ETH). Hence the connected wallet needs to hold
+            enough funds to cover these fees. Whenever a transaction that incurs such fees needs to be processed a
+            confirmation from within the wallet is necessary and will be prompted.
+        </TextBlock>
+        <div class="flex flex-row-reverse mt-3">
+            <BaseButton :disabled="disabled" :is-loading="isLoading" type="primary">
+                <span>
+                    Bid <format-currency :value="totalPrice" currency="DAI" /> for
+                    <format-currency :value="collateralAmount" :currency="collateralSymbol" />
```
valiafetisov, line 12: Let's show `Bidding...` while `isLoading`

frontend/components/transaction/BidBlock.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+    <div>
+        <TextBlock v-if="isExplanationsShown" title="Made a bid">
+            Auction participation incurs transaction fees (paid in ETH). Hence the connected wallet needs to hold
+            enough funds to cover these fees. Whenever a transaction that incurs such fees needs to be processed a
+            confirmation from within the wallet is necessary and will be prompted.
```
valiafetisov, line 6: Please use text from `ExecutionBlock`: 
> Auction bidding incurs transaction fee of approximately 0.001 ETH, hence, the connected wallet needs to hold enough funds to cover these fees. The transaction fee is a recommended value and based on the average fast fee. The amount can be edited by the participant to influence the speed of the transaction.

(making transactionFee as a prop)

valiafetisov (commented):
Please also accept transaction hash, just like `ExecutionBlock` to support finished state. I would certainly accept `AuctionTransaction` object here

frontend/components/transaction/BidBlock.vue
valiafetisov: `totalPrice` is the wrong thing to show

fegbert:
> Please also accept transaction hash, just like ExecutionBlock to support finished state

Should I also show something for the finished state then?

valiafetisov:
Yes, the same thing as in the Execution block. Note: when you copy a code without changes (like here, finished auction transaction message) you can just split it into a separate component and use it in several places.

frontend/components/transaction/BidBlock.vue
```diff
@@ -0,0 +1,81 @@
...
+        state(): string {
+            if (this.isLoading) {
+                return 'loading';
+            }
+            if (this.disabled) {
+                return 'disabled';
+            }
+            if (this.auctionTransaction.transactionProfit?.isLessThan(0)) {
+                return 'notProfitable';
+            }
```
valiafetisov, line 73: There is no profit in this transaction and therefore no `notProfitable` state

frontend/components/transaction/BidBlock.vue
```diff
@@ -0,0 +1,81 @@
+<template>
+    <div>
+        <TransactionMessage
+            :is-explanations-shown="isExplanationsShown"
+            :is-finished="state === 'executed'"
+            :transaction-address="auctionTransaction.transactionAddress"
```
valiafetisov, line 6: If `transactionAddress` exists that means it's `isFinished`, therefore, you can remove this extra prop

frontend/components/transaction/ExecutionBlock.vue
```diff
@@ -64,23 +45,17 @@
 import Vue from 'vue';
 import BigNumber from 'bignumber.js';
 import { Alert } from 'ant-design-vue';
-import TextBlock from '~/components/common/TextBlock.vue';
 import BaseButton from '~/components/common/BaseButton.vue';
-import FormatAddress from '~/components/utils/FormatAddress.vue';
-import FormatCurrency from '~/components/utils/FormatCurrency.vue';
-import Explain from '~/components/utils/Explain.vue';
 import ExecuteWithOtherWalletBlock from '~/components/transaction/ExecuteWithOtherWalletBlock.vue';
+import TransactionMessage from '~/components/transaction/TransactionMessage.vue';
```
valiafetisov, line 50: Can you please remove `transactionURL` from the `ExecutionBlock.vue` component? it's unrelated to your PR, but unused

valiafetisov (dismissed):

frontend/components/transaction/TransactionMessage.vue
```diff
@@ -0,0 +1,54 @@
...
+            <Explain text="average fast fee">
+                <a href="https://ethereum.org/en/developers/docs/gas/#why-can-gas-fees-get-so-high" target="_blank">
+                    Gas Prices can change
+                </a>
+                due to changing busyness of the Ethereum network. However, you can look up the current average gas
+                price
+                <a href="https://ethgasstation.info/" target="_blank"> here </a> </Explain
+            >. The amount can be edited by the participant to influence the speed of the transaction. <br /><br />
+            By default, the profit from a successful bidding will be sent to the same wallet which executed the
+            transaction. However, you can provide a different address which will receive the transaction profit.
```
LukSteib, line 20: - [ ] This part of the message should not be shown for the new `BidBlock` as it does not apply to the Dai bidding flow but only to the swapping based flow that we have implemented in the past. (so please remove here)

LukSteib (changes requested):
* Do we ensure that the button size increases dynamically even when numbers displayed can get very large?

fegbert:
> Do we ensure that the button size increases dynamically even when numbers displayed can get very large?

Yes, as long as there's space on the page the button can grow with bigger numbers 🙂

valiafetisov (approved):

PR #129, opened 2022-03-07
BidTransactionTable Component

### LINKED ISSUE #120: Implement `BidTransactionTable` component
### Goal

Implemented `BidTransactionTable` component in the storybook

### Context

Similar to the `SwapTransactionTable`, `BidTransactionTable` is a table showing key values of the auction. Besides showing the values the table offers interactive elements (marked with green color in the UI):
- an input to enter bid amount (defaults to total price _changing_  with every drop)
- a button to set bid amount to minimum
- a button to set bid amount to total price
- reset button (if the bid amount was changed) has the same functionality as `set bid amount to total price` above

Please note that additional complexity is in the details:
- the input should be focused on load
  - we allow use to input add any amount of floating point numbers (and always display them without formatting)
  - we disallow user to enter other symbols
  - when the value is `total`, we format it with animated number (ie with 2 decimals)
- because of the above, bidding value that is equal to total is stored as something else than a number
   proposal: `export type transactionAmountDAI = BigNumber | 'total'`
- editing is validated and the error message is shown in a non-closable tooltip with the red bg
  - we validate that the value is not less than new `auction.minimumBid` value
  - we validate that the value is not more than `auction.totalPrice - auction.minimumBid`
- total value is changing due to the dropping price, but if the user has edited it, it always stays with the edited value (but the validation is reactive, ie a computed prop)

Notes:
- it may be beneficial to move complex input logic to a separate component, but may also not (: 
- it may be beneficial to write validation logic as part of the `core`, since we would need it in other places as well

### Assets
MakerDAO@2x (15)

### Tasks
- [x] create new `BidTransactionTable` inside `transaction` folder
- [x] please handle `isFinished` and `isActive` states properly (disabling/cleaning up input)
- [x] create a story for all possible states
  - [x] implement a story that periodically drops price

fegbert:
closes #120 

**preview**

### DIFF
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -78,7 +78,7 @@
         </div>
         <div class="flex justify-between items-center">
             <div>The amount to bid</div>
-            <div class="w-2/5">
+            <div class="w-1/2">
                 <bid-input
                     v-model="amountToBid"
                     :minimum-deposit-dai="minimumDepositDai"
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -55,15 +55,15 @@
         >
             <div class="underline">Auction total price</div>
             <div>
-                <Popover
-                    v-if="!auctionTransaction.isActive && !auctionTransaction.isFinished"
-                    placement="topRight"
-                    content="Since the auction is not active, there is no total Auction Price for this auction."
-                    trigger="hover"
-                >
-                    <span class="opacity-50">Unknown</span>
-                </Popover>
-                <format-currency v-else :value="auctionTransaction.totalPrice" currency="DAI" />
+                <span v-if="!auctionTransaction.isActive && !auctionTransaction.isFinished" class="opacity-50">
+                    Unknown
+                </span>
+                <format-currency
+                    v-else-if="auctionTransaction.totalPrice"
+                    :value="auctionTransaction.totalPrice"
+                    currency="DAI"
+                />
+                <span v-else class="opacity-50">Unknown</span>
             </div>
         </div>
         <div
@@ -107,7 +107,6 @@
 <script lang="ts">
 import Vue from 'vue';
 import BigNumber from 'bignumber.js';
-import { Popover } from 'ant-design-vue';
 import BidInput from '../utils/BidInput.vue';
 import TimeTill from '~/components/common/TimeTill.vue';
 import FormatCurrency from '~/components/utils/FormatCurrency.vue';
@@ -119,7 +118,6 @@ export default Vue.extend({
         FormatCurrency,
         PriceDropAnimation,
         BidInput,
-        Popover,
     },
     props: {
         auctionTransaction: {
@@ -139,7 +137,7 @@ export default Vue.extend({
     computed: {
         amountToReceive(): BigNumber | undefined {
             if (!this.amountToBid || !this.auctionTransaction.approximateUnitPrice) {
-                return this.auctionTransaction.totalPrice;
+                return this.auctionTransaction.totalPrice || undefined;
             }
             return this.amountToBid.dividedBy(this.auctionTransaction.approximateUnitPrice);
         },
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -1,0 +1,88 @@
+<template>
+    <div class="flex flex-col space-y-1">
+        <div class="flex justify-between">
+            <div>Auction Ends</div>
+            <div>
+                <time-till :date="auctionTransaction.endDate" />
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>Auction Amount</div>
+            <div>
+                <FormatCurrency
+                    :value="auctionTransaction.collateralAmount"
+                    :currency="auctionTransaction.collateralSymbol"
+                />
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>Auction Price</div>
+            <div>
+                <PriceDropAnimation :auction="auctionTransaction" class="mr-1" />
+                <FormatCurrency :value="auctionTransaction.approximateUnitPrice" currency="DAI" /> per
+                <span class="uppercase">{{ auctionTransaction.collateralSymbol }}</span>
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>
+                Transaction Fee
+                <span class="text-gray-300"
+                    >(~<FormatCurrency
+                        v-if="auctionTransaction.biddingTransactionFeeETH"
+                        :value="auctionTransaction.biddingTransactionFeeETH"
+                        :decimals="5"
+                    />
+                    <span v-else>Unknown</span>
+                    ETH)</span
+                >
+            </div>
+            <div>
+                <FormatCurrency
+                    v-if="auctionTransaction.biddingTransactionFeeDAI"
+                    :value="auctionTransaction.biddingTransactionFeeDAI * -1"
+                    currency="DAI"
+                />
+                <span v-else class="opacity-50">Unknown</span>
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>Auction total price</div>
+            <div>
+                <format-currency :value="auctionTransaction.totalPrice" currency="DAI" />
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>Auction minimum bid</div>
+            <div></div>
+        </div>
+        <div class="flex justify-between">
+            <div>The amount to bid</div>
+            <div></div>
+        </div>
+        <div class="flex justify-between font-bold">
+            <div>The amount to receive</div>
+            <div></div>
+        </div>
+    </div>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import TimeTill from '~/components/common/TimeTill.vue';
+import FormatCurrency from '~/components/utils/FormatCurrency.vue';
+import PriceDropAnimation from '~/components/utils/PriceDropAnimation.vue';
+
+export default Vue.extend({
+    components: {
+        TimeTill,
+        FormatCurrency,
+        PriceDropAnimation,
+    },
+    props: {
+        auctionTransaction: {
+            type: Object as Vue.PropType<AuctionTransaction>,
+            required: true,
+        },
+    },
+});
+</script>
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -144,7 +144,7 @@ export default Vue.extend({
             return this.auctionTransaction.isActive && !this.auctionTransaction.isFinished;
         },
         isBidAmountNaN(): boolean {
-            return this.amountToBid?.isNaN() || this.amountToBid?.isEqualTo(0) || false;
+            return this.amountToBid?.isNaN() || false;
         },
     },
     methods: {
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -52,7 +52,7 @@
         <div
             class="flex justify-between"
             :class="{ ClickableText: isActive && auctionTransaction.totalPrice }"
-            @click="setAmountToBid(auctionTransaction.totalPrice)"
+            @click="setAmountToBid(undefined)"
         >
             <div :class="{ underline: isActive && auctionTransaction.totalPrice }">Auction total price</div>
             <div>
@@ -90,7 +90,7 @@
             <div>The amount to receive</div>
             <div>
                 <format-currency
-                    v-if="amountToReceive && isActive"
+                    v-if="amountToReceive && isActive && !isBidAmountNaN"
                     :value="amountToReceive"
                     :currency="auctionTransaction.collateralSymbol"
                 />
@@ -125,7 +125,7 @@ export default Vue.extend({
         },
         minimumDepositDai: {
             type: Object as Vue.PropType<BigNumber>,
-            default: null,
+            required: true,
         },
     },
     data() {
@@ -142,6 +142,9 @@ export default Vue.extend({
         },
         isActive(): boolean {
             return this.auctionTransaction.isActive && !this.auctionTransaction.isFinished;
+        },
+        isBidAmountNaN(): boolean {
+            return this.amountToBid?.isNaN() || false;
         },
     },
     methods: {
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -137,7 +137,7 @@ export default Vue.extend({
     computed: {
         amountToReceive(): BigNumber | undefined {
             if (!this.amountToBid || !this.auctionTransaction.approximateUnitPrice) {
-                return this.auctionTransaction.totalPrice || undefined;
+                return this.auctionTransaction.collateralAmount;
             }
             return this.amountToBid.dividedBy(this.auctionTransaction.approximateUnitPrice);
         },
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -49,11 +49,7 @@
                 <span v-else class="opacity-50">Unknown</span>
             </div>
         </div>
-        <div
-            class="flex justify-between"
-            :class="{ ClickableText: isActive && auctionTransaction.totalPrice }"
-            @click="setAmountToBid(undefined)"
-        >
+        <div class="flex justify-between">
             <button
                 :class="{ 'ClickableText underline': isActive && auctionTransaction.totalPrice }"
                 @click="setAmountToBid(undefined)"
--- a/frontend/components/transaction/BidTransactionTable.vue
+++ b/frontend/components/transaction/BidTransactionTable.vue
@@ -51,13 +51,15 @@
         </div>
         <div class="flex justify-between">
             <button
-                :class="{ 'ClickableText underline': isActive && auctionTransaction.totalPrice }"
+                class="ClickableText"
+                :disabled="!isActive || !auctionTransaction.totalPrice"
                 @click="setAmountToBid(undefined)"
             >
                 Buy all collateral
             </button>
             <button
-                :class="{ ClickableText: isActive && auctionTransaction.totalPrice }"
+                class="ClickableText"
+                :disabled="!isActive || !auctionTransaction.totalPrice"
                 @click="setAmountToBid(undefined)"
             >
                 <format-currency
@@ -70,13 +72,15 @@
         </div>
         <div class="flex justify-between">
             <button
-                :class="{ 'ClickableText underline': isActive && minimumDepositDai }"
+                class="ClickableText"
+                :disabled="!isActive || !minimumDepositDai"
                 @click="setAmountToBid(minimumDepositDai)"
             >
                 Set minimum bid
             </button>
             <button
-                :class="{ ClickableText: isActive && minimumDepositDai }"
+                class="ClickableText"
+                :disabled="!isActive || !minimumDepositDai"
                 @click="setAmountToBid(minimumDepositDai)"
             >
                 <format-currency v-if="minimumDepositDai && isActive" :value="minimumDepositDai" currency="DAI" />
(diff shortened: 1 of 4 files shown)


fegbert:
## What still needs to be done

- [ ]  Add `bid amount` input field

- all relevant infos are in the issue

Notes:
- The input should show the `totalPrice` as long as the user hasn't input anything. As always this should be done with an `animatedNumber`. This of course can't be done within the input field, so we can just overlay it and hide it when the user wants to input something. (it's fine to clear it on click, no need to show the total value inside the input field 😄)
- For filtering, we can limit what the user can type with a pattern. So numbers and dots I think. Then we have to validate that it's a valid `BigNumber` and if it's in the valid range. If it is the total value we can change the value back to `total` which triggers the overlayed `animatedNumber`. For an empty input we can show 0.

- [ ] Make the `Auction Total Price` and `Auction minimum bet` clickable and change the input value.
- [ ] Calculate `Amount to receive`

LukSteib (changes requested):
Firs review notes:

- [ ] The following error case outlined in initial issue is not covered properly:
> we validate that the value is not more than auction.totalPrice - auction.minimumBid
- [ ] Why do we show figures for `minimum bid` and `amount to receive` for `finished` and `inactive` auction states? This might be confusing to the user (proposal: go with `unknown` there as well)
- [ ] In `Finished` and `Inactive` states I would prevent editing possibility for Input field and selection rows
- [ ] In the Decreasing Price case we do not cover the case that the user makes a vaild input but then over time the total price falls below the user input -> once this happens we need to show an error as well

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,146 @@
...
+                    v-if="auctionTransaction.biddingTransactionFeeDAI"
+                    :value="auctionTransaction.biddingTransactionFeeDAI * -1"
+                    currency="DAI"
+                />
+                <span v-else class="opacity-50">Unknown</span>
+            </div>
+        </div>
+        <div
+            class="flex justify-between text-primary hover:text-primary-light cursor-pointer"
+            @click="amountToBid = auctionTransaction.totalPrice"
```
valiafetisov: This action should be the same as "resetting", meaning that it should go down with price drops
```suggestion
            @click="amountToBid = undefined"
```

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,146 @@
...
+    },
+    props: {
+        auctionTransaction: {
+            type: Object as Vue.PropType<AuctionTransaction>,
+            required: true,
+        },
+        minimumDepositDai: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: null,
+        },
```
valiafetisov, line 129: Please make it required (you also never check if it's null)
```suggestion
        minimumDepositDai: {
            type: Object as Vue.PropType<BigNumber>,
            required: true,
        },
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,137 @@
...
+    components: { FormatCurrency, Tooltip },
+    props: {
+        totalPrice: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: null,
+        },
+        minimumDepositDai: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: null,
+        },
```
valiafetisov, line 39: Same here

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,137 @@
...
+    },
+    methods: {
+        validateInput(newVal: string, oldVal: string): void {
+            this.error = undefined;
+            if (!newVal) {
+                return;
+            }
+            const value = new BigNumber(newVal);
+            if (isNaN(value.toNumber())) {
```
valiafetisov, line 95: You avoid using `isNaN` since it's an old-school function which converts type to number before checking it against `NaN` (eg `isNaN(null)` is false). Instead you should prefer newer `Number.isNaN` which does the same but expect number as input.

Moreover, `value` is `BigNumber` therefore it has its own method :
```suggestion
            if (value.isNaN()) {
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,137 @@
...
+                this.userInput = oldVal;
+            } else if (value.isGreaterThan(this.totalPrice)) {
+                this.error = `The bidding amount can not be greater than ${this.totalPrice.toFixed(2)} DAI`;
```
valiafetisov: I would not mix error logic with user input logic. Error logic can be refactored to a separate computed property

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,137 @@
...
+            } else if (value.isLessThan(this.minimumDepositDai)) {
+                this.error = `The bidding amount can not be smaller than ${this.minimumDepositDai.toFixed(2)} DAI`;
```
valiafetisov, line 100: Right now, when the error is shown, we still show the `The amount to receive` value:

Screenshot 2022-03-08 at 17 32 52

Which is confusing for the user who don't know that the _old_ value is used. The mock also indicated that we should show `Unknown` instead. 

Implementation-wise it's a bit complicated since you decided to use `undefined` value as `total` and didn't left an option to store `Unknown`. That's why in the issue description I proposed to define it as  composite type `BigNumber | 'total'` (or rather `undefined | BigNumber | 'total'`). This of course is tricky since you will need to check the against the total string everywhere. Other (potentially better) option would be to set `value` to `NaN` during errors and then handle `NaN` in the parent as well.

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,149 @@
...
+        },
+        hideTotalPrice(): void {
+            if (!this.value) {
+                this.setValue(new BigNumber(0));
+            }
+        },
+        validateFinished() {
+            if (!this.userInput) {
+                this.userInput = '0';
+            }
```
valiafetisov, line 113: Currently, focusing the input and then loosing the focus will result in 0 being set to the input. I would expect that it should return back to total

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,149 @@
...
+            if (!this.value || this.value.isEqualTo(this.totalPrice)) {
+                this.resetToTotal();
+            }
+        },
+        resetToTotal() {
+            this.userInput = undefined;
+            this.setValue(undefined);
```
valiafetisov, line 120: I would expect that `setValue(undefined)` would already reset `userInput = undefined` why is it not the case?

fegbert: setting `userInput` when setting `value` would format set the input to the formatted number which we don't want, but for `undefined` this is fine / always the case anyways you are right

valiafetisov (commented):
Currently it's very hard to read the code of the input and validate its logic, because so many places emit values up and almost each function has side effects. I believe it can be simplified to:
- `value` prop (or better be called `amountToBid.sync`) which contains `undefined | BigNumber | NaN`
- `userInput` data value (or better be called `amountToBidInputString`) which always contains string or empty string
- `amountToBidInputParsed` computed prop that converts `amountToBidInputString` to BigNumber or _undefined_
- `error` computed prop (or better be called `errorMessage`) that checks if `amountToBidInputParsed` is within limits
- `amountToBidInputParsed` watcher that emits itself as `amountToBid` if no `this.errorMessage` is present
- a method to set `amountToBid` to 0 on `focus` if `amountToBidInputString` is empty
- a method to set `amountToBid` to undefined on `blur` if `amountToBidInputString` is empty

I believe everything else can go.

Also, you can actually name the prop `amountToBid` and [make it `.sync`](https://v2.vuejs.org/v2/guide/components-custom-events.html?redirect=true#sync-Modifier), so it's easier to understand

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+            }
+            const maximumBid = this.totalPrice.minus(this.minimumDepositDai);
+            let error = undefined as string | undefined;
+            if (this.amountToBidInputParsed.isGreaterThan(maximumBid)) {
+                error = `The bidding amount can not be greater than ${maximumBid.toFixed(2)} DAI`;
+            }
+            if (this.amountToBidInputParsed.isLessThan(this.minimumDepositDai)) {
+                error = `The bidding amount can not be smaller than ${this.minimumDepositDai.toFixed(2)} DAI`;
+            }
+            return error;
```
valiafetisov, line 74: I don't think you need to have this variable anymore
```suggestion
            if (this.amountToBidInputParsed.isGreaterThan(maximumBid)) {
                return `The bidding amount can not be greater than ${maximumBid.toFixed(2)} DAI`;
            }
            if (this.amountToBidInputParsed.isLessThan(this.minimumDepositDai)) {
                return `The bidding amount can not be smaller than ${this.minimumDepositDai.toFixed(2)} DAI`;
            }
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+    data() {
+        return {
+            amountToBidInput: '' as string,
+        };
+    },
+    computed: {
+        amountToBidInputParsed(): BigNumber | undefined {
+            if (!this.amountToBidInput) {
+                return undefined;
+            }
```
valiafetisov, line 58: We may also want to remove spaces to be user friendly:
```suggestion
            if (this.amountToBidInput.trim() === '') {
                return undefined;
            }
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+        },
+        amountToBid(newVal: BigNumber | undefined) {
+            if (!newVal) {
+                this.amountToBidInput = '';
+                return;
+            }
+            if (newVal.isEqualTo(this.minimumDepositDai)) {
+                this.amountToBidInput = newVal.toString();
+            }
```
valiafetisov, line 98: Can't you just always set this?

```suggestion
            this.amountToBidInput = newVal.toFixed();
```

Also, `toString` can produce scientific notation like `2.3e-5`, as opposite to `toFixed`: https://mikemcl.github.io/bignumber.js/#toS

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+        },
+    },
+    methods: {
+        hideTotalPrice(): void {
+            if (!this.amountToBid) {
+                this.$emit('update:amountToBid', new BigNumber(0));
```
valiafetisov, line 104: 0 is actually an invalid input, since it's below minimum, therefore, I would better set it to NaN
```suggestion
                this.$emit('update:amountToBid', new BigNumber(NaN));
```

frontend/components/utils/BidInput.vue
valiafetisov: Also, I would move if into it's own computed prop `isInputEmpty` to improve readability, since you're checking if the input is empty in many places

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+            if (!this.amountToBidInput) {
+                return undefined;
+            }
+            const amountToBid = new BigNumber(this.amountToBidInput);
+            return amountToBid;
+        },
+        error(): string | undefined {
+            if (!this.amountToBidInputParsed) {
+                return undefined;
+            }
```
valiafetisov, line 65: Can be replaced with 
```suggestion
            if (this.isInputEmpty) {
                return undefined;
            }
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+    methods: {
+        hideTotalPrice(): void {
+            if (!this.amountToBid) {
+                this.$emit('update:amountToBid', new BigNumber(0));
+            }
+        },
+        showTotalPriceIfEmpty(): void {
+            if (!this.amountToBidInput) {
+                this.$emit('update:amountToBid', undefined);
+            }
```
valiafetisov, line 110:
```suggestion
            if (this.isInputEmpty) {
                this.$emit('update:amountToBid', undefined);
            }
```

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,164 @@
...
+            if (!this.amountToBid || !this.auctionTransaction.approximateUnitPrice) {
+                return this.auctionTransaction.collateralAmount;
+            }
+            return this.amountToBid.dividedBy(this.auctionTransaction.approximateUnitPrice);
+        },
+        isActive(): boolean {
+            return this.auctionTransaction.isActive && !this.auctionTransaction.isFinished;
+        },
+        isBidAmountNaN(): boolean {
+            return this.amountToBid?.isNaN() || this.amountToBid?.isEqualTo(0) || false;
```
valiafetisov, line 147: Then you can remove 0 from here
```suggestion
            return this.amountToBid?.isNaN();
```

valiafetisov:
In general, I think the implementation is much more cleaner now, well done! (:

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,182 @@
...
+        <div class="flex justify-between">
+            <button
+                :class="{ 'ClickableText underline': isActive && auctionTransaction.totalPrice }"
+                @click="setAmountToBid(undefined)"
+            >
+                Buy all collateral
+            </button>
+            <button
+                :class="{ ClickableText: isActive && auctionTransaction.totalPrice }"
+                @click="setAmountToBid(undefined)"
```
valiafetisov, line 61: It's better to use html-native `disabled` attribute, which will disable the click event for you (so you don't have to check it inside the method)
```suggestion
                class="Clickable"
                :disabled="!isActive || !auctionTransaction.totalPrice"
                @click="setAmountToBid(undefined)"
```
In css you can always use:
```css
.ClickableText:enabled {

}
```
or `.ClickableText:not([disabled])` to apply style only when the button is not disabled

valiafetisov: (same for 3 other places in this component)

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+        },
+        error(): string | undefined {
+            const maximumBid = this.totalPrice?.minus(this.minimumDepositDai) || new BigNumber(0);
+            if (this.amountToBidInputParsed?.isGreaterThan(maximumBid)) {
+                return `The bidding amount can not be greater than ${maximumBid.toFixed(2)} DAI`;
+            }
+            if (this.amountToBidInputParsed?.isLessThan(this.minimumDepositDai)) {
+                return `The bidding amount can not be smaller than ${this.minimumDepositDai.toFixed(2)} DAI`;
+            }
+            return undefined;
```
valiafetisov, line 70: It's ok to be explicit (so you don't have to edit anything here). I just hope that you know that undefined is returned by a function if no return is specified, so this is technically the same:
```suggestion
            return;
```
or
```suggestion
```

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
+<template>
+    <div class="w-full inline-block relative flex-shrink-0">
+        <Tooltip :visible="!!error" placement="topLeft" :title="error" class="items-center">
+            <div v-if="!amountToBid" class="Overlay TotalPrice">
+                <format-currency v-if="totalPrice && !disabled" :value="totalPrice" />
+                <span v-else class="opacity-50">UNKNOWN</span>
```
valiafetisov, line 6: Let's make it the same as in other table fields:
```suggestion
                <span v-else class="opacity-50">Unknown</span>
```

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,182 @@
...
+        <div class="flex justify-between font-bold">
+            <div>The amount to receive</div>
+            <div>
+                <format-currency
+                    v-if="amountToReceive && isActive && !isBidAmountNaN"
+                    :value="amountToReceive"
+                    :currency="auctionTransaction.collateralSymbol"
+                />
+                <div v-else>
+                    <span class="opacity-50">UNKNOWN</span>
```
valiafetisov, line 116:
```suggestion
                    <span class="opacity-50">Unknown</span>
```

frontend/components/utils/BidInput.vue
valiafetisov: > I think this would then automatically format the input value

I am not sure what you mean. `toFixed()` with no argument will not apply any formatting

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -0,0 +1,182 @@
...
+            if (!this.amountToBid || !this.auctionTransaction.approximateUnitPrice) {
+                return this.auctionTransaction.collateralAmount;
+            }
+            return this.amountToBid.dividedBy(this.auctionTransaction.approximateUnitPrice);
+        },
+        isActive(): boolean {
+            return this.auctionTransaction.isActive && !this.auctionTransaction.isFinished;
+        },
+        isBidAmountNaN(): boolean {
+            return this.amountToBid?.isNaN() || false;
```
valiafetisov, line 165: Why do you need to have `|| false`? Or is it just about being explicit?

frontend/components/utils/BidInput.vue
```diff
@@ -0,0 +1,136 @@
...
+        },
+        amountToBidInputParsed(): BigNumber | undefined {
+            if (this.isInputEmpty) {
+                return undefined;
+            }
+            const amountToBid = new BigNumber(this.amountToBidInput);
+            return amountToBid;
+        },
+        error(): string | undefined {
+            const maximumBid = this.totalPrice?.minus(this.minimumDepositDai) || new BigNumber(0);
```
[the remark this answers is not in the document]
fegbert: You're right I don't 🤦‍♂️ Tests were failing here so I thought I needed an if unknown case here, but the question mark was already enough.

frontend/components/transaction/BidTransactionTable.vue
fegbert: Since `amountToBid` can be unknown it wanted an else case, but I can also force it to be a boolean via `!!this.amountToBid?.isNaN()`

valiafetisov: Ah, true! I think the best would be to use `this.amountToBid?.isNaN() ?? false` 
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

This relatively new operator is easier to read since you know it only handles `null` or `undefined` and no weird type conversion is happening. `!!` is also fine, it's just a tip for the future!

valiafetisov (approved):

LukSteib (changes requested):
- [ ] I find the tooltip text (see S1 below) somewhat missleading as the `totalAmount` would also be greater but valid. Can we change it to "The bidding amount cannot be between {`totalAmount`- `minimumAmount`} and {`totalAmount`}"
   - [ ] And then we would have 3rd message for the case that `input` > `totalAmount` with tooltip text "The bidding amount cannot be greater than {`totalAmount`}"

What do you think @valiafetisov? 


S1
image

LukSteib:
> Actually, there are 3 separate error cases:

What about the case: 
Higher than `total` 


> For 3 we should display: "The value can only be total" or even disable the input, cause it doesn't make sense anymore?

Agree that for this case we can disable the input

valiafetisov:
> Higher than `total`

True, but the message can still stay the same (since we say what it can be instead of what it can't)

fegbert:
**error message**

image

**disabled on `totalPrice` > `min * 2`**

image

valiafetisov:
Looking at the disabled input, I think we also need to explain it. Eg a tooltip can say something like `The value can not be changed since the the leftover auction will be too small`

I also wonder if we can end up in this state with time? Eg, initially it was possible to take it partially, but since price dropped it is no longer the case? If so, we need to handle edge case where user had entered the value in advance. Technically we can add watcher (to the `isTooSmallToPartiallyTake` computed prop) to reset input value to the total price (ie undefined), otherwise the input will be disabled with whatever value was there (including an error value, for example) and the user will be confused.

fegbert:
**popup**

image

**decreasing into range**

valiafetisov (approved):
Special thanks for the `Decreasing Into Disabled Input` story! ✨

LukSteib (changes requested):
- [ ] Small typo in the tooltip (one `the` too much) 

image

valiafetisov (approved):

LukSteib (approved):

PR #142, opened 2022-03-14
Added WalletDepositWithDraw Block

### LINKED ISSUE #134: Implement `WalletDepositWithdrawBlock` component
### Goal

Implemented `WalletDepositWithdrawBlock` component in the storybook

### Context

For the [`Bidding with DAI` feature](https://github.com/sidestream-tech/unified-auctions-ui/issues/117), we need to provide the user possibility to deposit funds to VAT or withdraw them from VAT to the connected wallet. For the names of the props, please refer to the ([already implemented](https://github.com/sidestream-tech/unified-auctions-ui/pull/127)) `DepositBlock` component, which displays some of the same values as well as [`WalletTable` component](https://github.com/sidestream-tech/unified-auctions-ui/issues/133) if it's already implemented.

Please note: 
- deposit / withdraw is done via [`Button.Group`](https://antdv.com/components/button) and acts as a "tab" displaying different input field + buttons (those can be split into a separate component, if needed)
- by default we show _formatted_ maximum possible amount _over the input_ (please reuse/encapsulate as much as possible from the input implemented in #129)
  - when in the `deposit` mode maximum possible amount is `walletDAI`
  - when in the `withdraw` mode maximum possible amount is `walletVatDAI`
- if `walletDAI` or `walletVatDAI` changes and user did not entered their value, the amount is also updated

### Assets
Correct text for the component: https://docs.google.com/document/d/1BnHalnU9MWxUKvI_u39X0gDvc3YObQnHCsnwaK4c_7g/edit#heading=h.vzx3e4cwp5zr

Mock:
MakerDAO@2x (28)

### Tasks
- [ ] create new `WalletDepositWithdrawBlock` inside `wallet` folder
- [ ] create stories for all edge cases
- [ ] please check that `isLoading` prop is taken into the account (all buttons and the input should become disabled)
- [ ] please check that `isExplanationsShown` prop is taken into the account
- [ ] please check that `disabled` prop is taken into the account

benediktkaiser:
closes #134 

## Normal
Screenshot 2022-03-14 at 13 34 29

## Loading State
Screenshot 2022-03-14 at 13 40 17

## Expert mode
Screenshot 2022-03-14 at 13 45 41

## Extra Infos
After a call with Lukas, I also added the props `maxDeposit` and `maxWithdraw`, that limit how much DAI a user can enter in either input field. I also added a `max` button that lets to user easily deposit or withdraw all of their DAI. 

@valiafetisov, is there a better name for these props and do you like this addition?

Also. It was said I should add `disabled` and a `isLoading` prop. However, both would do the same thing (Disable all the elements). What should I call it? `disabled` or `isLoading`?


valiafetisov:
I don't think that max button is necessary, since the input should be pre-filled with the max value in the beginning. See issue description:

> by default we show _formatted_ maximum possible amount _over the input_

There are few points in one sentence:
- `by default` means: when the component is rendered or the tab is switched
- `formatted` means that if maximum is `2.2132432423`, we show `2.21`
- `over the input` means that although we show `2.21` on top of the input, we would still emit `2.2132432423` when the form is submitted
  - this magic formatting is only applied to the max value, all other values stay as they inputted
  - for details how it's implemented, you can refer to the #129

valiafetisov:
- [ ] let's have two separate events, one for deposit and the other one for withdraw
- [ ] the emitted number should be a big number

benediktkaiser:
> * over the input

Do you mean the ToolTip with this? I don't really know what this refers to. 

> * this magic formatting is only applied to the max value, all other values stay as they inputted

I used the magic formating from `AnimatedNumber`. I don't know if this is the kind of formating you want.

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,135 @@
...
+        <TextBlock v-if="isExplanationsShown" title="Moving funds">
+            The DAI funds can be moved freely between VAT and your wallet without any extra conversion. You only need
+            to be aware that each transfer incurs transaction fees. Also, if you do not have DAI funds to deposit yet,
+            there are several ways to obtain it:
+            <ul class="list-disc list-inside">
+                <li>By borrowing DAI against a collateral in the oasis.app</li>
+                <li>
+                    By purchasing it on a decentralized exchange like uniswap.org (correct DAI token address used on
+                    the “mainnet” network is
+                    <FormatAddress type="address" value="0x6b175474e89094c44da98b954eedeac495271d0f" shorten />)
```
valiafetisov, line 12: Please don't hardcode the value, but provide it via the prop, ie `tokenAddressDAI`?

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,135 @@
...
+    },
+    computed: {
+        canSubmit(): boolean {
+            if (this.selectedMethod === 'deposit') {
+                return this.depositAmount > 0 && this.depositAmount <= this.maxDeposit;
+            }
+            if (this.selectedMethod === 'withdraw') {
+                return this.withDrawAmount > 0 && this.withDrawAmount <= this.maxWithdraw;
+            }
+            return false;
```
valiafetisov, line 102: After replacing the input, this instead should check if the emitted value is NaN, since the validation logic already exists inside input

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,143 @@
...
+    },
+    props: {
+        isLoading: {
+            type: Boolean,
+            default: false,
+        },
+        minimumDaiAmount: {
+            type: Object as Vue.PropType<BigNumber>,
+            required: true,
+        },
```
valiafetisov, line 77: This doesn't need to be a prop, it's always 0

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,143 @@
...
+                <li>By borrowing DAI against a collateral in the oasis.app</li>
+                <li>
+                    By purchasing it on a decentralized exchange like uniswap.org (correct DAI token address used on
+                    the “mainnet” network is
+                    <FormatAddress type="address" :value="tokenAddressDAI" shorten />)
+                </li>
+            </ul>
+        </TextBlock>
+        <div>
+            <RadioGroup
```
valiafetisov, line 17: Please put it all inputs into a `form` element, so it's semantically correct form that can be operated from the keyboard and submitted via `enter` key

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,143 @@
...
+            :total-price="maxDeposit"
+            :minimum-deposit-dai="minimumDaiAmount"
+        />
+        <BidInput
+            v-else
+            :amount-to-bid.sync="withDrawAmount"
+            :disabled="isLoading"
+            :total-price="maxWithdraw"
+            :minimum-deposit-dai="minimumDaiAmount"
+        />
```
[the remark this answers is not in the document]
benediktkaiser: I added the keys, however the animation is still present.

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,142 @@
...
+            <BidInput
+                v-else
+                key="withdrawInput"
+                :amount-to-bid.sync="withDrawAmount"
+                :disabled="isLoading"
+                :total-price="maxWithdraw"
+                :minimum-deposit-dai="minimumDaiAmount"
+            />
+
+            <BaseButton type="primary" class="capitalize" :disabled="isLoading || !canSubmit" html-type="submit">
```
valiafetisov, line 45: So there is should be two types of being disabled:

- When the form was submitted and `isLoading` (therefore please pass `isLoading` to the button, so it's animated)
- When the form is disabled, eg till the depositing is allowed. Since depositing and withdrawing require different auth steps, we may want to disable them separately: let's add two new props `isDepositingAllowed` and `isWithdrawingAllowed` and disable them separately

frontend/components/wallet/WalletDepositWithdrawBlock.vue
```diff
@@ -0,0 +1,162 @@
...
+                :amount-to-bid.sync="depositAmount"
+                :disabled="isLoading"
+                :total-price="maxDeposit"
+                :minimum-deposit-dai="minimumDaiAmount"
+            />
+            <BidInput
+                v-else
+                key="withdrawInput"
+                :amount-to-bid.sync="withDrawAmount"
+                :disabled="isLoading"
```
valiafetisov, line 40:
```suggestion
                :disabled="isLoading || !isWithdrawingAllowed"
```

valiafetisov (dismissed):

valiafetisov (approved):

LukSteib (changes requested):
- [x] Let's provide `oasis.app` and `uniswap.org` as clickable links in the explainer block
- [ ] We are showing the wrong contract address for Dai token in the explainer block. 
   * Right contract: `0x6B175474E89094C44Da98b954EedeAC495271d0F`
   * etherscan link: https://etherscan.io/address/0x6b175474e89094c44da98b954eedeac495271d0f

valiafetisov:
> We are showing the wrong contract address for Dai token in the explainer block.

It will come from the props (since it's actually different on different networks) and currently it's a "fake" address generated in the storybook. But the link should already work (leading to the fake address)

benediktkaiser:
> * [x]  Let's provide `oasis.app` and `uniswap.org` as clickable links in the explainer block

I added the links.

LukSteib (changes requested):
- [ ] The tooltip for the error case is misleading since adapted from `BidTransactionTable` (see [1] below)
   * For deposit it should rather be: "The value can only be between 0.00 and {dai amount in wallet}
      * Note: As discussed this morning we assume unlimited allowance for now hence the dai amount in wallet is the limiting factor here -> will need to update once more granular allowance is possible
   * For withdraw it should rather be: "The value can only be between 0.00 and {Currently deposited in VAT}
- [ ] NIT: currently not able to specify a number with decimal that is between 0 and 1 
Dark mode related: 
- [ ] default input is of font color black -> should be white (see [2] below)
- [ ] `DAI` indicator in input field has font color black -> should be white (see [3] below)


[1]
image

[2]
image

[3]
image

valiafetisov (approved):

LukSteib (approved):

Issue #126, opened 2022-02-28
Create mock for the `Manage DAI in VAT` flow


valiafetisov:
### Goal

A mock for managing DAI in VAT

### Context

Mock started in #96 proposed a possibility to manage DAI in VAT, therefore we want to outline how it may look like.

### Assets
https://whimsical.com/makerdao-7jmkXyT32KBtY4nQ56WARx

### Tasks
- [x] Create new mock
- [x] Discuss, iterate
- [x] Finalize

valiafetisov:
Initial idea: 
- Create a `Wallet` popup which opens up by clicking on the `Manage DAI in VAT` or the wallet address in the header (since disconnect is dysfunctional and should be removed)
- Always prefill input with total value

Initial mock is below, but it misses explanations as well as authorisations:

![MakerDAO@2x (18)](https://user-images.githubusercontent.com/3393626/156028695-8e290f74-d827-4987-91db-efbb68aa294d.png)

valiafetisov:
![MakerDAO@2x (20)](https://user-images.githubusercontent.com/3393626/156142021-2f6dd75e-e988-4b9f-b184-f8d0218d5631.png)

@LukSteib what do you think? Would you write the text for the full popup? I think we may wait a little bit before finalizing the mock to see if it's actually enough to have one auth step here or if we need two (which I am starting to suspect 🙂)

LukSteib:
> opens up by clicking on the Manage DAI in VAT or the wallet address in the header

The latter might be confusing for a user that is not engaging in the Dai bidding but swapping flow and is then provided with the wallet popup. Right now I would vouch to only show on clicking `Manage Dai in VAT`


> what do you think? 

* I like the general idea of having this wallet popup

* Besides providing `Refresh` button to update balance indicators would we also provide automatic refreshs or only on user action? Would like to have an automatic refresh as well

* What is the idea behind the moving funds logic regarding input fields and buttons? Enter amount to deposit on left hand side input field and execute by clicking button below (same for right hand side withdraw logic)? 
  * I'd prefer the other mock you created first where the user can select between deposit and withdrawal -> enter amount -> press button 
  * What was your reasoning to switch? :) 

>  I think we may wait a little bit before finalizing the mock to see if it's actually enough to have one auth step here or if we need two 

Fine for me. No matter how many auth steps there will be, do you propose to have the auth steps for VAT interaction both inside this wallet popup and on our basic page layout?

[...]

Issue #123, opened 2022-02-28
Implement `setAllowanceAmount` logic


valiafetisov:
### Goal

Implemented `setAllowanceAmount` logic (store only, no UI)

### Context

Similar to the `authorizeWallet` and `fetchWalletAuthorizationStatus` this PR should extend `authorizations` store to include `setAllowanceAmount` and `fetchAllowanceAmount`. The actual, blockchain-related logic should be implemented inside `core/src/authorizations.ts`. Note that we also want to have some kind of de-auth possibility.

The actual transactions that should be executed can be found in the `liquidations.makerdao.com` source code:
https://github.com/makerdao/liquidations-portal/blob/0c6af9c5dcacbf60aeb39e9c3a7617180d2ae304/stores/approvals.ts#L89

Note: the UI for this functionality will be implemented in #115.

### Tasks
- [x] extend `authorizations` core
- [x] extend `authorizations` store with 3 new actions, new state
- [ ] extend `refetch` and `reset` actions to support fetching
- [x] in the PR description, provide example store commands to be called in the browser console for testing

benediktkaiser:
I had a look into this issue today and yesterday. As far as I can see the following steps must be executed in order to authorise Dai Access.

- Get `MCD_JOIN_DAI` contract
- Retrieve address from the retrieved contract
- Call `maker.getToken` and pass the address as a prop.

As we are no longer using `DAIJs`, I was wondering what exactly `maker.getToken` does. Does it call a contract or is it some other kind of function? 
Once I understand what it does, I might be able to mimimic it with `ether.js`. @valiafetisov, if you have any quick tips, I would love to hear them!

valiafetisov:
> I was wondering what exactly maker.getToken does

Judging by the name it's a dai token contract. You can get the same contract by calling `getContract(network, `DAI`)`. But you can, of course, just search for the `getToken` inside dai.js source code.

valiafetisov:
`authorizeAccessToDAI(network: string, amount?: BigNumber)`
- get `MCD_JOIN_DAI` contract address
- if no amount specified, set maximum possible amount, as in [`approveUnlimited`](https://github.com/makerdao/dai.js/blob/3f57eac9798a53a9bd96197514fe6d6fde5672de/packages/dai/src/eth/tokens/Erc20Token.ts#L68-L82)
- call `approve` on the `MCD_DAI` passing `MCD_JOIN_DAI` address and the amount

Issue #96, opened 2022-02-10
Bidding with DAI mock


valiafetisov:
### Goal
New mock for the `Bid with DAI` transaction type

### Tasks
- [x] Document existing flow at https://liquidations.makerdao.com
- [x] Create a new mock

valiafetisov:
### Current flow
1. Attach a wallet
2. Authorize depositing DAI to VAT (a transaction)
3. Deposit DAI in VAT
    - Fetch user wallet balance
    - Fetch user VAT balance
    - Input field
      - Max button to prefill
      - Let user edit deposited amount
    - Deposit button (a transaction)
4. Bid with DAI
    - Fetch user VAT balance
    - Input field for the DAI amount
      - Button to fill it with total price of the auction
      - Button to fill it with all DAI in the VAT
    - Fetch and show `dust limit`
    - Show errors in case
      - Bid is smaller than `dust limit`
      - `Total - bid` is smaller than `dust limit`
    - Calculate amount of collateral user will receive
    - Execute button (disabled in case of errors from above)
5. Withdraw DAI
    - Fetch user wallet balance
    - Fetch user VAT balance
    - Input field
      - Max button to prefill
      - User can edit
    - Withdraw button (a transaction)
      - Disabled in case of the problem

### Screenshots
<img width="463" alt="Screenshot 2022-02-10 at 13 25 53" src="https://user-images.githubusercontent.com/3393626/153408471-1f20f9d9-c096-4755-b649-abff934ee386.png">
<img width="472" alt="Screenshot 2022-01-27 at 15 03 32" src="https://user-images.githubusercontent.com/3393626/151387065-860a982e-2635-481c-bb40-c8444d37c4fb.png">
<img width="463" alt="Screenshot 2022-01-27 at 15 04 35" src="https://user-images.githubusercontent.com/3393626/151386973-44a24111-badf-4eca-9862-1511c06d9326.png">
<img width="467" alt="Screenshot 2022-01-27 at 15 05 58" src="https://user-images.githubusercontent.com/3393626/151387000-a5329c28-8b68-40f7-8e81-a105f3967a40.png">
<img width="466" alt="Screenshot 2022-01-27 at 16 06 46" src="https://user-images.githubusercontent.com/3393626/151387007-cc4cdd14-73f6-43fd-a751-93e0e7446427.png">

valiafetisov:
<details><summary>
Initial mock for the `Bid with DAI` transaction panel

</summary>

[...]

PR #143, opened 2022-03-14
Wallet logic

### LINKED ISSUE #124: Implement wallet-related logic
### Goal

Vuex store actions to fetch wallet data and `deposit`/`withdrew`

### Context

When the wallet is connected/disconnected/wallet address has changed, we want to update `wallet` store with new values:
- `walletDAI` – amount of DAI (`MCD_DAI`) tokens in the connected wallet
- `walletETH` – amount of ETH the connected wallet
- `walletVatDAI` – amount of DAI in the VAT related to that wallet

Therefore, we need to extend vuex store to add new variables and add two `core` functions to fetch that data. For inspiration, please refer to the [liquidations portal source code](https://github.com/makerdao/liquidations-portal).

We also need to add functions to `deposit` and `withdrew` DAI to/from VAT. For inspiration, please refer to https://github.com/makerdao/dai.js/blob/3511e9f4db70285d3a85776d22933dd092ba87e0/packages/dai-plugin-liquidations/src/LiquidationService.ts#L343-L354

### Tasks
- [x] investigate how it's done, document here with links
- [ ] implement required functions inside the `core`
- [ ] extend `wallet` store with the variables as well as new actions
- [ ] fetch this information when wallet is connected/changed
- [ ] add `deposit` and `withdrew` transactions to the `core`
- [ ] add `deposit` and `withdrew` actions
- [ ] in the PR description, provide example store commands to be called in the browser console for testing

valiafetisov:
Closes #124

Functional testing can be done by executing in the js console:

1. To fetch wallet balances and output fetched values:
    ```js
    await $nuxt.$store.dispatch('wallet/fetchWalletBalances'); ({ walletETH: $nuxt.$store.state.wallet.walletBalances.walletETH.toFixed(), walletDAI: $nuxt.$store.state.wallet.walletBalances.walletDAI.toFixed(), walletVatDAI: $nuxt.$store.state.wallet.walletBalances.walletVatDAI.toFixed() })
    ```

2. To deposit to VAT (if you have DAI and auth steps)
    ```js
    $nuxt.$store.dispatch('wallet/depositToVAT', '100')
    ```
    - then rerun `1` to validate new balances

3. To withdraw from VAT
    ```js
    $nuxt.$store.dispatch('wallet/withdrawFromVAT', '100')
    ```
    - then rerun `1` to validate new balances

### DIFF
--- a/core/src/abis/MCD_DAI.json
+++ b/core/src/abis/MCD_DAI.json
@@ -1,0 +1,563 @@
+[
+    {
+        "inputs": [
+            {
+                "internalType": "uint256",
+                "name": "chainId_",
+                "type": "uint256"
+            }
+        ],
+        "payable": false,
+        "stateMutability": "nonpayable",
+        "type": "constructor"
+    },
+    {
+        "anonymous": false,
+        "inputs": [
+            {
+                "indexed": true,
+                "internalType": "address",
+                "name": "src",
+                "type": "address"
+            },
+            {
+                "indexed": true,
+                "internalType": "address",
+                "name": "guy",
+                "type": "address"
+            },
+            {
+                "indexed": false,
+                "internalType": "uint256",
+                "name": "wad",
+                "type": "uint256"
+            }
+        ],
+        "name": "Approval",
+        "type": "event"
+    },
+    {
+        "anonymous": true,
+        "inputs": [
+            {
+                "indexed": true,
+                "internalType": "bytes4",
+                "name": "sig",
+                "type": "bytes4"
+            },
+            {
+                "indexed": true,
+                "internalType": "address",
+                "name": "usr",
+                "type": "address"
+            },
+            {
+                "indexed": true,
+                "internalType": "bytes32",
+                "name": "arg1",
+                "type": "bytes32"
+            },
+            {
+                "indexed": true,
+                "internalType": "bytes32",
+                "name": "arg2",
+                "type": "bytes32"
+            },
+            {
+                "indexed": false,
+                "internalType": "bytes",
+                "name": "data",
+                "type": "bytes"
+            }
+        ],
+        "name": "LogNote",
+        "type": "event"
+    },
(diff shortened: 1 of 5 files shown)


LukSteib (approved):
Tested all three steps on both `goerli` and `kovan` -> works ✅  (see below fetched funds on `goerli`)

Question:
Is the first one time auth step per wallet that we have in our flashloan based flow the same compared to the one required for depositing and withdrawing Dai? Since there was no authentication that I had to make.

Note:
The high figure for `walletVatDAI` made me wonder since I could not think of deliberately depositing Dai to VAT before. However, I realised that the incentives paid to restart an auction are paid to the VAT and that is how the balance grew in the past.

valiafetisov:
> Is the first one time auth step per wallet that we have in our flashloan based flow the same compared to the one required for depositing and withdrawing Dai? Since there was no authentication that I had to make.

It actually may be, you can always call `authorizations/deauthorizeWallet` and try again. But you, of course, also approved DAI withdrawal (that will be implemented in #123) once on the liquidation portal, which is still valid, since we're using the same contracts

benediktkaiser (approved):

valiafetisov:
Thanks for testing that!

> depositing Dai to Vat doesn't require authorisation

I think it requires another authorisation type ([it's part of the ERC-20 and is called `approve` there](https://docs.openzeppelin.com/contracts/2.x/api/token/erc20#IERC20-approve-address-uint256-)) from #123 in order to let specific contract withdraw DAI from your wallet. 

I guess that means the bid transaction mock is mostly correct, although the new authorisation should be part of the wallet block. I will update https://github.com/sidestream-tech/unified-auctions-ui/issues/96

> withdrawing Dai from Vat requires authorisation

Then we need to update the wallet popup mock https://github.com/sidestream-tech/unified-auctions-ui/issues/126 to add another authorisation type there.

PR #147, opened 2022-03-16
Fix auth status refetching

valiafetisov:
This PR fixes few auth-related bugs that I've noticed:

1. After the auth was executed, the auth button becomes active again for a second before turning into the authorized state:
    
    https://user-images.githubusercontent.com/3393626/158641177-c3abc6cd-5649-4546-8c8d-3665d550de5c.mov

2. If the user changes the `walletAddress` within metamask, we did not refetch auth states (until another auction is selected), it's solved now

3. If the user executed an auth step, changed the `walletAddress` within metamask, and tried to execute the same auth step with the new `walletAddress` was prevented due to caching logic, it's solved now

### DIFF
--- a/frontend/store/authorizations.ts
+++ b/frontend/store/authorizations.ts
@@ -116,8 +116,9 @@ export const actions = {
     async authorizeWallet({ commit, dispatch, rootGetters }: ActionContext<State, State>) {
         commit('setIsWalletAuthorizationLoading', true);
         const network = rootGetters['network/getMakerNetwork'];
+        const walletAddress = rootGetters['wallet/getAddress'];
         try {
-            await authorizeWallet(network, false, notifier);
+            await authorizeWallet(network, walletAddress, false, notifier);
             await dispatch('fetchWalletAuthorizationStatus');
         } catch (error) {
             console.error(`Wallet authorization error: ${error.message}`);
@@ -128,8 +129,9 @@ export const actions = {
     async deauthorizeWallet({ commit, dispatch, rootGetters }: ActionContext<State, State>) {
         commit('setIsWalletAuthorizationLoading', true);
         const network = rootGetters['network/getMakerNetwork'];
+        const walletAddress = rootGetters['wallet/getAddress'];
         try {
-            await authorizeWallet(network, true, notifier);
+            await authorizeWallet(network, walletAddress, true, notifier);
             await dispatch('fetchWalletAuthorizationStatus');
         } catch (error) {
             console.error(`Wallet authorization error: ${error.message}`);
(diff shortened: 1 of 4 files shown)


benediktkaiser (approved):

frontend/store/authorizations.ts
```diff
@@ -126,9 +129,10 @@ export const actions = {
     async deauthorizeWallet({ commit, dispatch, rootGetters }: ActionContext<State, State>) {
         commit('setIsWalletAuthorizationLoading', true);
         const network = rootGetters['network/getMakerNetwork'];
+        const walletAddress = rootGetters['wallet/getAddress'];
         try {
-            await authorizeWallet(network, true, notifier);
-            dispatch('fetchWalletAuthorizationStatus');
+            await authorizeWallet(network, walletAddress, true, notifier);
+            await dispatch('fetchWalletAuthorizationStatus');
```
BracketJohn, line 135: Should we discuss introducing https://eslint.org/docs/rules/require-await in order to:
- require await
- reduce count of accidentally `async` marked functions (that would trigger this lint and then lead the author to refactor and remove the `async`)

valiafetisov: It would be amazing! But as far as I can see, this rule unfortunately doesn't support `requiring await`, despite it's name, it can only check if the function has `async` keyword with no `await` inside. In other words it wouldn't prevent the situation from above. I think eslint can't do it on it's own, since this rule would require knowledge about function type, so it's more of a typescript ground.

In general, not awaiting a promise is a valid `fire and forget` practice. We already have places where we don't want to await promises (like in the bot loop) but it's more of an exception and yes, it would be nice to have such a rule and ability to ignore it sometimes with something like `// eslint-ignore-line await-promises` to make it explicit, but I can't find such a rule 🙁

BracketJohn: Fair point, but I'd argue that if the rule we desired (and I foolishly thought to have found without looking at the details) existed, we can add explicit `ignore`s to the palces where we want to fire and forget. That would be great signaling and actively enforce good practice and purposeful excuses from the rule.

How about: https://palantir.github.io/tslint/rules/no-floating-promises/? (: 

Btw, this proposal was inspired by rusts great developer experience, namely: https://rust-lang.github.io/rfcs/1940-must-use-functions.html

BracketJohn: edit/addition: a similar rule also exists for eslint I found it now, later: https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md

BracketJohn: Alright, then let's introduce https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md, can you open an issue in this repo to add it + add an issue here: https://github.com/sidestream-tech/vue-nuxt-ts-scaffold?

The latter is especially interesting as we can go through the issues of the scaffold repo to see all the rules we might  want to port over to a typescript v2 scaffold (:

florianstoecker (approved):
Looks good to me, tested also with @benediktkaiser at side

PR #149, opened 2022-03-17
Add Wallet Popup Component

### LINKED ISSUE #135: Implement `WalletPopup` component
### Goal

Implemented `WalletPopup` component in the storybook

### Context

For the [`Bidding with DAI` feature](https://github.com/sidestream-tech/unified-auctions-ui/issues/117), this issue is to assemble wallet popup from text and already implemented components:
- https://github.com/sidestream-tech/unified-auctions-ui/issues/133
- https://github.com/sidestream-tech/unified-auctions-ui/issues/134

Please note:
- the component should provide wallet information under a single prop `wallet` with all info (address, amounts, update date)
- the component should re-emit events from the sub components
  - an event to refetch wallet info
  - an event to deposit
  - an event to withdraw
- if the wallet is not connected, we should show `WalletBlock` instead of the table

### Assets
- Correct text for the component: https://docs.google.com/document/d/1BnHalnU9MWxUKvI_u39X0gDvc3YObQnHCsnwaK4c_7g/edit#heading=h.vzx3e4cwp5zr
- Latest mock: https://github.com/sidestream-tech/unified-auctions-ui/issues/126

### Tasks
- [ ] create new `WalletPopup` inside `wallet` folder
- [ ] create stories for all edge cases
- [ ] please check that `isConnecting` prop is taken into the account
- [ ] please check that `isLoading` prop is taken into the account
- [ ] please check that `isExplanationsShown` prop is taken into the account

### LINKED ISSUE #136: Integrate `WalletPopup` into the flow
### Goal

Complete and working `Manage DAI in VAT` popup

### Context
After all underlying issues are done, this issue is to modify `BidTransaction` component to include new `WalletPopup` popup and connect it to the store. Note: the button `Manage DAI in VAT` is already in the code, you just need to uncomment it.

### Assets
- `Manage DAI in VAT` mock https://github.com/sidestream-tech/unified-auctions-ui/issues/126

### Tasks
- [ ] Add `Manage DAI in VAT` button
- [ ] Trigger popup
- [ ] Write `WalletContainer` logic that interacts with the store

benediktkaiser:
Closes #135
Closes #136


benediktkaiser:
## Current State

I have the first working implementation of this feature!

Currently there is no button to open the modal, so please execute the following command in the console: 
```js
this.$nuxt.$store.dispatch("modals/setWalletModalAndFetchData", true)
```

## Issues

I ran into the following issues, that I find need to be addressed:
- When clicking on refresh, it takes a few seconds to update. During that time we show the user no visual feedback about the refresh. What could we add there? (My ideas would be: A spinner, A notification or clearing and resetting the table completely)
- The Max Deposit BigNumber is throwing me the following error: 
```js
Uncaught (in promise) Error: invalid BigNumber string 
(argument="value", value="457818107716361198836.546672822", code=INVALID_ARGUMENT, version=bignumber/5.5.0)
```
- Input fields do not reset properly when toggling between Deposit and Withdraw

If you find any other issues please document them! 

## Questions

- Where do I set my allowance in my wallet? Mine is currently set to 0, not allowing me to deposit any DAI into VAT.
- I withdrew 5 DAI in [this transaction](https://kovan.etherscan.io/tx/0x74e0f5e1d47563e121c9b6b885ee1f03ff410ea9c23ea1f8916be146e4bbce04). Did it work correctly? @valiafetisov

LukSteib (changes requested):
No additional problems encountered. Moste severe ones: 

- [x] Allowance Block in popup missing
- [ ] No Input field once Withdraw is selected

Additonaly when switching to Withdraw I do get a bunch of console errors: 
(1)
``` 
[Vue warn]: Invalid prop: type check failed for prop "totalPrice". Expected Object, got Number with value 1521.

found in

---> <BidInput> at components/utils/BidInput.vue
       <WalletDepositWithdrawBlock> at components/wallet/WalletDepositWithdrawBlock.vue
         <Anonymous>
           <Anonymous>
             <Portal>
               <PortalWrapper>
                 <Anonymous>
                   <AModal>
                     <ManageVATModal> at components/wallet/WalletModal.vue
                       <WalletPopupContainer> at containers/WalletPopupContainer.vue
                         <Layouts/default.vue> at layouts/default.vue
                           <Root>
```

(2)
```
TypeError: _a.isLessThan is not a function
    at VueComponent.isTooSmallToPartiallyTake (BidInput.vue?3fa7:81:1)
    at Watcher.get (vue.runtime.esm.js?2b0e:4495:1)
    at Watcher.evaluate (vue.runtime.esm.js?2b0e:4597:1)
    at VueComponent.computedGetter [as isTooSmallToPartiallyTake] (vue.runtime.esm.js?2b0e:4851:1)
    at VueComponent.eval (vue.runtime.esm.js?2b0e:503:1)
    at Watcher.get (vue.runtime.esm.js?2b0e:4495:1)
    at new Watcher (vue.runtime.esm.js?2b0e:4484:1)
    at VueComponent.Vue.$watch (vue.runtime.esm.js?2b0e:4961:1)
    at createWatcher (vue.runtime.esm.js?2b0e:4921:1)
    at initWatch (vue.runtime.esm.js?2b0e:4903:1)
```

(3)
```
TypeError: _a.minus is not a function
    at VueComponent.error (BidInput.vue?3fa7:69:1)
    at Watcher.get (vue.runtime.esm.js?2b0e:4495:1)
    at Watcher.evaluate (vue.runtime.esm.js?2b0e:4597:1)
    at VueComponent.computedGetter [as error] (vue.runtime.esm.js?2b0e:4851:1)
    at Object.get (vue.runtime.esm.js?2b0e:2081:1)
    at Proxy.render (BidInput.vue?71ee:13:1)
    at VueComponent.Vue._render (vue.runtime.esm.js?2b0e:3569:1)
    at VueComponent.updateComponent (vue.runtime.esm.js?2b0e:4070:1)
    at Watcher.get (vue.runtime.esm.js?2b0e:4495:1)
    at new Watcher (vue.runtime.esm.js?2b0e:4484:1)
```

benediktkaiser:
> Additonaly when switching to Withdraw I do get a bunch of console errors:

The missing inputs and errors were because I accidentally pushed something I didn't want to :o
They are gone now and the input works again!

LukSteib (changes requested):
- [ ] Withdraw attempt of the pre-filled input throws an error (see [1] below)
   - [ ] When entering the exact same float manually the withdrawal works though
- [ ] Entering input for withdrawal and then switching between the Deposit and Withdraw causes the input to get lost 
- [ ] Swap order of "What is Vat" block with allowance block
- [ ] If an error is thrown once the popup is displayed the underlying UI breaks (see [2]). This is the case for different console errors. 


[1]
```
client.js?06a0:103 Error: invalid BigNumber string (argument="value", value="1521863676086179560050.597897128", code=INVALID_ARGUMENT, version=bignumber/5.5.0)
    at Logger.makeError (index.js?ef0f:185:1)
    at Logger.throwError (index.js?ef0f:194:1)
    at Logger.throwArgumentError (index.js?ef0f:197:1)
    at Function.from (bignumber.js?e343:194:1)
    at NumberCoder.encode (number.js?7f46:16:1)
    at eval (array.js?560a:57:1)
    at Array.forEach (<anonymous>)
    at pack (array.js?560a:43:1)
    at TupleCoder.encode (tuple.js?17f8:51:1)
    at AbiCoder.encode (abi-coder.js?34ae:87:1)
``` 
[2]
image

image

benediktkaiser:
I fixed all issues aside from the following: 

- When you try to bid with the default set Max Value, it may say that the value is too high and you do not have enough DAI (This is due to rounding. If you have `232.787 DAI` it gets rounded to `232.79 DAI`, which you do not have). I cannot find where this rounding is happening. @valiafetisov do you have any pointers?
- When switching between WithDraw/Deposit with user filled values, they disappear. This is because when you switch back and forth the value inside the input becomes `undefined`, but the value in the wrapper component stays the user filled number. How can we sync these two again?

Other then these two issues I think I addressed everything!

valiafetisov:
> but the value in the wrapper component stays the user filled number.

Can't you also reset the value in the wrapper to be undefined when the user switches the tabs?

> * If you have `232.787 DAI` it gets rounded to `232.79 DAI`, which you do not have

This doesn't seem correct. The reason why we store maximum value as `undefined` is exactly because of this: the value is only _displayed_ as rounded, but the actual maximum number should be used for the transaction

valiafetisov:
> Withdraw attempt of the pre-filled input throws an error

The error reason was quite crazy actually. 

When we fetch amount of DAI in VAT, [we receive it in `RAD` format](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/vat.sol#L59), which is a number with **45** decimal points, eg:
```
1000.000000000000000000507676487841078701905040960
```

But the withdrawal function [expects number in `WAD` format](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/join.sol#L115), which is a number with **18** decimal points, eg:
```
1000.000000000000000000
```
And when we tried to pass actual total number it can get rounded up to:
```
1000.000000000000000001
```
Which aborts the transaction. Which means the user can never withdraw some small amount which they actually have, like:
```
0.000000000000000000507676487841078701905040960
```

florianstoecker (changes requested):
Fuctionality-wise:
In general, it worked nicely from my perspective! I've tested it on Kovan & Goerli, here you can see the transactions I've made:
- Goerli:
  - [Depositing](https://goerli.etherscan.io/tx/0x2fed4ba0d9aa114a623a81cee84e40cd1a4876da6da519273004976f546ba139)
  - [Withdrawing](https://goerli.etherscan.io/tx/0x702df095071205c31de4df6463c3a2bf5d175bd837542fcc5c1d3651b61b7f0e)
- Kovan:
  - [Deposting](https://kovan.etherscan.io/tx/0x5cef9db6aedfa9d989b5debf60a64948c0a939969d23292abfeede40e7ab5a27)
  - [Withdrawing](https://kovan.etherscan.io/tx/0x84cfa88937beb2e9f9e43d1aafc8f25f17102a0fa52a84145752c936ca03df1b)
  
I have two smallish functionality-abnormalities:
- [x] When I have withdrawn the maximum amount (default value that was visible), there are still some leftovers in my VAT: (It is not the case for depositing)
  image
- [x] It is not possible to enter amounts below 1 --> e.g. "0.5", "0.2"; Although I don't think there will be a use case for it. If that's an easy fix, let's fix it, else I would still propose to edit the info text that shows up, when entering a too high amount from "between 0.00 and `max`" to "between 1.00 and 'max'"



I have also found a few cosmetics that should be changed:
- [x] In the table, the value for `Last updated` is not centered
  image
- [x] For the first text block `What is VAT?`
  - [x] Change "inside of the MakerDAO protocol" to "inside the Maker Protocol"
  - Change "the core of the MakerDAO" to "the core of the Maker Protocol"
- [ ] For the second text block `Access to DAI`
  - [ ] There is a space in front of the first dot.  
    image
  - [ ] In the expandable explanation of the word `VAT`, there is a space included in the link
    image
- [ ] Third text block `Moving funds`
  - [ ] In the last bullet, change the text from "on the "mainnet" network is" to "On the "main" Network is" 
- [ ] When clicking on `Withdraw`the loading state should say "Withdrawing..."
  image
- [ ] Analogue to the point above: When clicking on `Deposit`, the loading state should say "Depositing..."
  image
- [ ] In the light mode, the selected option(Withdraw/Deposit) is in full-color green & in the dark mode it is only bordered green.
  image
  image

valiafetisov:
> When I have withdrawn the maximum amount (default value that was visible), there are still some leftovers in my VAT: (It is not the case for withdrawing)

See [my comment above](https://github.com/sidestream-tech/unified-auctions-ui/pull/149#issuecomment-1077844563). We can potentially try to round fetched value as well to not confuse a user, but not sure if it's a correct solution.

florianstoecker:
> See [my comment above](https://github.com/sidestream-tech/unified-auctions-ui/pull/149#issuecomment-1077844563). We can potentially try to round fetched value as well? I'm not sure if it's a correct solution, but definitely less confusing.

Sorry, missed that one! We can also keep it like this, as it shows the correct information (that there is still something left in the VAT) & create an issue in the backlog that handles this rounding problem.

LukSteib:
> We can potentially try to round fetched value as well? I'm not sure if it's a correct solution, but definitely less confusing.

I agree, maybe not 100% correct but the upside of being less confusing prevails imo.

---

Would like to add one more thing to the functional review list:

- [ ] opening the WalletPopup with no wallet connected throws an repeating console error and breaks the UI

console error:
`Fetching allowance amount error: invalid address or ENS name (argument="name", value=undefined, code=INVALID_ARGUMENT, version=contracts/5.5.0)`

broken UI:
image

valiafetisov:
> In the light mode, the selected option(Withdraw/Deposit) is in full-color green & in the dark mode it is only bordered green.

Nice catch. I've changed Withdraw/Deposit buttons to be outlined to not interfere with call-to-action primary button.

LukSteib (changes requested):
- [ ] NIT: the flow when connecting a wallet from within the popup is a bit cumbersome as wallet connection popup causes the WalletPopup to close. So the user will have to reopen the popup before proceeding (see recording 1 below). I know that overlaying two popups is also not nice but maybe better than having the user to actively open the walletpopup again



1
https://www.loom.com/share/5ecf5c038cd74d4baa6f940f1de200c7

LukSteib (approved):

florianstoecker (approved):

PR #151, opened 2022-03-18
`BidTransaction` Component

### LINKED ISSUE #121: Implement `BidTransaction` component
### Goal

Implemented `BidTransaction` component via the `AuctionFakeData` container

### Context

Similar to the `SwapTransaction`, we need to implement `BidTransaction` first using stub data. All the underlying component should be ready at this time: `AccessDaiBlock`, `DepositBlock`, `BidBlock`, `BidTransactionTable`. 

Note that `Bid with DAI` button should stay disabled in this issue as it will be implemented later.

### Assets

- https://github.com/sidestream-tech/unified-auctions-ui/issues/96#issuecomment-1054227644

### Tasks
- [ ] create new `BidTransaction` inside `transaction` folder
- [ ] please handle `isFinished` and `isActive` states properly (reuse error handling from another transaction component)
- [ ] the flow should be testable in `AuctionFakeData` or the storybook

fegbert:
closes #121 

**preview**

### DIFF
--- a/frontend/components/Auction.vue
+++ b/frontend/components/Auction.vue
@@ -1,9 +1,9 @@
 <template>
     <TextBlock :title="`Auction #${auctionId}`">
-        <Alert v-if="errorText" class="my-3" :message="errorText" type="error" />
+        <Alert v-if="auctionError" class="my-3" :message="auctionError" type="error" />
         <div v-if="auction">
             <RestartBlock
-                v-if="errorText === 'This auction is inactive and must be restarted'"
+                v-if="auctionError === 'This auction is inactive and must be restarted'"
                 :transaction-fee="auction.restartTransactionFeeETH"
                 :is-explanations-shown="isExplanationsShown"
                 :wallet-address="walletAddress"
@@ -203,19 +203,12 @@
             <TextBlock>
                 <div class="flex w-full justify-end flex-wrap mt-4">
                     <Tooltip placement="top">
-                        <!--<div slot="title">
-                            This website does not yet support bidding on the auction with your own DAI. In the
-                            meantime, you can use
-                            <a href="https://liquidations.makerdao.com/" target="_blank" class="underline text-primary"
-                                >Liquidation Portal</a
-                            >
-                        </div>-->
                         <div v-if="!isPageNetworkFake" slot="title">
                             This feature is currently only supported for stub data.
                         </div>
                         <div>
                             <Button
-                                :disabled="!isPageNetworkFake"
+                                :disabled="!isPageNetworkFake || !!auctionError"
                                 type="secondary"
                                 class="w-60 mb-4"
                                 @click="$emit('purchase')"
@@ -224,9 +217,14 @@
                             </Button>
                         </div>
                     </Tooltip>
-                    <Tooltip :title="errorText" placement="bottom">
+                    <Tooltip :title="swapTransactionError" placement="bottom">
                         <div>
-                            <Button :disabled="!!errorText" type="primary" class="w-60 ml-4" @click="$emit('swap')">
+                            <Button
+                                :disabled="!!swapTransactionError"
+                                type="primary"
+                                class="w-60 ml-4"
+                                @click="$emit('swap')"
+                            >
                                 Directly swap into profit
                             </Button>
                         </div>
@@ -326,7 +324,7 @@ export default Vue.extend({
         };
     },
     computed: {
-        errorText(): string | null {
+        auctionError(): string | null {
             if (!this.areAuctionsFetching && !this.areTakeEventsFetching && !this.auction && !this.takeEvents) {
                 return 'This auction was not found';
             } else if (this.error) {
@@ -335,7 +333,14 @@ export default Vue.extend({
                 return 'This auction is finished';
             } else if (!this.auction?.isActive && !this.areAuctionsFetching && !this.areTakeEventsFetching) {
                 return 'This auction is inactive and must be restarted';
-            } else if (
+            }
+            return null;
+        },
+        swapTransactionError(): string | null {
+            if (this.auctionError) {
+                return this.auctionError;
+            }
+            if (
                 !this.areAuctionsFetching &&
                 !this.areTakeEventsFetching &&
                 typeof this.auction?.marketUnitPriceToUnitPriceRatio === 'undefined'
@@ -346,7 +351,7 @@ export default Vue.extend({
             return null;
         },
         isPageNetworkFake(): Boolean {
-            return this.$store.getters['network/isPageNetworkFake'];
+            return this.$store?.getters['network/isPageNetworkFake'];
         },
     },
     watch: {
--- a/frontend/components/Auction.vue
+++ b/frontend/components/Auction.vue
@@ -203,13 +203,13 @@
             <TextBlock>
                 <div class="flex w-full justify-end flex-wrap mt-4">
                     <Tooltip placement="top">
-                        <div slot="title">
+                        <!--<div slot="title">
                             This website does not yet support bidding on the auction with your own DAI. In the
                             meantime, you can use
                             <a href="https://liquidations.makerdao.com/" target="_blank" class="underline text-primary"
                                 >Liquidation Portal</a
                             >
-                        </div>
+                        </div>-->
                         <div>
                             <Button
                                 :disabled="!!errorText"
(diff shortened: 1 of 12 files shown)


frontend/components/Auction.vue
```diff
@@ -203,15 +203,20 @@
...
                     <Tooltip placement="top">
-                        <div slot="title">
+                        <!--<div slot="title">
                             This website does not yet support bidding on the auction with your own DAI. In the
                             meantime, you can use
                             <a href="https://liquidations.makerdao.com/" target="_blank" class="underline text-primary"
                                 >Liquidation Portal</a
                             >
-                        </div>
+                        </div>-->
```
valiafetisov, line 212: I think you can safely delete this comment

frontend/components/Auction.vue
```diff
@@ -203,15 +203,20 @@
...
                         <div>
-                            <Button disabled type="secondary" class="w-60 mb-4" @click="$emit('purchase')">
+                            <Button
+                                :disabled="!!errorText"
```
valiafetisov, line 215: Currently there is an errorText with `Swap transaction is not possible...` which suppose to only block swap transaction button and not this one.

To solve it, you can have 2 or 3 different computed properties: `auctionError`, `swapTransactionError` and (if exists) `bidTransactionError`. `swapTransactionError` can first check `if(this.auctionError)` (:

frontend/components/MainFlow.vue
```diff
@@ -118,13 +159,31 @@ export default Vue.extend({
...
+            default: undefined,
+        },
+        minimumBidDai: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: null,
+        },
     },
     data: () => ({
         step: 0,
+        isPurchasing: false,
```
valiafetisov, line 185: I guess this is unused?
```suggestion
```

frontend/components/transaction/BidTransactionTable.vue
```diff
@@ -168,6 +168,17 @@ export default Vue.extend({
         isBidAmountNaN(): boolean {
             return !!this.amountToBid?.isNaN();
         },
+        isTooSmallToPartiallyTake(): boolean {
+            return this.auctionTransaction?.totalPrice.isGreaterThanOrEqualTo(this.minimumDepositDai?.multipliedBy(2));
+        },
```
valiafetisov, line 173: Why do you need to check it again here? I thought in this case, BidInput should emit NaN?

frontend/components/transaction/DepositBlock.vue
```diff
@@ -6,26 +6,30 @@
             the minimum amount to deposit in order to participate in the auction based on the bid you want to make. You
             can directly deposit the minimum required amount of DAI if you don't have sufficient deposits yet.
         </TextBlock>
-        <div class="text-base font-medium">
+        <div class="text-base font-medium text-gray-700 dark:text-gray-100">
             <div class="flex justify-between">
                 <span>Amount available in the wallet</span>
-                <format-currency v-if="walletDAI" :value="walletDAI" currency="DAI" />
+                <format-currency v-if="walletDai" :value="walletDai" currency="DAI" />
```
valiafetisov, line 12: I think we actually tried to always write `DAI` in uppercase, why did you decide to rename it everywhere?

LukSteib (changes requested):
Functional review only possible via storybook at this point time, correct?

- [x] Story for the finished auction doesn't show the link to succesfull transaction on etherscan
- [ ] In finished state the `Bidding` button keeps spinning
- [ ] Unwanted border color behavior for input field (dark mode only). In case of input error when the input field is selected but one hovers away the border changes back to green color (see screenshot below) 


image

frontend/components/transaction/BidTransactionTable.vue
fegbert: I wanted to disable the `Buy all collateral` and `Set minimum bid` buttons in this case, but using the `NaN` check here would also disable it when errors occur so this was the only way I figured out to isolate it.

frontend/components/transaction/DepositBlock.vue
fegbert: I renamed it because, if uppercased, the prop would change from `:minimum-deposit-dai="minimumDepositDai"` to `:minimum-deposit-d-a-i="minimumDepositDAI"` which I thought looked quite weird so I changed it, but if we try to keep it as uppercased everywhere I'll change it back 🙂

frontend/components/Auction.vue
```diff
@@ -337,6 +350,9 @@ export default Vue.extend({
             }
             return null;
         },
+        isPageNetworkFake(): Boolean {
+            return this.$store?.getters['network/isPageNetworkFake'];
+        },
```
valiafetisov, line 355: Wouldn't this break the storybook? I would not use $store outside of containers or layouts

Instead you can use `auction.network` to check it against `FAKE_NETWORK_NAME`

frontend/containers/AuctionsFakeData.vue
```diff
@@ -75,6 +95,11 @@ export default Vue.extend({
                 }
             },
         },
+        transactionAmountDai(newAmount) {
+            if (newAmount === undefined) {
+                this.transactionAmountDai = this.selectedAuction?.totalPrice || undefined;
```
valiafetisov, line 100:
```suggestion
                this.transactionAmountDai = this.selectedAuction?.totalPrice ?? undefined;
```

As we're working with numbers, it's better to use `??` operator to avoid automatic conversion of `0` into falsy statement. It's not very relevant here, since we're using BigNumber, but it's a good practice to avoid `||` where you want to handle only `?.` / `null` / `undefined`

frontend/containers/AuctionsFakeData.vue
```diff
@@ -120,6 +160,8 @@ export default Vue.extend({
                     this.selectedAuction.isFinished = true;
                     this.selectedAuction.endDate = new Date();
                     this.selectedAuction.transactionAddress = faker.finance.ethereumAddress();
+                    this.walletVatDai = this.walletVatDai.minus(this.transactionAmountDai!!);
```
valiafetisov, line 163: What's `!!` suppose to do here?

frontend/containers/AuctionsFakeData.vue
```diff
@@ -120,6 +160,8 @@ export default Vue.extend({
...
+                    this.transactionAmountDai = new BigNumber(NaN);
```
valiafetisov, line 164: Why does calling `authorizeWallet` sets `transactionAmountDai` to NaN?

frontend/components/transaction/DepositBlock.vue
valiafetisov: Although this component prop prettification is not required by vue (you can keep it as `:minimumDepositDAI=""`) and it's not part of our linter setup (but your editor setup, most probably fixable adding `"vetur.format.enable": false` to your VSC settings), I see your point! Let's keep it as is.

frontend/containers/AuctionsFakeData.vue
fegbert: Well, the second one was accidental 😅
One is to remove `undefined` since this step can't really be reached with `transactionAmountDai` being `undefined` and `minus` doesn't work with `BigNumber | undefined`

frontend/containers/AuctionsFakeData.vue
fegbert: Uhm.. good question

valiafetisov (dismissed):
Code approved! @LukSteib please check it with `npm run dev` including the old swap transaction flow on the stub and normal networks

LukSteib (changes requested):
Complete review done via `npm run dev`
Bid with Dai flow (stub data):
- [x] Expert mode not working 

Swap transaction flow (on kovan)
- [ ] We do no longer prevent auction execution when there is still a negative `Transaction Gross Profit`. This leads to an execution error. (Problem also apparent on production environment -> hence I would create a dedicated bug issue for it and not cover in this PR) Do you agree?

valiafetisov:
> I would create a dedicated bug issue for it and not cover in this PR) Do you agree?

Yes, let's create an issue for this

LukSteib (changes requested):
- [ ] We show the tx hash and etherscan link for a succesful flash loan based execution on two different places now (see below). 

image

LukSteib (approved):

PR #160, opened 2022-03-23
Integrate Bid Transactions into actual flow

### LINKED ISSUE #125: Integrate `BidTransaction` into the flow
### Goal

Complete and working `BidTransaction` flow

### Context
After all underlying issues are done, this issue is to modify `AuctionContainer` component to include new `BidTransaction` flow.

### Assets

- https://github.com/sidestream-tech/unified-auctions-ui/issues/96#issuecomment-1054227644

### Tasks
- [ ] Add `BidTransaction` to the `MainFlow`
- [ ] Write container logic that connects new components together with store

fegbert:
closes #125


fegbert:
Couple of questions on some values:

**Auction Minimum Bid**: What actually is this value / where does it come from?

**Allow access to DAI**: Is this also setting `isWalletAuthorised` or is this doing something else?

**AllowanceAmount**: What's this used for / Do I use even this or is it for the wallet popup?

valiafetisov:
> **Auction Minimum Bid**: What actually is this value / where does it come from?

Good question, I think I forgot to fetch this value. Will open a new PR to add this value to each auction

> **Allow access to DAI**: Is this also setting `isWalletAuthorised` or is this doing something else?

This is actually not a binary value as initially thought of, but a BigNumber. In the auth store it's called `allowanceAmount`. The value indicates how much DAI the user can deposit to VAT. So it should be not less then the `minimum to deposit`

valiafetisov:
So the proposal for this PR is to:
- [ ] remove `AccessDaiBlock` from the transaction panel (since it's only needed if there is not enough DAI in the wallet)
- [ ] simplify `DepositBlock` by only leaving `Manage DAI in VAT`/`Enough DAI is available` button
- [ ] attach `Manage DAI in VAT` to the actual modal
- [ ] update stories
- [ ] provide correct loading states

MakerDAO@2x

fegbert:
> simplify DepositBlock by only leaving Manage DAI in VAT/Enough DAI is available button

Do I keep `Minimum to deposit` and the explanation text like proposed in the mock below or really only the button?

valiafetisov:
> Do I keep Minimum to deposit and the explanation text like proposed in the mock below or really only the button?

Sorry for the confusion. You can keep the `Minimum to deposit` and other values in the table too (balance and balance in VAT), requested change only affects the buttons. I've updated the mock above to reflect this.

fegbert:
As discussed in today's meeting, the bidding logic is still missing and the button currently just executes the same way as a swap transaction.

frontend/components/transaction/BidTransaction.vue
```diff
@@ -140,17 +118,10 @@ export default Vue.extend({
...
-            default: null,
-        },
-        minimumBidDai: {
-            type: Object as Vue.PropType<BigNumber>,
-            default: null,
-        },
     },
     data() {
         return {
+            transactionAmountDai: undefined as BigNumber | undefined,
```
valiafetisov, line 124: We store the same value under 3 different names:
- `amountToBid`
- `transactionAmountDai`
- `inputBidAmount` (although this also can contain `undefined`, so it's technically different)

Would it make sense to refactor:
- `amountToBid` -> `transactionBidAmount`
- `transactionAmountDai` -> `transactionBidAmount`

therefore leaving just two `inputBidAmount` and `transactionBidAmount`?

frontend/components/transaction/BidTransaction.stories.js
```diff
@@ -93,41 +73,27 @@ const common = {
...
+        deposit() {
+            this.walletVatDai = this.transactionAmountDai;
         },
     },
     template: `
     <BidTransaction
         :auctionTransaction="auctionTransaction"
         :isConnecting="isConnecting"
-        :isGrantingAccess="isGrantingAccess"
-        :isDepositing="isDepositing"
```
valiafetisov, line 112: We still may want to provide `isDepositing` which will turn `Manage DAI in VAT` button into the loading state in case user closed the wallet modal before the transaction finished.

valiafetisov (commented):
Everything else looks good, the only thing left is to merge https://github.com/sidestream-tech/unified-auctions-ui/pull/163 and attach it to the button. If you're blocked, you can already do that (`git merge bidding-with-dai-logic` into your branch)

valiafetisov (dismissed):

LukSteib (changes requested):
- [x] The minimum to deposit amount in `DepositBlock` should update properly with each price drop (currently only after browser refresh)
- [ ] The finished auction state is not displayed properly. Bid button is still clickable (throwing an error) and the etherscan link is not displayed (see loom 1 below)

1
https://www.loom.com/share/cfaf4ba6abb54986a811015f64ff187d

valiafetisov:
> Bid button is still clickable (throwing an error) and the etherscan link is not displayed (see loom 1 below)

This should be probably done in https://github.com/sidestream-tech/unified-auctions-ui/pull/163
I removed transaction link because since we introduce partial bidding, there is a possibility that the user will bid part of the auction and we don't yet have logic to handle multiply transactions per auction 🙁

LukSteib:
> because since we introduce partial bidding, there is a possibility that the user will bid part of the auction and we don't yet have logic to handle multiply transactions per auction

Makes sense. Fine to remove it then for this flow for now.

LukSteib (approved):

valiafetisov (approved):

PR #161, opened 2022-03-24
BasePanel component

### LINKED ISSUE #152: Create `BasePanel` component
### Goal
New reusable `BasePanel` component in the storyook

### Context
Based on the proposal made in https://github.com/sidestream-tech/unified-auctions-ui/issues/145, we want to have a new kind of base element for our UI: a panel. A panel is a component consisting of:
- prop `state`
  - `inactive`
  - `incorrect`
  - `complete`
  - `notice`
- prop and slots for each state
- An icon (based on the state)
- Default child slot to include content within the frame
- An expandable logic _with animation_ (default is based on the state, but the title is still user-clickable/collapsable/expandable and reacts to the `state` change)

Notes: 
- This component doesn't need to have `isExplanationsShown` prop, since this logic will be inside the child slot
- For the `prop and slot` logic, please use "fallback content" feature https://v2.vuejs.org/v2/guide/components-slots.html#Fallback-Content
- Example usage of the component:
```
<BasePanel
    :state="currentState"
    inactive="The VAT balance is unknown until a wallet is connected"
    incorrect="The amount is not present in the VAT"
    complete="The amount is present in the VAT"
>
  Content and buttons
</BasePanel>
```

### Assets
- `inactive`
    MakerDAO@2x (44)

- `incorrect`
    MakerDAO@2x (43)

- `complete`
   MakerDAO@2x (45)

- `notice`
   MakerDAO@2x (47)
 
- Slotted title
    MakerDAO@2x (48)

- Stack of the panels (please note that the panel borders should overlap each other with `.BasePanel + .BasePanel` css selector)
    MakerDAO@2x (49)

### Tasks
- [x] Create a BasePanel component
- [ ] Add stories for all possible use-cases
  - [x] Multi-line title
  - [x] Title via named slot (eg with the wallet address)
  - [ ] Stack of the panels
  - [ ] Animation between `incorrect` and `complete` states

valiafetisov:
Closes #152 

Screenshot 2022-03-24 at 18 34 40

### Todo
- [ ] add animation (or in another PR, to unblock others)

### DIFF
--- a/frontend/components/common/BasePanel.stories.js
+++ b/frontend/components/common/BasePanel.stories.js
@@ -1,0 +1,53 @@
+import { storiesOf } from '@storybook/vue';
+import faker from 'faker';
+import BasePanel from './BasePanel.vue';
+
+const common = {
+    components: { BasePanel },
+    data: () => ({
+        inactive: `This is the inactive title`,
+        incorrect: `This is the incorrect title`,
+        correct: `This is the correct title`,
+        notice: `This is the notice`,
+    }),
+};
+
+storiesOf('common/BasePanel', module)
+    .add('Stack of panels', () => ({
+        ...common,
+        template: `<div>
+            <BasePanel v-bind="$data" currentState="incorrect"> Error content </BasePanel>
+            <BasePanel v-bind="$data" currentState="notice"> Notice content </BasePanel>
+            <BasePanel v-bind="$data" currentState="correct"> Correct content </BasePanel>
+            <BasePanel v-bind="$data" currentState="inactive"> Inactive content </BasePanel>
+        </div>`,
+    }))
+    .add('Inactive', () => ({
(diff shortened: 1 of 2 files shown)


florianstoecker (commented):
From a user perspective, I think that the `correctTitle` should also have the green color the icon has too, to further visualize that the step has been completed. Only the `unknownTItle` should be gray, as the icon also is.

Other than the component looks good to me in the storybook.

valiafetisov:
> should also have the green color the icon has too

We use green color in our UI as a primary color, mostly to highlights clickable elements and draw user attention to them. Therefore there is conflict of meanings. Since correct elements doesn't require user attention, we don't want to use green for that element.

florianstoecker (approved):

fegbert (approved):

PR #163, opened 2022-03-29
Bid with DAI logic

valiafetisov:
Contributes to #160 

In order to test:
1. Restart an auction, note down the id, eg `ETH-A:120`
2. Make sure you have enough DAI in the VAT (more than `Auction Price Total`)
   - open wallet modal via `$nuxt.$store.dispatch("modals/setWalletModalAndFetchData", true)`
   - deposit enough DAI
3. Participate using correct auction id
    ```
    id = 'ETH-A:120'; $nuxt.$store.dispatch("auctions/bidWithDai", { id, bidAmountDai: $nuxt.$store.state.auctions.auctionStorage[id].totalPrice.toFixed() })
    ```

Note: this will bid total amount.

### DIFF
--- a/bot-twitter/src/keeper.ts
+++ b/bot-twitter/src/keeper.ts
@@ -1,6 +1,6 @@
 import { AuctionInitialInfo } from 'auctions-core/src/types';
 import getSigner, { createSigner, setSigner } from 'auctions-core/src/signer';
-import { bidOnTheAuction, enrichAuction } from 'auctions-core/src/auctions';
+import { swapTheAuction, enrichAuction } from 'auctions-core/src/auctions';
 import {
     authorizeCollateral,
     authorizeWallet,
@@ -127,7 +127,7 @@ const checkAndParticipateIfPossible = async function (auction: AuctionInitialInf
 
     // bid on the Auction
     console.info(`keeper: auction "${auctionTransaction.id}": attempting swap execution`);
-    const bidHash = await bidOnTheAuction(ETHEREUM_NETWORK, auctionTransaction, walletAddress);
+    const bidHash = await swapTheAuction(ETHEREUM_NETWORK, auctionTransaction, walletAddress);
     console.info(`keeper: auction "${auctionTransaction.id}" was succesfully executed via "${bidHash}" transaction`);
 };
 
--- a/bot-twitter/src/keeper.ts
+++ b/bot-twitter/src/keeper.ts
@@ -1,6 +1,6 @@
 import { AuctionInitialInfo } from 'auctions-core/src/types';
 import getSigner, { createSigner, setSigner } from 'auctions-core/src/signer';
-import { swapTheAuction, enrichAuction } from 'auctions-core/src/auctions';
+import { bidWithCallee, enrichAuction } from 'auctions-core/src/auctions';
 import {
     authorizeCollateral,
     authorizeWallet,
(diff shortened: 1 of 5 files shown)


fegbert (approved):

LukSteib (commented):
* I was able to bid on an auction the way described in the initial comment. See: https://kovan.etherscan.io/tx/0xa626deff5b45a71a3210e75514e48d27d00883ed8dd06f4cec81bfb8763cd107
* Dai was withdrawn from VAT 
* However, I was expecting to receive the 0.1 ETH (= `Auction amount`) in my wallet. This is not true. AFAIK the ETH wallet balance did not change
  * Same for execution of a `BAL` auction -> `BAL` funds did not end up in wallet

LukSteib:
> Therefore, we may actually need initially proposed "collateral withdrawal" step 🙁

:( yes, seems like it. So we definitely need a solution for a user to get the funds in the wallet


> Anyway, take input data is correct: who has your address, therefore I suppose this functionality is implemented correctly on our side

Yes. Is there an easy command to fetch VAT balance for a certain collateral? Similar to the one we use for Dai. Just in order to doublecheck that collateral ends up in VAT associated with wallet address

valiafetisov:
> Is there an easy command to fetch VAT balance for a certain collateral?

Not really. Here is [where they are stored in vat contract](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/vat.sol#L58). So in order to get this number, you need to:
- open vat contract https://kovan.etherscan.io/address/0xba987bdb501d131f766fee8180da5d81b34b69d9#readContract
- execute `gem` read function, passing
  - `ilk`, bytes32 encoded collateral type [according to the source code](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/clip.sol#L62)
      - can be found in the clipper contract https://kovan.etherscan.io/address/0x7dd1fb6b9afdba9f28db89c81723b8c6b27a2fbe#readContract
      - for `ETH-A` it's `0x4554482d41000000000000000000000000000000000000000000000000000000 `
  - `address` – your wallet address `0x4f3b7fb6204c614b28f0e1f50dc1c99175631678`
- convert [returned number (`59083106099127643`)](https://user-images.githubusercontent.com/3393626/160643204-cb981b0f-8f5d-4087-99fd-8c1efcab22ff.png) from `wad` (ie move 18 decimal points left) via `new BigNumber('59083106099127643').shiftedBy(-18).toFixed()`
  - the result is `0.059083106099127643`

So it's actually less than `0.1`, most probably because you don't actually get the whole amount (just like with the swap) and profit calculation should be done depending on the debt: https://github.com/sidestream-tech/unified-auctions-ui/blob/5494c084e8f42cbdb962e39651b136c3dda196a6/core/src/price.ts#L44-L55

But I think we should open another PR for both implementing `gem` fetching and fixing "profit" calculation

LukSteib:
> So it's actually less than 0.1, most probably because you don't actually get the whole amount (just like with the swap) and profit calculation should be done depending on the debt

Yes, with the caveat that in the case of Dai bidding it also affects to amount to bid. So I assume the total amount to bid is not always `auction amount` * `auction price` :( 


So going to approve this PR and we need to figure out `gem` and profit / total amount to bid related parts in other PRs

valiafetisov:
> So I assume the total amount to bid is not always `auction amount` * `auction price`

No, that's impossible because `take` actually accepts [`amt: Upper limit on amount of collateral to buy`](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/clip.sol#L329) (and not DAI that user enters). In other words, user can only take auction partially, but can not bid more than required. The actual amount to be transferred to the user (`slice`) is calculated based on the `tab`, `amt` and current price here: https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/clip.sol#L357-L380

LukSteib (approved):

Issue #165, opened 2022-03-29
Update Bidding with DAI `the amount to receive` logic


valiafetisov:
### Goal

Fixed `amount to receive` indication in the BidTransactionTable

### Context

After executing first Bid with DAI transaction, we realised that the user doesn't receive the whole amount of collateral. More info here: https://github.com/sidestream-tech/unified-auctions-ui/pull/163#issuecomment-1082005735. Based on this, we need to:
- [x] Create new `core` function which will calculate actual `amount to receive` based on the input value
- [x] Add store logic to fetch collateral in the VAT
- [x] Add store logic to withdraw collateral from VAT to Wallet

### Tasks
- [x] Implement all the above
- [x] Provide testing commands in the PR description

PR #167, opened 2022-03-30
Collateral profit, balance and withdrawal logic

valiafetisov:
Closes https://github.com/sidestream-tech/unified-auctions-ui/issues/165

### Functional testing

1. Fetch and print collateral VAT balance:
```js
collateralType = 'ETH-A'; await $nuxt.$store.dispatch("wallet/fetchCollateralVatBalance", collateralType); $nuxt.$store.state.wallet.collateralVatBalanceStore[collateralType].toFixed();
```

2. Withdraw all collateral type
```js
collateralType = 'ETH-A'; await $nuxt.$store.dispatch("wallet/withdrawAllCollateralFromVat", collateralType); $nuxt.$store.state.wallet.collateralVatBalanceStore[collateralType].toFixed();
```

Note: 
- in case of ETH auctions, in the end user gets WETH, which they then have to unwrap themselves
- I'm hoping that this withdrawal step doesn't require user to approve another `MCD_JOIN_${collateralType}` contract, but we need to check that as well

### DIFF
--- a/core/src/abis/MCD_JOIN.json
+++ b/core/src/abis/MCD_JOIN.json
@@ -1,0 +1,236 @@
+[
+    {
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "vat_",
+                "type": "address"
+            },
+            {
+                "internalType": "bytes32",
+                "name": "ilk_",
+                "type": "bytes32"
+            },
+            {
+                "internalType": "address",
+                "name": "gem_",
+                "type": "address"
+            }
+        ],
+        "payable": false,
+        "stateMutability": "nonpayable",
+        "type": "constructor"
+    },
+    {
+        "anonymous": true,
+        "inputs": [
+            {
+                "indexed": true,
+                "internalType": "bytes4",
+                "name": "sig",
+                "type": "bytes4"
+            },
+            {
+                "indexed": true,
+                "internalType": "address",
+                "name": "usr",
+                "type": "address"
+            },
+            {
+                "indexed": true,
+                "internalType": "bytes32",
+                "name": "arg1",
+                "type": "bytes32"
+            },
+            {
+                "indexed": true,
+                "internalType": "bytes32",
+                "name": "arg2",
+                "type": "bytes32"
+            },
+            {
+                "indexed": false,
+                "internalType": "bytes",
+                "name": "data",
+                "type": "bytes"
+            }
+        ],
+        "name": "LogNote",
+        "type": "event"
+    },
+    {
+        "constant": false,
+        "inputs": [],
+        "name": "cage",
+        "outputs": [],
+        "payable": false,
+        "stateMutability": "nonpayable",
+        "type": "function"
+    },
+    {
+        "constant": true,
+        "inputs": [],
+        "name": "dec",
+        "outputs": [
+            {
+                "internalType": "uint256",
+                "name": "",
+                "type": "uint256"
+            }
+        ],
+        "payable": false,
+        "stateMutability": "view",
+        "type": "function"
+    },
+    {
+        "constant": false,
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "usr",
+                "type": "address"
+            }
+        ],
+        "name": "deny",
+        "outputs": [],
+        "payable": false,
+        "stateMutability": "nonpayable",
+        "type": "function"
+    },
+    {
+        "constant": false,
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "usr",
+                "type": "address"
+            },
+            {
+                "internalType": "uint256",
+                "name": "wad",
+                "type": "uint256"
+            }
+        ],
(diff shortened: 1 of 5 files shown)


fegbert (dismissed):

LukSteib (changes requested):
* Withdrawal of `BAL-A` was successfull (no further authorisation step was needed) ✅ 
* Withdrawal of `ETH-A` (in the form of `WETH`)  successfull (no further authorisation step was needed) ✅ 
  * We should definitely inform the user in the future about the fact of `WETH` instead of `ETH` (but no need tackle now -> will put it as a note in the backlog) 
* Fetching the balances in the VAT also works ✅ 

**Regarding calculations:**
Case 1:
image

What makes me wonder in the example case above (a recently restarted `LINK` auction)
* How come that the `Minimum bid` is way higher than the `Total price`
* User Bid Input is prevented right from the beginning with message "...leftover part will be too small"  
  * Is this due to the fact that the leftover link in this case falls below dust (or another parameter) How do we determine that?

Case 2:
https://www.loom.com/share/d5ffc8747e2c49e1b4ee9657be6f5e0f

* _untrelated:_ minimum bid selection is applied correctly but not rendered properly in input field
* there is no difference in the collateralOutcome we calculate when placing maximum or minimum bid. Without looking deeper in the calculation this feels wrong
  * I placed the minimum bid -> the calculated collateral amount was correct (checked VAT after transaction) (see loom 2 below)

Case 3: 
image

* for a 10 ETH auction (see case 2) I was able to adjust the bid. For the 15.5 ETH auction at hand I am not able to change it. Why?

---
loom 2 (calculation to actual outcome comparison)
https://www.loom.com/share/7d43100b14d243b6b18aaa22daf262d5

valiafetisov:
- I've changed minimum value to be `chost` as in the contract code, it's not significantly lower
- I've changed maximum value to be auction debt
- I've fixed various things including swap transaction execution before it's profitable

You can now execute an auction and compare the received amount with the calculated one.

<details><summary>It worked for me</summary>

Screenshot 2022-03-31 at 09 15 56
Screenshot 2022-03-30 at 22 31 31

</details>


The collateral withdrawal block will be implemented in a separate PR (if this one is merged before that)

LukSteib:
I am still in doubt regarding the values we are setting. Again have to describe some cases that just don't feel right to me. Tested on `goerli`:

**Case 1: 1. `ETH-A` auction**
image


* maximum bid and minimum bid are the same
* only when clicking `set maximum bid` there seems to be a valid input
* for all other cases (manually inserting an amount or clicking `set minimum bid`) shows error as seen in screenshot
  * tooltip values seem to be off "between XX and 0"

---

**Case 2: 2. `ETH-A` auction**
image

* Compared to case 1, why am I for this auction not even able to set any bid? 

---

**Case 3: `MATIC` auction**
image

* minimum bid is higher than maximum bid
* for all cases other than `set maximum bid` an error is shown and the tooltip message is off

valiafetisov:
> only when clicking set maximum bid there seems to be a valid input
> for all other cases (manually inserting an amount or clicking set minimum bid) shows error as seen in screenshot

Sorry, there was a bug in the determination if the input should be enabled or disabled. Now it's disabled for all auctions from the above.

> Compared to case 1, why am I for this auction not even able to set any bid?

That's correct. You are not able to set the bid because the leftover will be too small. But probably this needs explanation like we had before (a tooltip with the text `Not possible to change the amount since the leftover will be smaller than minimum`)

Screenshot 2022-03-31 at 12 21 29

> minimum bid is higher than maximum bid

I think that's a valid situation, when debt is smaller than minimum bid. As you may remember, `minimum bid` is the solution for the situation when user can exploit the system by creating many small vaults and getting a bigger fixed reward for kicking them. So to clear the confusion with bid being smaller than minimum, we may change the wording of `minimum bid`, but it's hard to explain what it actually is. Do you have any suggestions or should we keep it as is?

LukSteib:
> Now it's disabled for all auctions from the above.

Is it now disabled for basically all auctions? Cannot find an auction where it is possible any longer

> Do you have any suggestions or should we keep it as is?

I do not have a suggestion that doesn't cause more confusion. Tbh I haven't seen this case after you have fixed the bug.

--- 
- [ ] Bid input and indication on bid button do no longer match (see below)
image

valiafetisov:
> Is it now disabled for basically all auctions? Cannot find an auction where it is possible any longer

On kovan there are 2 that are both partially biddable, eg http://localhost:3000/collateral?network=kovan&auction=LINK-A%3A201

> Bid input and indication on bid button do no longer match

Ohhh, fixed it, as well as auction finished state.

LukSteib:
> Ohhh, fixed it, as well as auction finished state.

Confirm, fixed.

---

So was able to bid on an auction with `maximum bid` -> right outcome was indicated.
Was able to withdraw collateral via new block.

LukSteib (approved):

PR #178, opened 2022-04-04
`AllowanceAmountCheckPanel` and `WalletAuthorizationCheckPanel` stories

### LINKED ISSUE #174: Create stories for `AllowanceAmountCheckPanel` and `WalletAuthorizationCheckPanel`
### Goal
Stories that cover all edge cases

### Context
New components were implemented in the last sprint, but the stories are missing. This issue is about implementing the stories to check all possible edge cases and also learn how the components are implemented. 

Please note that every prop that receives BigNumber, should have a story with `NaN` and undefined. We should also check that `disabled` and `isLoading` works properly in all combinations. In the end I expect there is like 10+ stories for each component with every combination that makes sense.

### Tasks
- [x] create stories for `AllowanceAmountCheckPanel`
- [x] create stories for `WalletAuthorizationCheckPanel`
- [x] make sure `isCorrect` is handled via action

fegbert:
closes #174 

### `AllowanceAmountCheckPanel` stories
- Desired amount is within allowance
- Desired amount exceeds allowance
- Missing allowance amount (missing wallet)
- Missing desired amount
- Loading
- Disabled
- Expert Mode

### `WalletAuthorizationCheckPanel` stories
- Wallet is authorized
- Wallet is not authorized yet
- Wallet is missing
- Loading
- Disabled
- Expert Mode

--- 

Is the `unlimited access to DAI` functionality not added yet or why is it just setting to `undefined`?

And is the `isUnlimitedAllowance` stuff in the `AllowanceAmountCheckPanel` even necessary? Since the Panel will always be "minimized" when it's in the `correct` state and if the allowance is `unlimited`, `desired` can never be more so the panel is never in a state where it shows this information, right?

### DIFF
--- a/frontend/components/panels/AllowanceAmountCheckPanel.stories.js
+++ b/frontend/components/panels/AllowanceAmountCheckPanel.stories.js
@@ -51,23 +51,6 @@ storiesOf('Panels/AllowanceAmountCheckPanel', module)
             @setAllowanceAmount="setAllowanceAmount"
         />`,
     }))
-    .add('Unlimited Allowance', () => ({
-        ...common,
-        data() {
-            return {
-                allowanceAmount: new BigNumber('99999999999999999999999999999999999999999999999999'),
-                desiredAmount: new BigNumber(faker.finance.amount(0, 100)),
-                isLoading: false,
-            };
-        },
-        template: `
-        <AllowanceAmountCheckPanel
-            :allowance-amount="allowanceAmount"
-            :desired-amount="desiredAmount"
-            :is-loading="isLoading"
-            @setAllowanceAmount="setAllowanceAmount"
-        />`,
-    }))
     .add('Missing Allowance Amount', () => ({
         ...common,
         data() {
--- a/frontend/components/panels/AllowanceAmountCheckPanel.stories.js
+++ b/frontend/components/panels/AllowanceAmountCheckPanel.stories.js
@@ -10,17 +10,28 @@ const common = {
             allowanceAmount: new BigNumber(faker.finance.amount(0, 100)),
             desiredAmount: new BigNumber(faker.finance.amount(100, 200)),
             isLoading: false,
+            disabled: false,
+            isExplanationsShown: true,
         };
     },
     methods: {
         setAllowanceAmount(amount) {
             this.isLoading = true;
             setTimeout(() => {
-                this.allowanceAmount = amount;
+                this.allowanceAmount = amount || new BigNumber(Number.MAX_VALUE);
                 this.isLoading = false;
             }, 1000);
         },
     },
+    template: `
+        <AllowanceAmountCheckPanel 
+            :allowance-amount="allowanceAmount" 
+            :desired-amount="desiredAmount"
+            :is-loading="isLoading"
+            :disabled="disabled"
+            :isExplanationsShown="isExplanationsShown"
+            @setAllowanceAmount="setAllowanceAmount" 
+        />`,
 };
 
 storiesOf('Panels/AllowanceAmountCheckPanel', module)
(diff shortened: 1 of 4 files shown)


valiafetisov:
> Is the `unlimited access to DAI` functionality not added yet or why is it just setting to `undefined`?

If the store action is called with undefined it will be set to the maximum possible value (which is kind of unlimited, since it's bigger than total amount of DAI). So it's implemented and works. So the number should have more than 50 integers to be unlimited as you can see in the [isUnlimitedAllowance](https://github.com/sidestream-tech/unified-auctions-ui/blob/1c5aa7609d0fe065f9576339445be5e4771a7241/frontend/components/panels/AllowanceAmountCheckPanel.vue#L94).

Would you please add a story for that too?

> And is the `isUnlimitedAllowance` stuff in the `AllowanceAmountCheckPanel` even necessary? Since the Panel will always be "minimized" when it's in the `correct` state and if the allowance is `unlimited`, `desired` can never be more so the panel is never in a state where it shows this information, right?

Minimised state can always be expanded by clicking on the panel title. For example if the user is curious about exact allowance number. But since it's not primary action for the flow in this case all buttons should be `secondary` to not attract user attention too much 🙂

frontend/components/panels/AllowanceAmountCheckPanel.stories.js
```diff
@@ -0,0 +1,119 @@
...
+                isLoading: false,
+            };
+        },
+        template: `
+        <AllowanceAmountCheckPanel 
+            :allowance-amount="allowanceAmount" 
+            :desired-amount="desiredAmount"
+            :is-loading="isLoading"
+            @setAllowanceAmount="setAllowanceAmount" 
+        />`,
```
valiafetisov, line 42: If template stays unchanged over the stories, you can move it to the `common` object above and define only once

frontend/components/panels/WalletAuthorizationCheckPanel.stories.js
```diff
@@ -0,0 +1,102 @@
...
+                isLoading: false,
+            };
+        },
+        template: `
+        <WalletAuthorizationCheckPanel
+            :is-wallet-authorized="isWalletAuthorized"
+            :wallet-address="walletAddress"
+            :is-loading="isLoading"
+            @authorizeWallet="authorize"
+        />`,
```
valiafetisov, line 41: Same here, `isLoading` and `disabled` can be part of the `data`

frontend/components/panels/AllowanceAmountCheckPanel.stories.js
```diff
@@ -0,0 +1,119 @@
...
+    }))
+    .add('Missing Allowance Amount', () => ({
+        ...common,
+        data() {
+            return {
+                allowanceAmount: undefined,
+                desiredAmount: new BigNumber(faker.finance.amount(0, 100)),
+                isLoading: false,
+            };
+        },
```
valiafetisov, line 62: To make it a little more readable, please use:
```suggestion
        data: () => ({
           ...common.data(),
            allowanceAmount: undefined,
        }),
```
To only replace values that you need to replace. Here and everywhere else 🙂

valiafetisov (commented):
- [ ] make sure `isCorrect` is handled via action (as per issue)
- [ ] please add a story with `desiredAmount` being NaN
- [ ] please add a story with `allowanceAmount` being unlimited (see comment above)

frontend/components/panels/AllowanceAmountCheckPanel.stories.js
fegbert: Oops I tried that but forgot the brackets behind `data` 🤦‍♂️

valiafetisov (approved):

LukSteib (changes requested):
re: `AllowanceAmountCheckPanel`
- [x] For the case "Within Allowance" (ie. allowance high enough but no unlimited allowance was given) I would expect the secondary button to set specific allowance value to be disabled. IMO it doesn't make sense to set back the allowance to a lower number (opinions?!)
- [x] For the disabled states of this Panel (ie. stories "Missing Allowance Amount", "Missing Desired Amount", "Desired Amount is NAN") I would either expect them to be not expandable at all or at least having the buttons disabled (is there any consideration I am missing?)

re: `WalletAuthorizationCheckPanel`
- [ ] What is the reasoning behind trimming down the existing text for that part significantly?
- [ ] As above: For the disabled states of this Panel (ie. story "Missing Wallet" I would either expect them to be not expandable at all or at least having the buttons disabled (is there any consideration I am missing?)

valiafetisov:
> I would either expect them to be not expandable at all or at least having the buttons disabled (is there any consideration I am missing?)

I think there is no need to disable the buttons as soon as they are hidden/panel is collapsed. The user should be able to edit those values even if it's not needed for this case. The only rule I tried to hold true: if this step is not required it shouldn't have `primary` style.

> IMO it doesn't make sense to set back the allowance to a lower number (opinions?!)

I guess it does if user don't want the contract to have the ability to withdraw the money after the transaction. As above, the only rule I would hold true: to not use `primary` style if this step is not needed, since it will be collapsed anyway.

> re: `WalletAuthorizationCheckPanel` What is the reasoning behind trimming down the existing text for that part significantly?

This block is currently used in the wallet popup and will be used in the both transaction pages, therefore it should be universal and can not talk about auctions. Better proposal are welcome!

LukSteib:
> The only rule I tried to hold true: if this step is not required it shouldn't have primary style

ok fine if we stick to this "golden" rule only. So if against expectation a user would still expand the collapsed panel and hit the button, we would just through an error I assume. Because if no wallet is connected no allowance can be give (see [1] below for reference) 

> This block is currently used in the wallet popup and will be used in the both transaction pages, therefore it should be universal and can not talk about auctions

True, forgot that we also display it on WalletPopup where the auction reference doesn't make sense. Then I only propose:
- [ ] Grammar: Make one comma separated sentence out of the two explainer sentences for `WalletAuthorizationCheckPanel`

[1]
image

LukSteib (approved):

valiafetisov (approved):

PR #185, opened 2022-04-05
`WalletDaiDepositCheckPanel` and `WalletVatDaiWithdrawCheckPanel ` stories

### LINKED ISSUE #175: Create stories for `WalletDaiDepositCheckPanel` and `WalletVatDaiWithdrawCheckPanel`
### Goal
Stories that cover all edge cases

### Context
New components were implemented in the last sprint, but the stories are missing. This issue is about implementing the stories to check all possible edge cases and also learn how the components are implemented. 

Please note that every prop that receives BigNumber, should have a story with `NaN` and undefined. We should also check that `disabled` and `isLoading` works properly in all combinations. In the end I expect there is like 10+ stories for each component with every combination that makes sense.

### Tasks
- [ ] create stories for `WalletDaiDepositCheckPanel`
- [ ] create stories for `WalletVatDaiWithdrawCheckPanel`
- [ ] make sure `isCorrect` is handled via action

benediktkaiser:
Closes #175 

### `WalletDaiDepositCheckPanel` stories
- Insufficient amount of DAI present
- Sufficient amount of DAI present
- Missing Desired Amount
- Desired Amount is NaN
- Desired Amount is negative
- No wallet connected
- Loading
- Disabled
- Expert Mode

### `WalletVatDaiWithdrawCheckPanel` stories
- Insufficient amount of DAI present
- Sufficient amount of DAI present
- Missing Desired Amount
- Desired Amount is NaN
- Desired Amount is negative
- No wallet connected
- Loading
- Disabled
- Expert Mode

### DIFF
--- a/frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
+++ b/frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
@@ -53,18 +53,18 @@ storiesOf('Panels/WalletDaiDepositCheckPanel', module)
             desiredAmount: undefined,
         }),
     }))
-    .add('Desired Amount is negative', () => ({
-        ...common,
-        data: () => ({
-            ...common.data(),
-            desiredAmount: new BigNumber(-10),
-        }),
-    }))
     .add('Desired Amount is NaN', () => ({
         ...common,
         data: () => ({
             ...common.data(),
             desiredAmount: new BigNumber(NaN),
+        }),
+    }))
+    .add('Desired Amount is negative', () => ({
+        ...common,
+        data: () => ({
+            ...common.data(),
+            desiredAmount: new BigNumber(-10),
         }),
     }))
     .add('No wallet connected', () => ({
--- a/frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
+++ b/frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
@@ -1,0 +1,97 @@
+import { storiesOf } from '@storybook/vue';
+import { action } from '@storybook/addon-actions';
+import faker from 'faker';
+import BigNumber from 'bignumber.js';
+import WalletDaiDepositCheckPanel from '~/components/panels/WalletDaiDepositCheckPanel';
+
+const common = {
+    components: { WalletDaiDepositCheckPanel },
+    data() {
+        return {
+            walletDai: new BigNumber(faker.finance.amount(0, 100)),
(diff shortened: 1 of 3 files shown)


benediktkaiser:
I noticed one small wording issue in the `WalletVatDaiWithdrawCheckPanel`. When the state is set to correct and you manually extend the panel it still gives you the following message:

Screenshot 2022-04-05 at 10 43 17

On other panels, the error messages are formatted in a more neutral way, that allow them to be displayed both during the error and success states. Do we maybe want to reword this to follow in the other panels footprints, or do we add some conditional text? (or just leave it like this 😄)

valiafetisov:
> When the state is set to correct and you manually extend the panel it still gives you the following message:

Well, that's exactly the reason to have stories. Can you please fix the wording or do you need a suggestion?

valiafetisov:
Maybe something like this?

```
If you do not have enough DAI funds in the VAT, you will need to deposit them, before you can continue.
```

frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
```diff
@@ -0,0 +1,97 @@
...
+            ...common.data(),
+            desiredAmount: new BigNumber(NaN),
+        }),
+    }))
+    .add('Desired Amount is negative', () => ({
+        ...common,
+        data: () => ({
+            ...common.data(),
+            desiredAmount: new BigNumber(-10),
+        }),
```
valiafetisov, line 68: I wonder if this is actually properly handled by the panel 😅

valiafetisov (approved):

frontend/components/panels/WalletDaiDepositCheckPanel.stories.js
benediktkaiser: It is! That is also why I added the story 😄 

See: https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/panels/WalletDaiDepositCheckPanel.vue#L87

valiafetisov: I remember I did it somewhere, but I don't remember where exactly 🔍

benediktkaiser: If I ever have some free time, I might go through and see if we are missing stories for any other components. 

I agree that a higher coverage rate would be nice to have.

PR #186, opened 2022-04-05
Add `WalletConnectionCheckPanel`

### LINKED ISSUE #176: Implement new `WalletConnectedCheckPanel`
### Goal

New panel that checks wallet connection

### Context
Since we're moving to the universal panel approach, we should make a new component that checks "if a wallet is connected" and "display actions to connect it in case it's not". The states of the panel can be on [the mock](https://github.com/sidestream-tech/unified-auctions-ui/issues/145#issuecomment-1070902434) and on the existing WalletBlock, but please edit the texts to be be universal so the panel can be reused in many different places regardless of the context. This particular panel will be used in:
- bid transaction page
- swap transaction page
- reset auction
- wallet popup

### Tasks
- [ ] create the panel
- [ ] add all possible stories with all combination of props

fegbert:
closes #176 

image

image

### DIFF
--- a/frontend/components/panels/WalletConnectionCheckPanel.stories.js
+++ b/frontend/components/panels/WalletConnectionCheckPanel.stories.js
@@ -1,0 +1,72 @@
+import { storiesOf } from '@storybook/vue';
+import { action } from '@storybook/addon-actions';
+import faker from 'faker';
+import WalletConnectionCheckPanel from './WalletConnectionCheckPanel';
+
+const common = {
+    components: { WalletConnectionCheckPanel },
+    data: () => ({
+        walletAddress: null,
+        disabled: false,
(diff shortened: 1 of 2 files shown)


valiafetisov:
Let's keep texts similar to the WalletBlock:

```
To execute transactions, you need to connect a wallet. If you don’t have a wallet yet, you can sign up for it on a service like <a href="https://metamask.io/" target="_blank">MetaMask</a>.
```
and if the wallet is connected:
```
All transactions will be carried over the connected wallet <format-address shorten :value="walletAddress" type="address" />
```

LukSteib:
> Let's keep texts similar to the WalletBlock:

Agree to this proposal

valiafetisov (approved):

PR #187, opened 2022-04-05
Added `AuctionRestartPanel`

### LINKED ISSUE #182: Implement new `AuctionRestartPanel`

benediktkaiser:
closes #182 

Screenshot 2022-04-05 at 15 00 46

Screenshot 2022-04-05 at 15 01 21

Currently blocked by #186


valiafetisov:
From my opinion, currently there is too much repeating / spread information. I imagine it like this:
- Auction is active = nothing is shown
- Auction is inactive
  - We don't show error `alert` for this state anymore
  - We show new restart panel, which has error state `This auction is inactive and must be restarted`
     - Inside the panel, we have another panel which checks wallet (please don't actually put it inside the restart panel, but rather provide it via slot)

MakerDAO@2x (3)


Would it make sense?

LukSteib:
> Would it make sense or is it still too much?

To me it makes a lot of sense! I like to proposal to keep all of the restart related information in one panel that includes the walletcheckpanel. Reducing information further would make it unintuitive again. Especially since inactive auctions (especially on mainnet) are rather an edge case an hence need proper explanation.

benediktkaiser:
New Design: 

Screenshot 2022-04-07 at 14 07 17

LukSteib (approved):

frontend/components/Auction.vue
```diff
@@ -1,17 +1,21 @@
 <template>
     <TextBlock :title="`Auction #${auctionId}`">
-        <Alert v-if="auctionError" class="my-3" :message="auctionError" type="error" />
+        <div class="my-3">
+            <Alert v-if="auctionError && auctionError.showBanner" :message="auctionError.error" type="error" />
+        </div>
         <div v-if="auction">
-            <RestartBlock
-                v-if="auctionError === 'This auction is inactive and must be restarted'"
+            <AuctionRestartPanel
```
valiafetisov, line 7: I think to set rounded corners on the bottom, this panel should be set into its own `div`

frontend/components/Auction.vue
```diff
@@ -1,17 +1,21 @@
...
+                :disabled="!walletAddress"
```
valiafetisov, line 8: You're already passing `walletAddress` as a prop, so it's disabled state can be determined based on the `walletAddress`, which you already doing by the way. (but of course you can keep disabled prop in case we need to disable it even if there is a wallet)

frontend/components/Auction.vue
```diff
@@ -1,17 +1,21 @@
...
+                :auction-is-active="auction.isActive"
```
valiafetisov, line 9: I suppose that you don't need to pass `isActive` as a prop since the panel shouldn't be shown if the action is active, right? (in contrast to other panels, so maybe it's not that obvious)
```suggestion
                v-if="auction.isActive"
```

frontend/components/panels/AuctionRestartPanel.vue
```diff
@@ -44,55 +39,98 @@
...
+            return this.disabled || !!this.auctionIsActive || !this.walletAddress;
+        },
+    },
+    watch: {
+        currentStateAndTitle: {
+            immediate: true,
+            handler(newCurrentStateAndTitle) {
+                this.$emit('update:isCorrect', newCurrentStateAndTitle.name === 'correct');
+            },
+        },
```
valiafetisov, line 133: I guess this panel doesn't need that one, since it's unused

valiafetisov (approved):

PR #190, opened 2022-04-06
Added `CollateralAuthorizationCheckPanel`

### LINKED ISSUE #179: Implement new `CollateralAuthorizationCheckPanel`
### Goal

New panel that checks authorization for collateral type

### Context
Since we're moving to the universal panel approach, we should make a new component that checks "if the collateral type of the auction at hand was already authorized for the wallet" and "display actions to connect it in case it's not". The states of the panel can be found on [the mock](https://github.com/sidestream-tech/unified-auctions-ui/issues/145#issuecomment-1070902434) and on the existing AuthorisationBlock, but please edit the texts to be be universal so the panel can be reused in many different places regardless of the context. This particular panel will be used in:
- bid transaction page
- swap transaction page

### Tasks
- [x] create the panel

fegbert:
closes #179 

**no wallet**
image

**unauthorized**
image

**authorized**
image

### DIFF
--- a/frontend/components/panels/CollateralAuthorizationCheckPanel.stories.js
+++ b/frontend/components/panels/CollateralAuthorizationCheckPanel.stories.js
@@ -1,0 +1,86 @@
+import { storiesOf } from '@storybook/vue';
+import { action } from '@storybook/addon-actions';
+import faker from 'faker';
+import CollateralAuthorizationCheckPanel from './CollateralAuthorizationCheckPanel';
+import { generateFakeAuctionTransaction } from '~/helpers/generateFakeAuction';
+
+const fakeAuctionTransaction = generateFakeAuctionTransaction();
+
+const common = {
+    components: { CollateralAuthorizationCheckPanel },
+    data: () => ({
+        collateralType: fakeAuctionTransaction.collateralType,
+        authTransactionFeeETH: fakeAuctionTransaction.authTransactionFeeETH,
+        walletAddress: faker.finance.ethereumAddress(),
+        isCollateralAuthorized: false,
+        isWalletAuthorized: true,
+        isLoading: false,
+        disabled: false,
+        isExplanationsShown: true,
+    }),
+    methods: {
+        authorizeCollateral() {
+            this.isLoading = true;
+            setTimeout(() => {
+                this.isCollateralAuthorized = true;
+                this.isLoading = false;
+            }, 1000);
+        },
+        isCorrect: action('isCorrect'),
+    },
+    template: `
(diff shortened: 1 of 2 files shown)


frontend/components/panels/CollateralAuthorizationCheckPanel.stories.js
```diff
@@ -0,0 +1,87 @@
...
+const fakeAuctionTransaction = generateFakeAuctionTransaction();
+
+const common = {
+    components: { CollateralAuthorizationCheckPanel },
+    data: () => ({
+        collateralType: fakeAuctionTransaction.collateralType,
+        authTransactionFeeETH: new BigNumber(fakeAuctionTransaction.authTransactionFeeETH),
+        walletAddress: faker.finance.ethereumAddress(),
+        isCollateralAuthorized: false,
+        isWalletAuthorized: true,
```
valiafetisov, line 17: You don't need to pass and to check `isWalletAuthorized` inside this component, since the panels suppose to have isolated logic and only check one thing. Therefore, please remove this prop and stories related to it

benediktkaiser: I also removed all mentions and usages of `walletAddress`, as it was not really needed. 

If no wallet is connected, we can just pass `disabled` to the component and it would have done the same thing, which is a bit easier to manage on the long run.

frontend/components/panels/CollateralAuthorizationCheckPanel.vue
```diff
@@ -79,12 +71,6 @@ export default Vue.extend({
     },
     computed: {
         currentStateAndTitle(): PanelProps {
-            if (!this.walletAddress) {
-                return {
-                    name: 'inactive',
-                    title: `The ${this.collateralType} authorization is unknown until a wallet is connected`,
-                };
-            }
```
valiafetisov: Without connected wallet we can not get collateral auth status, as it's done per wallet. So this was actually correct

But there is no interdependency between `isCollateralAuthorized` and `isWalletAuthorized`, therefore my previous comment was to remove `isWalletAuthorized` from props and the logic, not `walletAddress`

frontend/components/panels/CollateralAuthorizationCheckPanel.stories.js
valiafetisov: > we can just pass `disabled` to the component and it would have done the same thing

Passing disabled unfortunately wouldn't trigger `inactive` state 😞

valiafetisov (approved):

LukSteib (approved):

Issue #181, opened 2022-04-04
Implement new `NetProfitCheckPanel`


LukSteib:
### Goal

New panel that checks net profit of transaction

### Context
Since we're moving to the universal panel approach, we should make a new component that checks "if transaction net profit is positive" and "warn user to execute with negative net profit". One state of the panel can be found on [the mock](https://github.com/sidestream-tech/unified-auctions-ui/issues/145#issuecomment-1070902434).
States needed (please extend list if you come up with more relevant ones):
* net profit positive
* net profit negative
* Loading
* Unknown
* Expert mode

Please edit the texts to be be universal so the panel can be reused in many different places regardless of the context. This particular panel will be used in:
- swap transaction page

Please refer to the existing panels `AllowanceAmountCheckPanel`, `WalletAuthorizationCheckPanel`, `WalletDaiDepositCheckPanel` and `WalletVatDaiWithdrawCheckPanel` for the inspiration. Please note that the panel should emit valid `isCorrect` value on state change. Please add stories that cover all potential edge cases.

### Tasks
- [x] create the panel
- [x] add all possible stories with all combination of props
- [x] edit / propose new text and titles for different states in the code itself

PR #191, opened 2022-04-06
Added `GrossProfitCheckPanel`/`NetProfitCheckPanel` and expanded `BasePanel` functionality

### LINKED ISSUE #180: Implement new `GrossProfitCheckPanel`
### Goal

New panel that checks gross profit of transaction

### Context
Since we're moving to the universal panel approach, we should make a new component that checks "if transaction gross profit is positive" and "prevent user from being able to execute if gross profit is negative". One state of the panel can be found on [the mock](https://github.com/sidestream-tech/unified-auctions-ui/issues/145#issuecomment-1070902434).
States needed (please extend list if you come up with more relevant ones):
* gross profit positive
* gross profit negative
* Unknown
* Expert mode

Please edit the texts to be be universal so the panel can be reused in many different places regardless of the context. This particular panel will be used in:
- swap transaction page

### Tasks
- [x] create the panel
- [x] edit / propose new text and titles for different states in the code itself

benediktkaiser:
closes #180 #181

Screenshot 2022-04-06 at 12 13 33

Screenshot 2022-04-06 at 12 35 18

I also added a `disabled` prop to the BasePanel. This is useful for when you have a text only panel (with no call to action), which would result in just an empty expand in the expert mode. When you set disabled to true, it will disable the expanding of the panel. I used it in this component, to disable the expanding in the expert mode.

### DIFF
--- a/frontend/components/panels/GrossProfitCheckPanel.vue
+++ b/frontend/components/panels/GrossProfitCheckPanel.vue
@@ -1,5 +1,5 @@
 <template>
-    <BasePanel :current-state="currentStateAndTitle.name" :disabled="!isExplanationsShown">
+    <BasePanel :current-state="currentStateAndTitle.name">
         <template #title> {{ currentStateAndTitle.title }} </template>
         <TextBlock>
             You cannot execute an auction while it is not yet profitable. Once the auction price drops below the price
--- a/frontend/components/panels/GrossProfitCheckPanel.vue
+++ b/frontend/components/panels/GrossProfitCheckPanel.vue
@@ -1,0 +1,59 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name" :disabled="!isExplanationsShown">
+        <template #title> {{ currentStateAndTitle.title }} </template>
+        <TextBlock v-if="isExplanationsShown">
+            You cannot execute an auction while it is not yet profitable. Once the auction price drops below the price
+            on UniSwap, you may continue with the bidding process
+        </TextBlock>
+    </BasePanel>
+</template>
+
+<script lang="ts">
+import Vue from 'vue';
+import BigNumber from 'bignumber.js';
+import BasePanel from '../common/BasePanel.vue';
+import TextBlock from '../common/TextBlock.vue';
+
+export default Vue.extend({
+    name: 'GrossProfitCheckPanel',
+    components: { TextBlock, BasePanel },
+    props: {
+        grossProfit: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: undefined,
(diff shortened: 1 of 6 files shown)


frontend/components/panels/GrossProfitCheckPanel.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name" :disabled="!isExplanationsShown">
+        <template #title> {{ currentStateAndTitle.title }} </template>
+        <TextBlock v-if="isExplanationsShown">
+            You cannot execute an auction while it is not yet profitable. Once the auction price drops below the price
+            on UniSwap, you may continue with the bidding process
```
LukSteib, line 6: NIT: instead of "with the bidding process" I would write "with auction participation". Rationale: Since the grossprofitcheck is only used for the swap transaction flow a user doesn't bid.

frontend/components/panels/NetProfitCheckPanel.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name" :disabled="!isExplanationsShown">
+        <template #title> {{ currentStateAndTitle.title }} </template>
+        <TextBlock v-if="isExplanationsShown">
+            Executing an auction with a negative net profit is possible, but according to our calculations it will
+            result in the net loss due to the transaction fees being higher then the gross profit.
```
LukSteib, line 6: - [ ] let's write "a net loss" instead of "the net loss"
- [ ] "higher than" instead of "higher then"

LukSteib (changes requested):
> I also added a disabled prop to the BasePanel. This is useful for when you have a text only panel (with no call to action), which would result in just an empty expand in the expert mode.

Thinking about that I would suggest instead of having the `disabled` prop to show the texts in expert mode as well. IMO even as an expert this information is useful as it explains "why" and not "what" as the texts of the other panels. Up for discussion, would like to hear @valiafetisov perspective on that.

frontend/components/panels/GrossProfitCheckPanel.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name" :disabled="!isExplanationsShown">
```
valiafetisov, line 2: I though the proposal from lukas was to still show text even in expert mode? I also find it sensible, so let's get rid of the `disabled` in those instances

valiafetisov (approved):

LukSteib (approved):

PR #194, opened 2022-04-06
Fix gas price

### LINKED ISSUE #189: Improve gas fetching
### Goal

Better strategy for gas fetching

### Context

Currently we use a regular https endpoint to fetch fast gas price data, which is not optimal: it sometimes overloaded and down and it doesn't work with forking. Instead, we can implement support for the chainlink fast gas oracle https://data.chain.link/ethereum/mainnet/gas/fast-gas-gwei. To get current gas data we just need to fetch `latestAnswer` on [0x169e633a2d1e6c10dd91238ba11c4a708dfef37c](https://etherscan.io/address/0x169e633a2d1e6c10dd91238ba11c4a708dfef37c#readContract) or even better using `fast-gas-gwei.data.eth` handle. Potentially we can also use it with hardhat forking (instead of currently hardcoded values) to see actual historical gas data of the fork.

### Tasks
- [x] Replace gas data fetching with new logic
- [x] Optimise for [EIP-1559](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md)
- [x] Remove hardcoded value for the `localhost` network
- [x] Test

valiafetisov:
Closes #189 
Closes https://github.com/sidestream-tech/auction-ui/issues/401

Not sure if it actually fixes the problem, but this way we compute `gasLimit` ourselves with higher gas prices.

Since the gas prices are applied differently on different networks, testing is only possible if there is an auction on the mainnet or on the hardhat fork.

### DIFF
--- a/core/src/gas.ts
+++ b/core/src/gas.ts
@@ -9,6 +9,18 @@ import getProvider from './provider';
 
 const CHAINLINK_ORACLE_ADDRESS = '0x169e633a2d1e6c10dd91238ba11c4a708dfef37c';
 const GAS_CACHE_MS = 10 * 1000;
+const maxPriorityFeePerGas = new BigNumber(process.env.MAX_PRIORITY_FEE_PER_GAS_WEI || '1500000000').shiftedBy(
+    -ETH_NUMBER_OF_DIGITS
+);
+
+export const getBaseFeePerGas = async function (network: string): Promise<BigNumber> {
+    const provider = await getProvider(network);
+    const pendingBlock = await provider.getBlock('latest');
+    if (!pendingBlock?.baseFeePerGas?._hex) {
+        return new BigNumber(NaN);
+    }
+    return new BigNumber(pendingBlock?.baseFeePerGas?._hex).shiftedBy(-ETH_NUMBER_OF_DIGITS);
+};
 
 const getOracleGasPrice = async function (network: string): Promise<BigNumber> {
     const provider = await getProvider(network);
@@ -20,8 +32,8 @@ const getOracleGasPrice = async function (network: string): Promise<BigNumber> {
 const _getEIP1559Values = async function (network: string): Promise<GasParameters> {
     const oracleGasPrice = await getOracleGasPrice(network);
     return {
-        maxFeePerGas: oracleGasPrice.shiftedBy(ETH_NUMBER_OF_DIGITS).toFixed(0),
-        maxPriorityFeePerGas: '1500000000',
+        maxPriorityFeePerGas: maxPriorityFeePerGas.shiftedBy(ETH_NUMBER_OF_DIGITS).toFixed(0),
+        maxFeePerGas: oracleGasPrice.plus(maxPriorityFeePerGas).shiftedBy(ETH_NUMBER_OF_DIGITS).toFixed(0),
     };
 };
 
--- a/core/src/gas.ts
+++ b/core/src/gas.ts
@@ -11,7 +11,7 @@ const GAS_CACHE = 10 * 1000;
 
 const _getCurrentGasPrice = async function (): Promise<BigNumber> {
     const gasData = await fetch(API_URL).then(r => r.json());
-    const gasPriceString = ethers.utils.formatUnits(gasData[TRANSACTION_SPEED] / 10, 'gwei');
+    const gasPriceString = ethers.utils.formatUnits(gasData[TRANSACTION_SPEED], 'gwei');
     return new BigNumber(gasPriceString);
 };
 
--- a/core/src/gas.ts
+++ b/core/src/gas.ts
@@ -1,23 +1,25 @@
 import { ethers } from 'ethers';
+import memoizee from 'memoizee';
 import BigNumber from './bignumber';
 import { getNetworkConfigByType } from './constants/NETWORKS';
 import { ETH_NUMBER_OF_DIGITS } from './constants/UNITS';
-import memoizee from 'memoizee';
+import CHAINLINK_FAST_GAS_ABI from './abis/CHAINLINK_FAST_GAS.json';
+import getProvider from './provider';
 
-const API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
-const TRANSACTION_SPEED = 'fast';
+const CHAINLINK_FAST_GAS_ADDRESS = '0x169E633A2D1E6c10dD91238Ba11c4A708dfEF37C';
+const GAS_CACHE_MS = 10 * 1000;
 
-const GAS_CACHE = 10 * 1000;
-
-const _getCurrentGasPrice = async function (): Promise<BigNumber> {
-    const gasData = await fetch(API_URL).then(r => r.json());
-    const gasPriceString = ethers.utils.formatUnits(gasData[TRANSACTION_SPEED] / 10, 'gwei');
-    return new BigNumber(gasPriceString);
+const _getCurrentGasPrice = async function (network: string): Promise<BigNumber> {
+    const provider = await getProvider(network);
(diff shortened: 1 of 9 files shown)


core/src/gas.ts
```diff
@@ -1,23 +1,25 @@
...
 import { ETH_NUMBER_OF_DIGITS } from './constants/UNITS';
-import memoizee from 'memoizee';
-
-const API_URL = 'https://ethgasstation.info/json/ethgasAPI.json';
-const TRANSACTION_SPEED = 'fast';
+import CHAINLINK_FAST_GAS_ABI from './abis/CHAINLINK_FAST_GAS.json';
+import getProvider from './provider';
 
-const GAS_CACHE = 10 * 1000;
+const CHAINLINK_FAST_GAS_ADDRESS = '0x169E633A2D1E6c10dD91238Ba11c4A708dfEF37C';
```
benediktkaiser, line 9: Will this address never change or will we need to manually recode this every once in a while?

benediktkaiser (changes requested):

core/src/gas.ts
valiafetisov: Just like with other addresses, they can't change. The only possibility is that they would be discontinued, it's an interesting question how can we prepare for this situation, but kind of unrelated to the discussion (:

benediktkaiser (approved):

LukSteib:
Was able to execute transactions on hardhat fork.

What makes me wonder is the very high `Max base fee` that we seem to set and that no `Priority Fee` is set. Based on the information provided by Metamask:

What we are suggesting:
image

What the `Aggressive` settings by metamask would be:
image



---

Still trying to understand what we are now doing. Summarising my understanding below with questions included (please correct if it is completely off):
* fetch the current `gasPrice` from chainlink contract (method: latestAnswer)
  * _question: What is this Fast-Gas-Price indicator we are fetching here actually refer to?_ 
* estimate `gasLimit` ourselves
  * setting `maxPriorityFeePerGas` to `gasPrice`
  * setting `maxFeePerGas` to `gasPrice` x 2
  * _question: Are we using a contract for this estimation or how do we actually end up with the `gasLimit` estimate?_
* Send `gasLimit` and `maxFeePerGas` for tx execution
  * _question: If we send the `maxFeePerGas` only my expectation was that the `maxPriorityFeePerGas` is set automatically - which doesn't seem to be the case when looking at the screenshots above._

valiafetisov:
> Based on the information provided by Metamask

1) I am not sure why metamask shows suggested value as `Max base fee` in the new "experimental" interface that you use, while it's actually `Max fee` (`maxFeePerGas`), in my old one it is named property
2) I don't think that metamask suggests correct numbers, since the transaction fails when I try to execute using suggested `high` values.

<details><summary>Our value of 356,9 (marked as "higher than necessary")</summary>
Screenshot 2022-04-11 at 10 56 43
</details>

<details><summary>Suggest high value of 25,714034617</summary>
Screenshot 2022-04-11 at 11 02 19
</details>

<details><summary>Error, if executed with suggested value</summary>
Screenshot 2022-04-11 at 11 02 37
</details>

Can you please confirm that value suggested on localhost actually works for you?

Although I'm not sure why metamask throws an error here. It seems like they recently changed the logic and no longer allow submitting transactions that will not be "executable" in the current block. That's why we encountered https://github.com/sidestream-tech/auction-ui/issues/401 only recently.

> Still trying to understand what we are now doing. Summarising my understanding below with questions included (please correct if it is completely off):
> 
> * fetch the current `gasPrice` from chainlink contract (method: latestAnswer)
>   
>   * _question: What is this Fast-Gas-Price indicator we are fetching here actually refer to?_

Just like before, it's the amount of gas that is required to execute transaction quickly. This number didn't change it significance, EIP-1559 only changed that we can now provide gas value as _relative_ to the current block, potentially saving some money in case `base gas fee` for the block it's mined is lower than `base gas fee` for the block it's submitted for, effectively limiting how much money the miner can grab.

> * estimate `gasLimit` ourselves
>   
>   * setting `maxPriorityFeePerGas` to `gasPrice`
>   * setting `maxFeePerGas` to `gasPrice` x 2
>   * _question: Are we using a contract for this estimation or how do we actually end up with the `gasLimit` estimate?_

So the `gasLimit` estimation is the process when actual transaction is executed in a virtual machine instead of being actually submitted for mining. Therefore it can only be estimated if _all transaction parameters_ are correct _for the current block_. If the transaction can not be executed for any reason (eg not enough ETH in the wallet, gas price is too low, authorization haven't been executed) than gas limit estimation will fail and therefore we wouldn't be able to submit the transaction. So every `can not estimate gas limit` error actually means `transaction parameters are not correct for the current block`.

Setting `maxFeePerGas` to be `gasPrice` x 2 will leverage the problem with the gas prices, but actually is still problematic if the user doesn't have x2 amount of ETH to pay for the fees. So maybe it's not the best solution.

> * Send `gasLimit` and `maxFeePerGas` for tx execution
>   
>   * _question: If we send the `maxFeePerGas` only my expectation was that the `maxPriorityFeePerGas` is set automatically - which doesn't seem to be the case when looking at the screenshots above._

Yes, if we send only `maxFeePerGas`, `maxPriorityFeePerGas` should be calculated automatically. But we may also provide that one as well, using `baseFeePerGas`, although then we need to constantly fetch `baseFeePerGas` for each block.

Based on the all above, we may 
- want to set `maxPriorityFeePerGas`
- fetch `baseFeePerGas` periodically
- remove `gasPrice` x 2 calculation and let the transaction fail
  - better handle estimation error by displaying the reason, eg "error reason, try in the next block", idk?

LukSteib:
> Can you please confirm that value suggested on localhost actually works for you?

You are right. When trying to execute the tx with "Aggressive" Metamask default settings, I receive the same error as reference by you 

> Just like before, it's the amount of gas that is required to execute transaction quickly.

Sorry for being persistent on that one, but try to understand: Is the `gasPrice` that we know fetch from chainlink now equivalent to the `base gas fee` in the EIP-1559 concept? Or how does it fit into this distinction between `base gas fee` and `priority fee`? 

> Setting maxFeePerGas to be gasPrice x 2 will leverage the problem with the gas prices, but actually is still problematic if the user doesn't have x2 amount of ETH to pay for the fees. So maybe it's not the best solution.

Is there a way to pass different values? Meaning gasPrice x 2 for the virtual machine based estimation and then the actual gasPrice for the actual tx? 

> want to set maxPriorityFeePerGas

Sounds good to me

> fetch baseFeePerGas periodically

Will that again blow up our quota? 

> remove gasPrice x 2 calculation and let the transaction fail

So this would mean that the user will just not be able to execute the transaction in the current block, right? IMO we should try to prevent it. And if setting gasPrice x 2 is the only approach to prevent this type of tx failure due to failing estimation, I would keep it

valiafetisov:
Should there be a way to test / see those gas prices in the dashboard? I can imagine adding something like gas price value to the dashboard and a auth/deauth buttons to all collaterals?

LukSteib:
Trying to summarise oral discussion outcomes:
* we will use `chainlink fas gas price` and set it as `maxFeePerGas`
* for the `maxPriorityFeePerGas` we have two options
  * hardcode it to a value like 1.5 GWEI (as done by ether.js)
  * fetch `base gas fee`, compare `chainlink fast gas price` with `base gas fee`, decide based on comparison a suitable value for `maxPriorityFeePerGas` 
* we are going to display the `maxFeePerGas` value that we are contiously fetching on the dashboard page
---
* we allow a user to adjust `maxFeePerGas` and `maxPriorityFeePerGas` as env variables for the keeper
  * default values will be those that we are setting

valiafetisov:
I also suggest to open a new issue to display gas-related data in the dashboard: `GasParameters` (`gasPrice` or `maxPriorityFeePerGas` and `maxFeePerGas` depending on the network), `baseFeePerGas`, `approximateTransactionFees`

PR #195, opened 2022-04-07
Updated `SwapTransaction` to use panels

### LINKED ISSUE #155: Update `SwapTransaction` to use panels
### Goal
Updated SwapTransaction flow that uses new `BasePanel` component

### Context
Since the update made in https://github.com/sidestream-tech/unified-auctions-ui/issues/145, we came up with a more friendly UI for the SwapTransaction. The logic is very similar to the `Bid with DAI` flow and most (if not all) panels should be reusable from the `Bid with DAI` flow.

### Assets
MakerDAO@2x (52)

### Tasks
- [ ] Update SwapTransaction to use `BasePanel` component

benediktkaiser:
closes #155 

Screenshot 2022-04-07 at 13 37 29


frontend/components/panels/NetProfitCheckPanel.vue
```diff
@@ -22,9 +22,19 @@ export default Vue.extend({
             type: Object as Vue.PropType<BigNumber>,
             default: undefined,
         },
+        negativeGrossProfit: {
+            type: Boolean,
+            default: false,
+        },
```
valiafetisov, line 28: I would suggest to either: not provide it here at all or to actually combine net and gross profits into a single `ProfitCheckPanel` (and therefore provide both values)

frontend/components/transaction/SwapTransaction.vue
```diff
@@ -29,27 +29,39 @@
...
-            :disabled="!auctionTransaction.isActive"
-            :is-loading="isConnecting"
+        <WalletConnectionCheckPanel
             :wallet-address="walletAddress"
+            :is-loading="isConnecting"
             :is-explanations-shown="isExplanationsShown"
+            :disabled="!auctionTransaction.isActive"
             @connectWallet="$emit('connect')"
             @disconnectWallet="$emit('disconnect')"
         />
```
valiafetisov, line 39: The idea of the panels is that they also incapsulate reusable `isCorrect` checks. 
See wallet popup for the example how it's used:
https://github.com/sidestream-tech/unified-auctions-ui/blob/a4fbe882660e45f65ba46b1244874abfbceb12a6/frontend/components/wallet/WalletDepositWithdrawBlock.vue#L27

Therefore, please use `isCorrect.sync` props on each panel

frontend/components/transaction/SwapTransaction.vue
```diff
@@ -29,27 +29,39 @@
...
+            :is-loading="isAuthorizing"
+            :disabled="!auctionTransaction.isActive || !isWalletAuthorized"
+            :is-explanations-shown="isExplanationsShown"
             @authorizeCollateral="$emit('authorizeCollateral', auctionTransaction.collateralType)"
         />
+        <GrossProfitCheckPanel :gross-profit="auctionTransaction.transactionGrossProfit" />
+        <NetProfitCheckPanel
+            :net-profit="auctionTransaction.transactionNetProfit"
+            :negative-gross-profit="!auctionTransaction.transactionGrossProfit.isGreaterThanOrEqualTo(0)"
+        />
```
valiafetisov, line 62: To make corners of the panels rounded, you need to keep the panels in a separate html element (ie a div), then the css will be applied properly.(I know it's not obvious)

https://github.com/sidestream-tech/unified-auctions-ui/blob/a4fbe882660e45f65ba46b1244874abfbceb12a6/frontend/components/common/BasePanel.vue#L95-L103

frontend/components/transaction/SwapTransaction.vue
```diff
@@ -29,27 +29,39 @@
...
         <ExecutionBlock
```
valiafetisov, line 63: Please pass separate `isCorrect` values into the `disabled` prop of the `ExecutionBlock` (you also would need to clean up `ExecutionBlock` to not accept any extra props it doesn't need, eg `transaction-gross-profit`)

frontend/components/panels/NetProfitCheckPanel.vue
benediktkaiser: I would avoid this as it would result in me needing to wrap both panels in a div. This would then mess up the rounding of the panels:

Screenshot 2022-04-08 at 11 47 19

frontend/components/transaction/SwapTransaction.vue
benediktkaiser: I removed `transaction-gross-profit`, but `wallet-address` is needed in enable the send profit to another wallet feature.

frontend/components/panels/NetProfitCheckPanel.vue
valiafetisov: No, I didn't mean to wrap two panels into one component, but to merge them into one panel which is: 0) `inactive` if there is no wallet 1) `incorrect` when net profit is negative 2) `notice` when gross profit is negative 3) `correct` when gross profit is positive. Would that make sense?

LukSteib (approved):
I like the way to handle both gross and net profit in one panel
(NIT: The story "Positive Gross Profit, Negative Net Profit" is impossible to happen :) )

frontend/components/transaction/SwapTransaction.vue
```diff
@@ -29,32 +29,49 @@
...
+                :is-collateral-authorized="isCollateralAuthorised"
+                :auth-transaction-fee-e-t-h="auctionTransaction.authTransactionFeeETH"
+                :wallet-address="walletAddress"
+                :is-loading="isAuthorizing"
+                :disabled="!isAuctionActiveAndNotFinished || !isWalletDAIAuthorizationCheckPassed"
+                :is-explanations-shown="isExplanationsShown"
+                @authorizeCollateral="$emit('authorizeCollateral', auctionTransaction.collateralType)"
+            />
+            <ProfitCheckPanel
+                :is-correct.sync="isProfitCheckPassed"
```
valiafetisov, line 63: Currently, `isProfitCheckPassed` will be false in case of `The transaction net profit is negative`, but should be true, since (according to the text of the notice) the transaction should be executable

(you have to change `currentStateAndTitle` watcher, but github doesn't let me comment there)

frontend/components/panels/ProfitCheckPanel.vue
```diff
@@ -1,10 +1,14 @@
...
         <template #title> {{ currentStateAndTitle.title }} </template>
-        <TextBlock>
+        <TextBlock v-if="currentStateAndTitle.title === 'The transaction gross profit is negative'">
             You cannot execute an auction while it is not yet profitable. Once the auction price drops below the price
             on UniSwap, you may continue with the auction participation
         </TextBlock>
+        <TextBlock v-else>
+            Executing an auction with a negative net profit is possible, but according to our calculations it will
+            result in a net loss due to the transaction fees being higher than the gross profit.
+        </TextBlock>
```
valiafetisov, line 11: Currently, this text will be displayed when `The transaction profit is unknown` and `The transaction profit is positive`. I don't think it's desired. I think we should aim for the
- Generic text what explains what are both things are in one text
- "Negative net profit" text that is shown only when `notice`

frontend/components/panels/ProfitCheckPanel.vue
```diff
@@ -15,20 +19,30 @@ import BasePanel from '../common/BasePanel.vue';
...
         currentStateAndTitle(): PanelProps {
-            if (this.grossProfit === undefined) {
+            if (this.grossProfit === undefined || this.netProfit === undefined) {
                 return {
-                    name: 'notice',
-                    title: `The transaction gross profit is unknown`,
+                    name: 'incorrect',
+                    title: `The transaction profit is unknown`,
+                };
+            }
```
valiafetisov, line 41: Since we have two different profits, let's always specify which one are we talking about. Here and everywhere else:
```suggestion
            if (this.grossProfit === undefined || this.grossProfit?.isNaN()) {
                return {
                    name: 'incorrect',
                    title: `The transaction gross profit is unknown`,
                };
            }
```

frontend/components/panels/ProfitCheckPanel.vue
```diff
@@ -0,0 +1,84 @@
...
+    },
+    computed: {
+        currentStateAndTitle(): PanelProps {
+            if (this.grossProfit === undefined || this.grossProfit.isNaN()) {
+                return {
+                    name: 'incorrect',
+                    title: `The transaction gross profit is unknown`,
+                };
+            }
+            if (this.netProfit === undefined || this.netProfit.isNaN()) {
```
valiafetisov, line 51: Same here

valiafetisov (dismissed):

LukSteib (changes requested):
- [ ] The collateral authorisation state seems not to be reflected properly in the UI which prevents auction execution (in other words: After executing `Hope` function for different collateral the UI still requires this authorisation step) See recording below: I am authorising the collateral type -> tx gets confirmed -> but authorisation status is not reflected properly in UI


https://www.loom.com/share/7c7d2bb29fe54e288ccfaec615a7b4e1

LukSteib (changes requested):
- [ ] NIT and maybe unrelated to this PR: I have a feeling that we again set `kovan` to the default network -> can you pls have a look and ensure that default network is `mainnet`

Rest is looking good

fegbert:
If not set default is still `mainnet` 🙂 

```js
DEFAULT_ETHEREUM_NETWORK: process.env.DEFAULT_ETHEREUM_NETWORK || 'mainnet'
```

LukSteib (approved):

valiafetisov (approved):

PR #196, opened 2022-04-11
Added `WithdrawCollateralPanel`

### LINKED ISSUE #183: Implement new `WithdrawCollateralPanel`
### Goal

New panel that allows user to withdraw collateral from VAT

### Context
Since we're moving to the universal panel approach, we should make a new component that allows a user to withdraw collateral of the auction at hand from VAT to wallet. The states of the panel can be found on the existing WithdrawCollateralBlock, but please edit the texts to be be universal so the panel can be reused in many different places regardless of the context. This particular panel will be used in:
- bid transaction page
- swap transaction page

### Tasks
- [x] create the panel

fegbert:
closes #183 

**Nothing to withdraw (inactive)**
image

**Something to withdraw (notice)**
image

### DIFF
--- a/frontend/components/common/BasePanel.vue
+++ b/frontend/components/common/BasePanel.vue
@@ -43,8 +43,8 @@ const STATES = [
     },
     {
         name: 'notice',
-        titleClass: 'text-yellow-500',
-        iconClass: 'text-yellow-500',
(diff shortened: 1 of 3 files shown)


valiafetisov (approved):
Will you integrate it into to the transaction page in #197?

PR #201, opened 2022-04-12
Add `no-floating-promises` rule

### LINKED ISSUE #150: Add new eslint rule `no-floating-promises`
### Goal

New rule is added to properly handle promises and the codebase is updated to comply

### Context
During the implementation of a fix in https://github.com/sidestream-tech/unified-auctions-ui/pull/147#discussion_r829070616, we agreed to add a new eslint rule to this repo as an experiment as well as use its exceptions if needed.

The rule is [`no-floating-promises`](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md)

### Tasks
- [x] add new rule
- [x] update codebase
- [ ] if works for us, add new issue to the https://github.com/sidestream-tech/vue-nuxt-ts-scaffold

fegbert:
closes #150


frontend/.eslintrc.js
```diff
@@ -30,6 +34,7 @@ module.exports = {
         curly: 'warn',
         'no-undef': 'off',
         'eol-last': ['error', 'always'],
+        '@typescript-eslint/no-floating-promises': ['error'],
```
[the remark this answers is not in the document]
fegbert: I think so? The rule wouldn't work otherwise, right?

frontend/tsconfig.json
```diff
@@ -32,7 +32,8 @@
     },
     "include": [
         "package.json",
-        "**/*.ts"
+        "**/*.ts",
+        ".eslintrc.js"
```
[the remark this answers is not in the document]
fegbert: Moved into jsconfig 👍

fegbert:
CI took 03:14 to lint. (Around 30 seconds before)
Think that's still fine though.

frontend/.eslintrc.typescript.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+    parserOptions: {
+        project: ['tsconfig.json', 'jsconfig.json'],
+        tsconfigRootDir: __dirname,
+    },
+    rules: {
+        '@typescript-eslint/no-floating-promises': ['error'],
+    },
+};
```
valiafetisov, line 9: Please import and spread the original configuration here:

```suggestion
import eslintrc from `./.eslintrc.js`;

module.exports = {
    ...eslintrc,
    parserOptions: {
        project: ['tsconfig.json', 'jsconfig.json'],
        tsconfigRootDir: __dirname,
    },
    rules: {
        ...eslintrc.rules,
        '@typescript-eslint/no-floating-promises': ['error'],
    },
};
```

frontend/jsconfig.json
```diff
@@ -8,5 +8,11 @@
       "@@/*": ["./*"]
     }
   },
-  "exclude": ["node_modules", ".nuxt", "dist"]
+  "exclude": ["node_modules", ".nuxt", "dist"],
+  "include": [
+      "**/*.js",
+      "**/*.vue",
+      ".eslintrc.js",
+      ".eslintrc.typescript.js"
```
valiafetisov, line 16: Why do you need to include it into jsconfig?

frontend/lib/wallets/MetaMask.ts
```diff
@@ -76,12 +76,15 @@ export default class MetaMask extends AbstractWallet {
...
+        await window.$nuxt.$store.dispatch('network/setWalletChainId', window.ethereum.chainId);
     }
 
     public accountsChangedHandler(addresses: Array<string>) {
         this.addresses = addresses;
-        window.$nuxt.$store.dispatch('wallet/setAddress', this.address);
+        window.$nuxt.$store.dispatch('wallet/setAddress', this.address).then(
+            () => {},
+            () => {}
+        );
```
valiafetisov, line 87:
```suggestion
        window.$nuxt.$store.dispatch('wallet/setAddress', this.address).catch(() => {});
```

Since the rule only requires us to catch things, I believe that will be a little shorter modification

frontend/containers/AuctionsFakeData.vue
```diff
@@ -147,7 +147,10 @@ export default Vue.extend({
         restart() {
             if (this.selectedAuction) {
                 this.selectedAuction.isActive = true;
-                message.success('The auction has been restarted!');
+                message.success('The auction has been restarted!').then(
+                    () => {},
+                    () => {}
+                );
             }
         },
```
valiafetisov, line 155: You can mark `restart` as async to avoid that here

valiafetisov (commented):
Let's also add the same rule to the `/bot`!

frontend/.eslintrc.typescript.js
fegbert: Right now `eslint` uses the default `.eslintrc.js` and this file as an "extra" config, so this would just duplicate the config. I could do this and add the `no-eslintrc` option to the command which lets eslint ignore the default file.

frontend/jsconfig.json
fegbert: Without it this error gets thrown:

```bash
The file does not match your project config: .eslintrc.js.
The file must be included in at least one of the projects provided. eslint
````

So the files get found by eslint but it doesn't know what to do with them. There's then 2 options to fix this:

1. Ignore files - add to .eslintignore
2. Manually tell to lint them - what I did

I thought manually adding & linting them is probably better than just ignoring them, right?

File: frontend/.eslintrc.typescript.js (hunk 3 above)
Comment (valiafetisov) [reply frontend/.eslintrc.typescript.js:9]: > and this file as an "extra" config

Interesting, but how does this "extra" logic handles new rules? Are they merged together with the default ones or do they overwrite them? I expect that they will overwrite, that's why I think it's required to merge them manually, otherwise default rules wouldn't be checked in the CI. I would anyway would make it explicit by accepting the suggestion (or what's the downside of applying them "twice"?)

File: frontend/jsconfig.json (hunk 4 above)
Comment (valiafetisov) [reply frontend/jsconfig.json:16]: > I thought manually adding & linting them is probably better than just ignoring them, right?

Correct! Thanks for the explanation

File: frontend/store/auctions.ts [hunk 7]
```diff
@@ -242,12 +242,12 @@ export const actions = {
...
         const auction = getters.getAuctionById(id);
         if (!auction) {
-            message.error(`Auction reset error: can not find auction with id "${id}"`);
+            await message.error(`Auction reset error: can not find auction with id "${id}"`);
             return;
         }
         const walletAddress = getWallet().address;
         if (!walletAddress) {
-            message.error('Bidding error: can not find wallet');
+            await message.error('Bidding error: can not find wallet');
```
valiafetisov, line 250: Btw, what does awaiting a message implies? Does the promise gets resolved as soon as the message is shown or as soon as it's finished showing (so the promise will be resolved in 3 seconds)?

valiafetisov:
- [ ] Please [don't forget to add the same config to the bot](https://github.com/sidestream-tech/unified-auctions-ui/pull/201#pullrequestreview-945905631). Although you can wait to the other currently open `bot`-related PRs to be merged until working on that

frontend/.eslintrc.typescript.js
fegbert: They do get merged together actually, but I agree that there's probably no downside to simply applying them twice, and for readability it makes sense to spread the original config here so that it's easier to understand that this file is just the normal config + some extra stuff.

fegbert:
> Although you can wait to the other currently open bot-related PRs to be merged until working on that

Will do that 👍

frontend/store/auctions.ts
fegbert: Wow yes, it actually is a promise that resolves after the message is finished. Guess this is used for sequenced messages? But that shouldn't be awaited, right?

valiafetisov: > But that shouldn't be awaited, right?

Yeah, then I don't think we need to await them. Let's also use `.catch(() => {})` or `// eslint-disable-line no-floating-promises`

valiafetisov:
- https://github.com/sidestream-tech/unified-auctions-ui/pull/225
- https://github.com/sidestream-tech/unified-auctions-ui/pull/208

Are both merged, you can proceed with modifying `bot`. 
Also, since `core` have their own eslint, we need to add it there as well 🙂

frontend/containers/AuctionsFakeData.vue
```diff
@@ -147,6 +147,7 @@ export default Vue.extend({
         restart() {
             if (this.selectedAuction) {
                 this.selectedAuction.isActive = true;
+                // eslint-disable-next-line @typescript-eslint/no-floating-promises
```
benediktkaiser, line 150: I read your conversation history about these lines. I have one more idea:

We write our own custom wrapper for the AntDesign messenger, that we can then call in all of the other places we currently use the message component. 

We can then disable the typescript check only in the helper, which removes the need to add these comments everywhere in the rest of our codebase. What do you think of this idea @valiafetisov?

valiafetisov: Yeah, sounds a little cleaner!

benediktkaiser (approved):

frontend/helpers/messageWrapper.ts
```diff
@@ -0,0 +1,52 @@
...
+    warn(content: ConfigType, duration?: ConfigDuration, onClose?: ConfigOnClose): void;
+    info(content: ConfigType, duration?: ConfigDuration, onClose?: ConfigOnClose): void;
+    error(content: ConfigType, duration?: ConfigDuration, onClose?: ConfigOnClose): void;
+    loading(content: ConfigType, duration?: ConfigDuration, onClose?: ConfigOnClose): void;
+    open: (config: MessageOptions) => void;
+    config: (options: MessageConfigOptions) => void;
+    destroy: () => void;
+}
+
+export const message: WrapperMessage = {
```
valiafetisov, line 24: I no longer sure that was actually a good idea, probably adding `.catch(() => {})` to every `message` was enough. But let's keep it as it is, since it's already there

fegbert: Message actually doesn't have `catch`, only `then`. Not sure why

frontend/helpers/messageWrapper.ts
```diff
@@ -0,0 +1,52 @@
...
+import type {
+    MessageOptions,
+    MessageConfigOptions,
+    ConfigType,
+    ConfigDuration,
+    ConfigOnClose,
+} from 'ant-design-vue/types/message';
+import { message as antdMessage } from 'ant-design-vue';
+
+declare interface WrapperMessage {
```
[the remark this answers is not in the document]
fegbert: I wasn't sure how to do it with the solution you provided, but I found [this answer](https://stackoverflow.com/a/55032655) in another thread which lets me make a modified version of the `Message` type

Is this what you imagined?

frontend/helpers/messageWrapper.ts
valiafetisov: > `MessageType` actually doesn't have `catch`, only `then`

Oh wow, sounds very strange 😱

fegbert:
> MessageType actually doesn't have catch, only then.

Nevermind! After looking at the `MessageType` Type a little more, I found out it actually does expose the promise itself, so I can use
```ts
message.success("blabla").promise.catch(() => {})
```

And remove the whole wrapper 🙂

valiafetisov:
Closing as outdated

PR #203, opened 2022-04-13
Added Gas Data to Dashboard

### LINKED ISSUE #198: Add gas-related data to the dashboard
### Goal
Dashboard contains another small table with gas-related data

### Context
During the discussion in https://github.com/sidestream-tech/unified-auctions-ui/issues/194#issuecomment-1096318570, we realised that it's currently impossible to check that gas prices work correctly in the mainnet (or any other network) if there is no auctions. Therefore, we want to display this data in the dashboard.

The table should contain at least:
- 3 rows with `GasParameters` (`gasPrice` or `maxPriorityFeePerGas` and `maxFeePerGas` depending on the network some are expected to be undefined)
- 1 row with `baseFeePerGas`
- 3 rows with `approximateTransactionFees`

| Gas Parameter | Value in ETH | Value in DAI |
|---|---|---|
| `gasPrice` | 0.0000134 | Unknown |
| `maxPriorityFeePerGas` |  | Unknown |
| `maxFeePerGas` |  | Unknown |
| `baseFeePerGas` |  | Unknown |
| `biddingTransactionFee` |  |  |
| `authTransactionFee` |  |  |
| `restartTransactionFee` |  |  |

Please add enough decimal points to the value formatting, so it's actually visible

### Tasks
- [x] Create store action to fetch all required gas data
- [x] Add new table to the dashboard
- [ ] Test that all gas prices are correct

benediktkaiser:
closes #198 

Screenshot 2022-04-13 at 13 18 22

In this issue it was stated that the decimals should be as exact as possible. To ensure this I used our `FormatCurrency` component, which had a built in formatting we agreed on for the rest of the UI. Is this formatting fine, or should I always show the exact number, no matter how long it is?

### DIFF
--- a/frontend/components/DashboardAuctionsView.vue
+++ b/frontend/components/DashboardAuctionsView.vue
@@ -27,19 +27,32 @@
             </TextBlock>
             <CalleeTable :callees="callees" />
         </div>
+        <div class="Block space-y-4 md:space-y-8 max-w-screen-sm">
+            <TextBlock v-if="isExplanationsShown" title="Gas prices">
+                These are the current gas prices on your selected network.
+            </TextBlock>
+            <GasTable
+                :base-fee-per-gas="baseFeePerGas"
+                :gas-parameters="gasParameters"
+                :transaction-fees="transactionFees"
+            />
+        </div>
     </div>
 </template>
 
 <script lang="ts">
-import type { CollateralRow, CalleeAddresses } from 'auctions-core/src/types';
+import type { CollateralRow, CalleeAddresses, GasParameters, TransactionFees } from 'auctions-core/src/types';
 import Vue from 'vue';
+import BigNumber from 'bignumber.js';
 import CollateralTable from './CollateralTable.vue';
 import CalleeTable from './CalleeTable.vue';
+import GasTable from './GasTable.vue';
 import LandingBlock from '~/components/layout/LandingBlock.vue';
 import TextBlock from '~/components/common/TextBlock.vue';
 
 export default Vue.extend({
     components: {
+        GasTable,
         LandingBlock,
         TextBlock,
         CollateralTable,
@@ -57,6 +70,18 @@ export default Vue.extend({
         callees: {
             type: Object as Vue.PropType<CalleeAddresses>,
             default: () => ({}),
+        },
+        baseFeePerGas: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: undefined,
+        },
+        gasParameters: {
+            type: Object as Vue.PropType<GasParameters>,
+            default: undefined,
+        },
+        transactionFees: {
+            type: Object as Vue.PropType<TransactionFees>,
+            default: undefined,
         },
     },
     computed: {
--- a/frontend/components/DashboardAuctionsView.vue
+++ b/frontend/components/DashboardAuctionsView.vue
@@ -18,7 +18,7 @@
             </TextBlock>
         </div>
         <div class="Block max-w-screen-lg">
-            <CollateralTable :collaterals="collaterals" class="overflow-x-scroll" />
+            <CollateralTable :collaterals="collaterals" class="overflow-x-auto" />
         </div>
         <div class="Block space-y-4 md:space-y-8 max-w-screen-sm">
             <TextBlock v-if="isExplanationsShown" title="Exchange Callee Contracts">
--- a/frontend/components/GasTable.vue
+++ b/frontend/components/GasTable.vue
@@ -12,11 +12,7 @@
                         <span> Gas Price </span>
                     </td>
                     <td class="Body">
-                        <FormatCurrency
-                            v-if="gasParameters && gasParameters.gasPrice"
-                            :value="parseInt(gasParameters.gasPrice)"
-                            currency="Gwei"
-                        />
+                        <span v-if="gasParameters && gasParameters.gasPrice"> {{ gasParameters.gasPrice }} Gwei </span>
                         <span v-else class="UnknownValue"> Unknown </span>
                     </td>
                     <td class="Body">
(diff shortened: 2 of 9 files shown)


frontend/components/DashboardAuctionsView.vue
```diff
@@ -18,7 +18,7 @@
             </TextBlock>
         </div>
         <div class="Block max-w-screen-lg">
-            <CollateralTable :collaterals="collaterals" class="overflow-x-scroll" />
+            <CollateralTable :collaterals="collaterals" class="overflow-x-auto" />
```
benediktkaiser, line 21: I also fixed a small issue here with the scrollbar. Before the scrollbar was always shown, even if it was not needed. With this new class it only shows once it is actually needed.

benediktkaiser:
> `getGasParameters` is in WEI

That makes a lot of sense. I was a bit sceptical myself haha.

valiafetisov:
- [ ] The third column contains value in DAI, as specified in the title, so no need to add more than 2 decimals
- [ ] First 4 rows will contain very small values and using mathmatical notation is not acceptable as suggested, please use enough decimal points or format it to GWEI by shifting it according to the table https://user-images.githubusercontent.com/9872658/34238788-2f7fa014-e636-11e7-8f62-cae47040f247.png by 9 decimal points

benediktkaiser:
Does this look correct?
Screenshot 2022-04-13 at 16 24 36

valiafetisov:
Can I also ask you to add restart transaction fee in DAI? I just checked that it's not in the types, but you can quickly add it to the `getApproximateTransactionFees`.

benediktkaiser:
Added `RestartFeeDAI` and corrected all the formatting! 

Screenshot 2022-04-13 at 16 48 02

valiafetisov:
`maxPriorityFeePerGas` suppose to be 1.5 Gwei according to this line: https://github.com/sidestream-tech/unified-auctions-ui/blob/493dcd57d7e7b43b9d0a47183e99e0be27230a75/core/src/gas.ts#L12 Why is it displayed as 1?

benediktkaiser:
> `maxPriorityFeePerGas` suppose to be 1.5 Gwei according to this line:
> 
> https://github.com/sidestream-tech/unified-auctions-ui/blob/493dcd57d7e7b43b9d0a47183e99e0be27230a75/core/src/gas.ts#L12
> 
> Why is it displayed as 1 now?

It was because I was using `pharseInt()`. I removed this and it all looks good now!

Screenshot 2022-04-13 at 17 05 20

frontend/components/GasTable.vue
```diff
@@ -44,11 +38,9 @@
                         <span> Max Fee Per Gas </span>
                     </td>
                     <td class="Body">
-                        <FormatCurrency
-                            v-if="gasParameters && gasParameters.maxFeePerGas"
-                            :value="parseInt(gasParameters.maxFeePerGas)"
```
valiafetisov: Please convert it to BigNumber, not number and let's still use `FormatCurrency` since it's animated (:

valiafetisov (commented):
Do we also periodically refetch those values? Once in 30 seconds will be enough here.

frontend/store/gas.ts
```diff
@@ -0,0 +1,102 @@
...
+        }
+        const initialGasParameters = await getGasParametersForTransaction(network);
+
+        const gasParameters: GasParameters = {
+            maxFeePerGas:
+                initialGasParameters.maxFeePerGas &&
+                new BigNumber(initialGasParameters.maxFeePerGas)
+                    .shiftedBy(-ETH_NUMBER_OF_DIGITS)
+                    .shiftedBy(GWEI_NUMBER_OF_DIGITS)
+                    .toString(),
```
valiafetisov, line 73: As discussed, please never use `.toString()` on big numbers. Also, my suggestion would be of course to format numbers in the view, not in the store

benediktkaiser: I moved the formatting to the component, bypassing all this `toString` back and forth.

benediktkaiser:
> Do we also periodically refetch those values? Once in 30 seconds will be enough here.

I missed this. I can add it right now.

valiafetisov (approved):

frontend/components/GasTable.vue
```diff
@@ -150,6 +159,11 @@ export default Vue.extend({
             default: undefined,
         },
     },
+    methods: {
+        formatGasParameters(value): BigNumber {
+            return new BigNumber(value).shiftedBy(-ETH_NUMBER_OF_DIGITS).shiftedBy(GWEI_NUMBER_OF_DIGITS);
```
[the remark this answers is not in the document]
benediktkaiser: I committed the change!

valiafetisov (approved):

LukSteib (changes requested):
- [ ] The `Gas Price` in row one is continously UNKNOWN. Do we need it in the first place? AFAIK we use it for `Max Fee Per Gas`, so IMO we can also get rid of the `Gas Price` here
- [ ] We might rename the header row of the second column from `Value in ETH` to `Value` only since we display a mix of ETH and GWEI

benediktkaiser:
> * [ ]  The `Gas Price` in row one is continously UNKNOWN. Do we need it in the first place? AFAIK we use it for `Max Fee Per Gas`, so IMO we can also get rid of the `Gas Price` here

> gasPrice or maxPriorityFeePerGas and maxFeePerGas depending on the network some are expected to be undefined

I think @valiafetisov mentioned that not all of the three values will be found on every network.

valiafetisov:
One other neat peak: can we please format all units the same way? I suppose, uppercase?
> Screenshot 2022-04-14 at 09 29 05

LukSteib (approved):

valiafetisov:
Approved, btw

PR #207, opened 2022-04-13
Added whitelisting feature to keeper

### LINKED ISSUE #206: Whitelisting collaterals
### Goal

Keeper user is able to specify collaterals for the keeper as env variable

### Context
Currently our [keeper bot](https://github.com/sidestream-tech/unified-auctions-ui/tree/main/bot#readme) checks and tries to participate in auctions for all supported collaterals. Same with the twitter bot. Instead, we want the possibility to specify which collaterals should be "whitelisted". This however should be an optional variable and the default behaviour should stay the same.

### Tasks
- [x] introduce new env variable for the bot, eg `ETH-A, MATIC-A`
- [x] extend readme with explanation and example
- [x] parse provided value and validate it (providing `ETH-X` or `FKSALJF` should throw an error on startup)
- [x] extend logs with whitelisting status (eg `auctions: fetching only ETH-A, MATIC-A auctions`)
- [x] test bot using [our hardhat setup](https://github.com/sidestream-tech/unified-auctions-ui/tree/main/hardhat#readme)

Tejas-Choudekar:
closes #206 

To do:
Write documentation in readme.md

### DIFF
--- a/bot/src/keeper.ts
+++ b/bot/src/keeper.ts
@@ -7,10 +7,22 @@ import {
     getCollateralAuthorizationStatus,
     getWalletAuthorizationStatus,
 } from 'auctions-core/src/authorizations';
-import { ETHEREUM_NETWORK, KEEPER_MINIMUM_NET_PROFIT_DAI, KEEPER_WALLET_PRIVATE_KEY } from './variables';
+import { ETHEREUM_NETWORK, KEEPER_MINIMUM_NET_PROFIT_DAI, KEEPER_WALLET_PRIVATE_KEY, KEEPER_WHITELIST } from './variables';
 
 let isSetupCompleted = false;
 const currentlyExecutedAuctions = new Set();
+function isCollateralInWhitelist (collateralSymbol: string): boolean{
+    if(!KEEPER_WHITELIST){
+        return true;
+    }
+    //First parse the string and make it into an array
+    const whiteListArray = KEEPER_WHITELIST.split(',');
+
+    console.log('the item exist in the whitelist? ', whiteListArray.includes(collateralSymbol))
+    //Check if the collateral symbol is in your newly created array
+    return whiteListArray.includes(collateralSymbol);
+}
+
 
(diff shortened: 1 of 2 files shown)


benediktkaiser:
Tested on Localhost. LGTM but further testing would be appreciated! 

Screenshot 2022-04-13 at 15 35 41

bot/src/keeper.ts
```diff
@@ -40,6 +55,14 @@ const checkAndParticipateIfPossible = async function (auction: AuctionInitialInf
         return;
     }
 
+    // check if collateral is white listed
+    if (!isCollateralInWhitelist(auction.collateralSymbol)) {
+        console.info(
+            `keeper: auction "${auction.id}" was not executed because "${auction.collateralSymbol}" is not in the whitelist.`
+        );
+        return;
+    }
```
valiafetisov, line 64: As specified in the issue, this should be applied not only for the keeper, but for the twitter bot as well. Hint: we shouldn't even fetch auctions of the non-whitelisted collaterals

bot/src/keeper.ts
```diff
@@ -7,10 +7,25 @@ import {
...
+} from './variables';
 
 let isSetupCompleted = false;
 const currentlyExecutedAuctions = new Set();
+function isCollateralInWhitelist(collateralSymbol: string): boolean {
+    if (!KEEPER_WHITELIST) {
+        return true;
+    }
+    // First parse the string and make it into an array
+    const whiteListArray = KEEPER_WHITELIST.split(',');
```
valiafetisov, line 24: Given the string `"ETH-A, MATIC-A"` it will produce `" MATIC-A"` which will not be equal to `"MATIC-A"`

valiafetisov (commented):
Please check the tasks in the issue description that was addressed so far!

bot/src/keeper.ts
valiafetisov: Also, the env variable is parsed on every run of the function and not on startup as requested

PR #208, opened 2022-04-14
Added option to whitelist collaterals

### LINKED ISSUE #206: Whitelisting collaterals

benediktkaiser:
closes #206 

**Correctly setup:**
Screenshot 2022-04-14 at 06 36 07

The `1/7` in the message is structured as follows `WHITELISTED_AUCTIONS/ALL_AUCTIONS`. I thought this would be better to avoid confusion in case you forgot to deactivate your whitelist and you are wondering why there are so few auctions.

**Error in whitelist (shown on startup):**
Screenshot 2022-04-14 at 06 36 33

**No whitelist set:**
Screenshot 2022-04-14 at 06 36 53

_I was wondering, would it make sense to also include a `blacklist`? I was thinking, what if someone only wanted to exclude say two collaterals. They would have to make a very long list of every collateral except for the two they don't want._

### DIFF
--- a/bot/src/auctions.ts
+++ b/bot/src/auctions.ts
@@ -1,5 +1,7 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
 import { fetchAllInitialAuctions } from 'auctions-core/src/auctions';
+import { COLLATERAL_WHITELIST } from './variables';
+import { parseCollateralWhitelist } from './whitelist';
 
 const THRESHOLD_FOR_NEW_AUCTIONS = 5 * 60 * 1000;
 const knownAuctionIds = new Set();
@@ -10,6 +12,17 @@ const checkIfAuctionIsAlreadyKnown = function (auction: AuctionInitialInfo): boo
 
 const markAuctionAsKnown = function (auction: AuctionInitialInfo): void {
     knownAuctionIds.add(auction.id);
+};
+
+const checkIfAuctionCollateralIsInWhitelist = function (
+    auction: AuctionInitialInfo,
+    whitelist: string[] | undefined
+): boolean {
+    // if whitelist is disabled all auctions should be used
+    if (!whitelist) {
+        return true;
+    }
+    return whitelist.includes(auction.collateralType);
 };
 
 export const getNewAuctionsFromActiveAuctions = function (activeActions: AuctionInitialInfo[]): AuctionInitialInfo[] {
@@ -24,8 +37,14 @@ export const getNewAuctionsFromActiveAuctions = function (activeActions: Auction
 };
 
 export const getAllAuctions = async function (network: string): Promise<AuctionInitialInfo[]> {
+    let collateralWhitelist: undefined | string[];
+    if (COLLATERAL_WHITELIST) {
+        collateralWhitelist = parseCollateralWhitelist(COLLATERAL_WHITELIST);
+        console.info(`auctions: whitelist is enabled, only fetching auctions from "${COLLATERAL_WHITELIST}"`);
+    }
+
     const auctions = await fetchAllInitialAuctions(network);
     const auctionIds = auctions.map(auction => `"${auction.id}"`).join(', ');
     console.info(`auctions: found "${auctions.length}" auctions (${auctionIds}) on "${network}" network`);
-    return auctions;
+    return auctions.filter(auction => checkIfAuctionCollateralIsInWhitelist(auction, collateralWhitelist));
 };
--- a/bot/src/auctions.ts
+++ b/bot/src/auctions.ts
@@ -1,7 +1,6 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
 import { fetchAllInitialAuctions } from 'auctions-core/src/auctions';
-import { WHITELISTED_COLLATERALS } from './variables';
-import { parseCollateralWhitelist } from './whitelist';
+import { getWhitelistedCollaterals } from './whitelist';
 
 const THRESHOLD_FOR_NEW_AUCTIONS = 5 * 60 * 1000;
 const knownAuctionIds = new Set();
@@ -26,14 +25,7 @@ export const getNewAuctionsFromActiveAuctions = function (activeActions: Auction
 };
 
 export const getAllAuctions = async function (network: string): Promise<AuctionInitialInfo[]> {
-    let collateralWhitelist: undefined | string[];
-    if (WHITELISTED_COLLATERALS) {
-        collateralWhitelist = parseCollateralWhitelist(WHITELISTED_COLLATERALS);
-        console.info(`auctions: whitelist is enabled, only fetching auctions of type "${WHITELISTED_COLLATERALS}"`);
-    }
-
-    // if collateralWhitelist is undefined all auctions will be fetched
-    const auctions = await fetchAllInitialAuctions(network, collateralWhitelist);
+    const auctions = await fetchAllInitialAuctions(network, getWhitelistedCollaterals());
 
     const auctionIds = auctions.map(auction => `"${auction.id}"`).join(', ');
     console.info(`auctions: found "${auctions.length}" auctions (${auctionIds}) on "${network}" network`);
--- a/bot/src/auctions.ts
+++ b/bot/src/auctions.ts
@@ -1,6 +1,6 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
-import { fetchAllAuctionsFromCollateralTypes } from 'auctions-core/src/auctions';
-import { COLLATERAL_WHITELIST } from './variables';
+import { fetchAllInitialAuctions } from 'auctions-core/src/auctions';
+import { WHITELISTED_COLLATERALS } from './variables';
 import { parseCollateralWhitelist } from './whitelist';
 
 const THRESHOLD_FOR_NEW_AUCTIONS = 5 * 60 * 1000;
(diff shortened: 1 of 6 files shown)


valiafetisov:
> The `1/7` in the message is structured as follows `WHITELISTED_AUCTIONS/ALL_AUCTIONS`. I thought this would be better to avoid confusion in case you forgot to deactivate your whitelist and you are wondering why there are so few auctions.

It's a good idea from UX perspective, but it's not good in case there are 100+ auctions and your keeper is too slow fetching auctions that you don't need. As suggested in the issue: let's not even _fetch_ auctions that we don't need

> No whitelist set:

It's also nice to use actual env variable name in the message, so the user knows what to do: `collateral whitelisting: skipping setup due to missing WHITELISTED_COLLATERALS`

> They would have to make a very long list of every collateral except for the two they don't want.

I think it's more predictable to whitelist actions than blacklist since list of collaterals may be changed after bot update for example. Having both may also create confusion.

benediktkaiser:
## Updated parser to include whitespaces:
Screenshot 2022-04-20 at 08 41 26

## No whitelist set (all auctions are fetched):
Screenshot 2022-04-20 at 08 42 42

## Whitelist set (only collateralTypes in whitelist are fetched):
Screenshot 2022-04-20 at 08 42 59

core/src/auctions.ts
```diff
@@ -110,6 +111,27 @@ export const fetchAllInitialAuctions = async function (network: string): Promise
...
+): Promise<AuctionInitialInfo[]> {
+    // If no collateralTypes list is given return all auctions from all collateralTypes
+    if (!collateralTypes) {
+        return await fetchAllInitialAuctions(network);
+    }
+
+    // Map through given collateral types and validate that our UI supports them
+    collateralTypes.forEach(collateralType => {
+        getCollateralConfigByType(collateralType);
+    });
```
valiafetisov, line 126: Didn't you already checked that?

core/src/auctions.ts
```diff
@@ -110,6 +111,27 @@ export const fetchAllInitialAuctions = async function (network: string): Promise
     return auctionGroups.flat();
 };
 
+export const fetchAllAuctionsFromCollateralTypes = async function (
+    network: string,
+    collateralTypes: string[] | undefined
+): Promise<AuctionInitialInfo[]> {
```
valiafetisov, line 117: The title is misleading, I would instead just extended existing `fetchAllInitialAuctions` with optional `collateralTypes` prop

bot/src/variables.ts
```diff
@@ -1,3 +1,4 @@
 export const ETHEREUM_NETWORK = process.env.ETHEREUM_NETWORK || 'kovan';
 export const KEEPER_MINIMUM_NET_PROFIT_DAI = parseInt(process.env.KEEPER_MINIMUM_NET_PROFIT_DAI || '');
 export const KEEPER_WALLET_PRIVATE_KEY = process.env.KEEPER_WALLET_PRIVATE_KEY;
+export const COLLATERAL_WHITELIST = process.env.WHITELISTED_COLLATERALS || false;
```
valiafetisov, line 4: Can you please rename it to `WHITELISTED_COLLATERALS` everywhere to stay consistent?

bot/src/auctions.ts
```diff
@@ -24,7 +26,15 @@ export const getNewAuctionsFromActiveAuctions = function (activeActions: Auction
...
 export const getAllAuctions = async function (network: string): Promise<AuctionInitialInfo[]> {
-    const auctions = await fetchAllInitialAuctions(network);
+    let collateralWhitelist: undefined | string[];
+    if (WHITELISTED_COLLATERALS) {
+        collateralWhitelist = parseCollateralWhitelist(WHITELISTED_COLLATERALS);
+        console.info(`auctions: whitelist is enabled, only fetching auctions of type "${WHITELISTED_COLLATERALS}"`);
+    }
+
+    // if collateralWhitelist is undefined all auctions will be fetched
+    const auctions = await fetchAllInitialAuctions(network, collateralWhitelist);
```
valiafetisov, line 36: And then here you will execute `getWhitelistedCollaterals` in order to get a valid list or undefined
```suggestion
    const auctions = await fetchAllInitialAuctions(network, getWhitelistedCollaterals());
```

bot/src/variables.ts
```diff
@@ -1,3 +1,4 @@
 export const ETHEREUM_NETWORK = process.env.ETHEREUM_NETWORK || 'kovan';
 export const KEEPER_MINIMUM_NET_PROFIT_DAI = parseInt(process.env.KEEPER_MINIMUM_NET_PROFIT_DAI || '');
 export const KEEPER_WALLET_PRIVATE_KEY = process.env.KEEPER_WALLET_PRIVATE_KEY;
+export const WHITELISTED_COLLATERALS = process.env.WHITELISTED_COLLATERALS || false;
```
valiafetisov, line 4: Please don't mix types, if we expect string, let's keep it string, undefined will still be falsy

bot/src/whitelist.ts
```diff
@@ -0,0 +1,21 @@
+import { getCollateralConfigByType } from 'auctions-core/src/constants/COLLATERALS';
+import { WHITELISTED_COLLATERALS } from './variables';
+
+const validateWhitelist = function (whitelist: string[]) {
+    whitelist.forEach(collateralType => {
+        getCollateralConfigByType(collateralType);
+    });
```
valiafetisov, line 7: the cherry on top would be to give user a list of collaterals that threw error, not just a first one (but that's not required to implement)

bot/src/auctions.ts
```diff
@@ -24,7 +26,15 @@ export const getNewAuctionsFromActiveAuctions = function (activeActions: Auction
 };
 
 export const getAllAuctions = async function (network: string): Promise<AuctionInitialInfo[]> {
-    const auctions = await fetchAllInitialAuctions(network);
+    let collateralWhitelist: undefined | string[];
+    if (WHITELISTED_COLLATERALS) {
+        collateralWhitelist = parseCollateralWhitelist(WHITELISTED_COLLATERALS);
+        console.info(`auctions: whitelist is enabled, only fetching auctions of type "${WHITELISTED_COLLATERALS}"`);
```
valiafetisov, line 32: This console log should be inside setup, otherwise it will be printed all the time, right?

benediktkaiser: I would actually reprint it every time auctions are fetched. If someone where to enter the logs of the bot, after it has already been running for a few days, it might be nice to easily see which auctions are being fetched in the console. If we only show it on setup the logs would be too far up. 

What do you think?

bot/src/whitelist.ts
benediktkaiser: We do like cherries! 

Screenshot 2022-04-21 at 16 57 04

bot/src/whitelist.ts
```diff
@@ -1,20 +1,41 @@
 import { getCollateralConfigByType } from 'auctions-core/src/constants/COLLATERALS';
 import { WHITELISTED_COLLATERALS } from './variables';
 
-const validateWhitelist = function (whitelist: string[]) {
-    whitelist.forEach(collateralType => {
-        getCollateralConfigByType(collateralType);
+const validateWhitelist = function (whitelist: string[]): Promise<void> {
+    return new Promise((resolve, reject) => {
+        const errors: string[] = [];
```
valiafetisov, line 6: Those are not `errors`, but `notFoundCollateralTypes` or `unsupportedCollateralTypes`

bot/src/whitelist.ts
```diff
@@ -1,20 +1,41 @@
 import { getCollateralConfigByType } from 'auctions-core/src/constants/COLLATERALS';
 import { WHITELISTED_COLLATERALS } from './variables';
 
-const validateWhitelist = function (whitelist: string[]) {
-    whitelist.forEach(collateralType => {
-        getCollateralConfigByType(collateralType);
+const validateWhitelist = function (whitelist: string[]): Promise<void> {
+    return new Promise((resolve, reject) => {
```
valiafetisov, line 5: I don't see anything asynchronous in this function, why did you decide return promise here?

bot/src/auctions.ts
valiafetisov: I think it's ok to print it only once, but if you want, you can maybe add this info to the existing log about fetching? https://github.com/sidestream-tech/unified-auctions-ui/blob/2858a7e72d656e218ac2a0e497faaa82fe580196/bot/src/auctions.ts#L31

bot/src/whitelist.ts
benediktkaiser: I don't know what I was doing there to be honest 😅. I removed it now!

bot/src/auctions.ts
benediktkaiser: I thought about this again and I actually changed my mind. I noticed that if the user has a very long whitelist, that this is going to spam the console way too much. 

Therefore I agree with you, that it does not make sense to display it with every fetch.

valiafetisov (approved):

LukSteib (approved):
Tested via hardhat fork:
* whitelisting collateral types works as expected ✔️ 
* not whitelisting any collateral type leads to keeper fetching all auctions ✔️ 
* whitelisting an unknown collateral type yields keeper error ✔️ 

---
Unrelated question:
There are no actual price drops on hardhat fork. The 30 sec refetching always sets back price to its initial value. I assume this is expected and I was just not aware of it up until now, right?

valiafetisov:
> I assume this is expected and I was just not aware of it up until now, right?

Yeah, that's correct, the block time / state stays frozen up until something is mined (eg authorization). I'm not sure if there is nice solution to that except if we want to "mine" blocks from time to time by themselves

valiafetisov:
Sidenote: we can actually add the same env variable to the frontend. But do we want to?

PR #211, opened 2022-04-14
Fix bid with NaN

valiafetisov:
Fixes a recently discovered edge case with `Bid with DAI` transaction by disabling a button. Later we would need to add another panel explaining [`Clipper/no-partial-purchase` error](https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/clip.sol#L374)

Screenshot 2022-04-14 at 17 05 23

### DIFF
--- a/frontend/components/transaction/BidTransaction.vue
+++ b/frontend/components/transaction/BidTransaction.vue
@@ -64,7 +64,8 @@
                 !isWalletConnected ||
                 !isEnoughDeposited ||
                 !isWalletAuthorizedCorrect ||
-                !isCollateralAuthorized
+                !isCollateralAuthorized ||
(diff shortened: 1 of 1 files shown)


LukSteib (approved):
Approved based on screenshot (since none of the testnet auctions shows `Unknown` right now)

valiafetisov:
This one shows http://localhost:3000/collateral?network=goerli&auction=LINK-A%3A1

valiafetisov:
We definitely have to add another panel, [since this state is very cryptic](https://github.com/sidestream-tech/unified-auctions-ui/pull/211#issuecomment-1099295528): everything is correct, but not possible to participate

PR #218, opened 2022-04-19
UniswapV3Callee support

### LINKED ISSUE #137: Support `UniswapV3Callee` callee
### Goal
Swap regular ERC20 collaterals via new `UniswapV3Callee` callee instead of the old `UniswapV2CalleeDai`

### Context
Since Uniswap V2 is loosing popularity and doesn't provide good source of the liquidity anymore, we should add support for the `UniswapV3Callee` callee. Reference implementation can be found in the [auction-demo-keeper repo](https://github.com/makerdao/auction-demo-keeper/tree/main/src) and we also already use Uniswap V3 inside `WstETHCurveUniv3Callee` (which we should split into separate function instead): 
https://github.com/sidestream-tech/unified-auctions-ui/blob/5494c084e8f42cbdb962e39651b136c3dda196a6/core/src/calleeFunctions/WstETHCurveUniv3Callee.ts#L51-L63

Note: according to the auction-demo-keeper config, [one collateral is still using the old UniV2 callee](https://github.com/makerdao/auction-demo-keeper/blob/6db7431aa07b657d4e9cf2aae35f9c1302595a79/config/mainnet.json#L183-L195), therefore we should:
- investigate the reasoning
  - is it about the liquidity size?
  - do we need to dynamically/manually compare prices between v2 vs v3 exchanges?
- do not drop support for the old v2 callee

### Tasks
- [ ] Implement callee functions, according to the [Collateral onboarding](https://github.com/sidestream-tech/unified-auctions-ui/tree/main/core#collateral-onboarding) guide
- [ ] Update `WstETHCurveUniv3Callee` to use new Univ3 helpers
- [ ] Investigate the reasoning to keep both
  - [ ] Potentially change type of the `COLLATERALS[collateralType].exchange.callee` from `string` to `string[]`

valiafetisov:
Closes #137

### DIFF
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -93,7 +93,6 @@ export const enrichAuctionWithPriceDrop = async function (auction: Auction): Pro
             transactionGrossProfitDate,
         };
     } catch (error) {
-        console.warn(`auction price drop information is not available for auction "${auction.id}"`);
         return {
             ...auction,
             totalPrice: auction.collateralAmount.multipliedBy(auction.unitPrice),
--- a/core/src/calleeFunctions/UniswapV3Callee.ts
+++ b/core/src/calleeFunctions/UniswapV3Callee.ts
@@ -1,0 +1,45 @@
+import type { CalleeFunctions, CollateralConfig } from '../types';
+import { ethers } from 'ethers';
+import BigNumber from '../bignumber';
+import { getContractAddressByName, getJoinNameByCollateralType } from '../contracts';
+import { convertCollateralToDaiUsingRoute, encodeRoute } from './helpers/uniswapV3';
+
+const getCalleeData = async function (
+    network: string,
+    collateral: CollateralConfig,
+    profitAddress: string
+): Promise<string> {
+    if (collateral.exchange.callee !== 'UniswapV2CalleeDai') {
+        throw new Error(`getCalleeData called with invalid collateral type "${collateral.ilk}"`);
+    }
+    const joinAdapterAddress = await getContractAddressByName(network, getJoinNameByCollateralType(collateral.ilk));
+    const minProfit = 0;
+    const uniswapV3route = encodeRoute(network, [collateral.symbol, ...collateral.exchange.route]);
+    const typesArray = ['address', 'address', 'uint256', 'bytes', 'address'];
+    return ethers.utils.defaultAbiCoder.encode(typesArray, [
+        profitAddress,
+        joinAdapterAddress,
+        minProfit,
+        uniswapV3route,
+        ethers.constants.AddressZero,
+    ]);
+};
+
+const getMarketPrice = async function (
+    network: string,
+    collateral: CollateralConfig,
+    collateralAmount: BigNumber
+): Promise<BigNumber> {
+    // convert collateral into DAI
+    const daiAmount = await convertCollateralToDaiUsingRoute(network, collateral.symbol, collateralAmount);
+
+    // return price per unit
+    return daiAmount.dividedBy(collateralAmount);
+};
+
+const UniswapV2CalleeDai: CalleeFunctions = {
+    getCalleeData,
+    getMarketPrice,
+};
+
+export default UniswapV2CalleeDai;
--- a/core/src/calleeFunctions/UniswapV3Callee.ts
+++ b/core/src/calleeFunctions/UniswapV3Callee.ts
@@ -9,12 +9,12 @@ const getCalleeData = async function (
     collateral: CollateralConfig,
     profitAddress: string
 ): Promise<string> {
-    if (collateral.exchange.callee !== 'UniswapV2CalleeDai') {
+    if (collateral.exchange.callee !== 'UniswapV3Callee') {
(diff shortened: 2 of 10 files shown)


valiafetisov:
But I would actually wait merging it, cause it seems that price for a lot of collaterals actually better on uniswap v2, so we may consider adding support for multiple collees per collateral.

UniV3 prices for TUSD
Screenshot 2022-04-19 at 22 23 56

UniV2 prices for TUSD
Screenshot 2022-04-19 at 22 24 09

You can compare them yourself by opening [dashboard on staging](https://main.auction-ui.k8s.sidestream.tech/dashboard) and comparing it with the locally running branch http://localhost:3000/dashboard?network=mainnet

valiafetisov:
So the proposal for this PR is:
- [ ] to deploy callee onto test networks?
- [ ] to confirm that callee works for all regular collaterals
- [ ] to check which collaterals are currently cheaper via new callee vs old using `/dashboard`?
- [ ] to apply callee to those collaterals from above and merge
- [ ] discuss the way forward: I propose to introduce support for multiply callees / routes so the either the user can select the exchange they want to use or we can suggest / select one automatically. Both require changing how callees are configured, market prices are stored as well as some rework of the UI

LukSteib:
>  to deploy callee onto test networks?

Agree. Since we are otherwise unable to test flashloan based execution on `goerli` for the collaterals that rely on uniV3 callee

>  to confirm that callee works for all regular collaterals

getting a bidding error. See console error [1]. Tested for collateral types `UNI-A` & `ETH-C` on hardhat fork#


>  to check which collaterals are currently cheaper via new callee vs old using /dashboard?

Created a dashboard overview that shows price differences but also liquidity differences between selected collaterals on uniV2 and uniV3: https://docs.google.com/spreadsheets/d/1JdHTCLiGun2R0SSTutk4d3PhAjhDBGPJcLLR-ySohAE/edit?usp=sharing

Based on that and personal discussion I propose to: 
* sacrifice better liquidity on V3 for some better prices on V2 and move forward with the implementation to replace v2 with v3 callee for all standard ERC-20 collaterals
* Create a follow up issue that allows to switch between the callees 
* Create a follow up issue to include liquidity information per collateral to `/dashboard` table

>  to apply callee to those collaterals from above and merge

See previous comment

> discuss the way forward: I propose to introduce support for multiply callees / routes so the either the user can select the exchange they want to use or we can suggest / select one automatically. Both require changing how callees are configured, market prices are stored as well as some rework of the UI

I like the proposal to let the user switch between relevant callees. A nice add-on and to reduce complexity for user I would like to add requirement that more profitable callee is preselected. (At least check feasibility of pre-executing and comparing to then preselect is feasible) 


[1]
`Bidding error Error: getCalleeData called with invalid collateral type "UNI-A"`

valiafetisov:
> getting a bidding error. See console error [1]

The error should be fixed, tested on the hardhat

valiafetisov:
> sacrifice better liquidity on V3 for some better prices on V2 and move forward with the implementation to replace v2 with v3 callee for all standard ERC-20 collaterals

Done

valiafetisov:
> to deploy callee onto test networks

Currently I'm still getting the `Transaction mined but execution failed` error while trying to deploy the same callee contract as on the mainnet to a testnet with appropriate constructor parameters:

status | false Transaction mined but execution failed
:-- | :--
transaction hash | [0xf21ba88e524f9e79869eece624981c9cbd12fde8ca215210b3643802329f3cf4](https://dashboard.tenderly.co/tx/goerli/0xf21ba88e524f9e79869eece624981c9cbd12fde8ca215210b3643802329f3cf4)
decoded input | { 	"address uniV3Router_": "0xE592427A0AEce92De3Edee1F18E0157C05861564", 	"address daiJoin_": "0xE592427A0AEce92De3Edee1F18E0157C05861564" }

Debugging it doesn't give more insight:
```
creation of UniswapV3Callee pending...
creation of UniswapV3Callee errored: VM error: revert.

revert
	The transaction has been reverted to the initial state.
```

valiafetisov:
After cleaning up remix editor and restarting the process again with the same files and arguments, was able to deploy the callee to both test networks that we support:

https://goerli.etherscan.io/address/0x323533ad6355ade4ecfa97e49de7cba27dff1208#code
https://kovan.etherscan.io/address/0xd028e722c00434182c9558d84802bb5706219d26#code

Although there is on ETH that have liquidity on goerli and a few ones on kovan, so test chains apparently becoming less and less useful:

http://localhost:3000/dashboard?network=goerli
http://localhost:3000/dashboard?network=kovan

LukSteib (approved):

LukSteib:
> Although there is on ETH that have liquidity on goerli and a few ones on kovan, so test chains apparently becoming less and less useful:

The `missing revert data` error and the fact that auctions (other than ETH) are no longer tradable on `goerli` and `kovan` is due to the missing liquidity you are referring to? Just want to ensure that the problem is not on our side


Revert errors for most collaterals on `goerli` and `kovan` (dashboard view) after introduction of V3 callee:
image

PR #224, opened 2022-04-20
Added `Collaterals` popup

### LINKED ISSUE #214: Create collaterals popup
### Goal

User is able to see authorization status on the supported collaterals, their balances in VAT and able to withdraw all collateral.

### Context

Currently, user is able to see collateral balance in VAT only if the is an active auction of this type. This means there is no option to withdraw collateral when there are no auctions and there is no option to pre-authorize it. Therefore, we want to create a new popup with this information and actions. 

### Assets

<details><summary>The mock</summary>

MakerDAO@2x (5)

</details>

### Tasks
- [x] create a `ManageCollateralTable` component and its stories according to the mock
  - [x] provide data as a single object and a new type `CollateralStatus`: `{type: string; address: string; isAuthorized: boolean; balance: BigNumber;}`
  - [x] Sort collaterals alphabetically
  - [x] Add icons like in the dashboard
  - [x] Token addresses should be clickable
  - [ ] Propose better explanations text above the table
  - [x] Make sure that if there is no authorizations, withdrawal button is disabled
  - [x] Make sure that if balance is 0, withdrawal button is disabled
- [x] create `ManageCollateralPopup` and its container (as with the wallet, fetching only should start on popup open)
- [x] add new option `Manage Collaterals in VAT` to the wallet selector dropdown

fegbert:
closes #214 

**preview**
image


frontend/components/ManageCollateralTable.vue
```diff
@@ -0,0 +1,144 @@
...
+    @apply Element text-gray-500 dark:text-gray-300;
+}
+
+.Button {
+    @apply text-primary hover:text-primary-light underline;
+}
+
+.Button:disabled {
+    @apply text-gray-400 dark:text-gray-500 no-underline cursor-not-allowed;
+}
```
valiafetisov, line 143: Then this can be removed

frontend/store/collaterals.ts
```diff
@@ -69,6 +80,62 @@ export const actions = {
...
     },
+    async fetchCollateralStatuses({ commit, dispatch, rootGetters }: ActionContext<State, State>) {
+        const network = rootGetters['network/getMakerNetwork'];
+        if (!network) {
+            return;
+        }
+        for (const collateral of Object.values(COLLATERALS)) {
+            const tokenAddress = await getTokenAddressByNetworkAndSymbol(network, collateral.symbol).catch(() => {
+                return undefined;
+            });
```
valiafetisov, line 91: Why do you need to implement collateral fetching 2 times? Can't you simply call `fetchCollateralStatus` here in the loop?

frontend/store/collaterals.ts
```diff
@@ -69,6 +80,62 @@ export const actions = {
...
+        const tokenAddress = await getTokenAddressByNetworkAndSymbol(network, collateral.symbol).catch(() => {
+            return undefined;
+        });
+        if (!tokenAddress) {
+            return;
+        }
+        await dispatch('wallet/fetchCollateralVatBalance', collateralType, { root: true });
+        await dispatch('authorizations/fetchCollateralAuthorizationStatus', collateralType, { root: true });
+        const isAuthorized = rootGetters['authorizations/collateralAuthorizations'].includes(collateralType);
+        const balance = rootGetters['wallet/collateralVatBalanceStore'][collateralType];
```
valiafetisov, line 130: You don't need to take it from one store here and put it in another store. You can do it at the time you return that data (inside the getter you can join data from different places). See next comment

frontend/containers/ManageCollateralModalContainer.vue
```diff
@@ -0,0 +1,78 @@
...
+        },
+    },
+    methods: {
+        ...mapActions('wallet', ['withdrawAllCollateralFromVat']),
+        setManageCollateralModal(open: boolean): void {
+            this.$store.commit('modals/setManageCollateralModal', open);
+        },
+        async authorize(collateralType: string) {
+            await this.$store.dispatch('authorizations/authorizeCollateral', collateralType);
+            await this.$store.dispatch('collaterals/fetchCollateralStatus', collateralType);
```
valiafetisov, line 70: This will allow you to not refetch the whole thing again (and 2 times) since `authorizations/authorizeCollateral` already refetches the status itself
```suggestion
```

frontend/containers/ManageCollateralModalContainer.vue
```diff
@@ -0,0 +1,78 @@
...
+        },
+        async withdraw(collateralType: string) {
+            await this.$store.dispatch('wallet/withdrawAllCollateralFromVat', collateralType);
+            await this.$store.dispatch('collaterals/fetchCollateralStatus', collateralType);
```
valiafetisov, line 74: Same here
```suggestion
```

frontend/components/ManageCollateralTable.vue
```diff
@@ -0,0 +1,128 @@
+<template>
+    <div>
+        <TextBlock v-if="isExplanationsShown" class="mb-2">
+            This is a list of collaterals supported by the Maker. Each row provides the possibility to withdraw
```
LukSteib, line 4: - [ ] change to "supported by the Maker Protocol"

valiafetisov:
> How is it possible that for ETH-A and ETH-B different withdrawal amounts are fetched (see 2 below). I am expecting the same underlying ERC-20 token especially since the same token contract address is displayed as well

In VAT they are stored by collateral types, not by tokens, so it's correct that different amounts are displayed. Same logic applies to the authorizations and withdrawal executions.

LukSteib:
> In VAT they are stored by collateral types, not by tokens, so it's correct that different amounts are displayed. Same logic applies to the authorizations and withdrawal executions.

I see, thanks. So generally speaking to withdraw all of my `ETH` from `VAT` I probably would have to do 3 auth steps (`ETH-A`, `ETH-B`, `ETH-C`) and 3 withdrawals repsectively (given I engage in auction for all of the three different ilk types).

fegbert:
Finished the decoupling of the collateral auth loading states. 
Still missing is the decoupling of the withdrawal loading and the re-fetching of the withdrawal amount without having to re-opening the popup.

LukSteib (approved):

PR #240, opened 2022-05-03
Added `Notification-Service` for contract events

### LINKED ISSUE #235: Implement notification service for chainlog
**Goal:**
We get notified when chainlog contract emits an event

**Context:**
During the offboarding of `T-USD` the `abacus` contract for this collateral type was changed right before the first liquidations were started. We recognised the change quickly, however with a notification service in place our response time might be improved even further. Additionally a notification would also help tracking the onboarding of new collateral types since it can inform when relevant contracts (e.g. `clipper`) for the new collateral are deployed. 
Given the outlined motivation we want to build a notification service. 

Relevant input:
* chainlog contract: https://etherscan.io/address/0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F
* chainlog contract can emit various events: `UpdateVersion`, `UpdateAddress`, ...
* chainlog contract exists on testnets (e.g. `goerli`) 

Desired flow:
* chainlog contract emits an event
  * most relevant event: `UpdateAddress`
* we receive an email that informs about the event
  * email content informs in understandable terms about the event
  * e.g. for `UpdateAddress`: 
    * What address has been updated?
    * What is the contract name of the new address?
    * When did the update happen?
    * On which network did the event happen?

**Tasks:**
- [x] Clarify open questions to understand scope
- [x] outline implementation approach in issue
- [x] implement notification service
- [x] find a way to test

benediktkaiser:
closes #235 

For the first version I made the following simple UI:

Console view:
Screenshot 2022-05-04 at 08 06 31

Email Template:
Screenshot 2022-05-04 at 08 31 15
_Email links to [Etherscan](https://kovan.etherscan.io/tx/0xb7d44d618366f5a6158c4889edfd13a62644486da8f87ed0e6af3b0b2d4938f7#eventlog)._

I was thinking that this would be enough as a first version, as you can view the data on etherscan. I can work on a second PR afterwards that takes care of actually displaying the differences in the email. In that PR I will also include the functionality to customize the email message per subscription, as outlined by the issue. However I wanted to get a first version up and running before diving into this. I hope this is okay!

## How to test @LukSteib 

As the event we later want to observe, does not have many updates, I added a second, quite frequently used event as a test. 

In `notifcation-service/src/constants/SUBSCRIPTIONS.ts` You can re-add the following lines:

https://github.com/sidestream-tech/unified-auctions-ui/blob/add-notification-service/notification-service/src/constants/SUBSCRIPTIONS.ts#L24-L28

_Note: The commented out contract only works on the  `kovan` network, so be sure to test it on there or use a alternative contract._

### Email setup
By default, if no SMTP values are provided, I connect with a service called [Ethereal Email](https://ethereal.email/), as this was included in my library. However, this service is down today 🙃, therefore you need to add your own `SMTP` settings.

I would advise to use a service called [MailTrap](https://mailtrap.io/). Once you logged in and click on your inbox you will find a dropdown titled `Integrations`. By default it should have something called `cURL` selected. Select `NodeMailer` from the list. 

You can then set the environment variables to the following:
- SMTP_HOST="smtp.mailtrap.io"
- SMTP_PORT="2525"
- SMTP_USERNAME="YOUR_USER_ID"
- SMTP_PASSWORD="YOUR_USER_PASSWORD"

What is also a bit different in this project is how to start it. Running `npm run dev` will only live compile it, but not execute it. Therefore you should run `npm run start`. This will automatically build and start the service.


notification-service/src/recievers.ts
```diff
@@ -0,0 +1,33 @@
+import { RECEIVERS } from './variables';
+
+const validateReceivers = function (receivers: string[]) {
+    const invalidEmails: string[] = [];
+    receivers.forEach(receiver => {
+        if (!receiver.includes('@') || !receiver.includes('.')) {
```
valiafetisov, line 6: `.@` would be a valid email, right? What about using a package instead?
- https://www.npmjs.com/package/email-validator
- https://www.npmjs.com/package/validator

notification-service/src/variables.ts
```diff
@@ -0,0 +1,10 @@
+export const INFURA_PROJECT_ID = process.env.INFURA_PROJECT_ID;
+export const ETHEREUM_NETWORK = process.env.ETHEREUM_NETWORK || 'mainnet';
+
+export const SMTP_HOST = process.env.SMTP_HOST;
+export const SMTP_PORT = parseInt(process.env.SMTP_PORT || '465');
+export const SMTP_USERNAME = process.env.SMTP_USERNAME;
+export const SMTP_PASSWORD = process.env.SMTP_PASSWORD;
+export const SMTP_EMAIL = process.env.SMTP_EMAIL || 'notification@unified-auctions.makerdao.com';
```
valiafetisov, line 8: Please don't hardcode any emails, as specified in the issue

notification-service/src/recievers.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+    return undefined;
+};
+
+export const setupReceiverList = function () {
+    if (RECEIVERS) {
+        const parsedList = parseReceiverList(RECEIVERS);
+        validateReceivers(parsedList);
+    } else {
+        console.warn(`receivers: skipping setup due to missing RECEIVERS`);
```
valiafetisov, line 31: Why is `RECEIVERS` optional?

notification-service/src/mailer.ts
```diff
@@ -0,0 +1,61 @@
...
+    mailData: {
+        eventName: string;
+        contractAddress: string;
+        transactionHash: string;
+    }
+) {
+    const etherscanURL = `${getEtherscanURL(network)}/tx/${mailData.transactionHash}#eventlog`;
+    const info = await transporter.sendMail({
+        from: SMTP_EMAIL,
+        to: RECEIVERS,
```
valiafetisov, line 50: This doesn't seems optional to me :)

notification-service/README.md
```diff
@@ -0,0 +1,75 @@
...
+
+In order to add a custom event subscription you need to edit the file `/constants/SUBSCRIPTIONS.ts`. Inside you will
+find an array of Event Subscriptions. You can add your own with the following format:
+
+```js
+export const SUBSCRIPTIONS: EventSubscription[] = [
+  {
+    id: 'ChainlogUpdateVersion',
+    contract: '0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F',
+    abi: CHAINLOG_ABI,
```
valiafetisov, line 51: Can we please fetch this from etherscan? https://docs.etherscan.io/api-endpoints/contracts#get-contract-abi-for-verified-contract-source-codes

notification-service/src/helpers/generateEmail.ts
```diff
@@ -0,0 +1,178 @@
+export function generateTextEmail(eventName: string, contractAddress: string, link: string, network: string) {
+    return `We just detected an update of the event "${eventName}" from the contract "${contractAddress}" on the network "${network}". View more info on ${link}`;
+}
+
+export default function generateEmail(eventName: string, contractAddress: string, link: string, network: string) {
```
valiafetisov, line 5: I think for the first version it would've been enough to send text-only emails (with links) (:

Where did you get this html?

valiafetisov (commented):
The PR is missing functionality:
- to subscribe different emails to different events
- to specify `SUBSCRIPTIONS` via env variable (instead of hardcoding specific addresses)
- to parse event data / fetch previous data (as specified above)

notification-service/src/helpers/generateEmail.ts
benediktkaiser: The template is free to use and has a MIT License. 

You can find it here:;

notification-service/README.md
benediktkaiser: Do we have an API for this? 

https://docs.etherscan.io/getting-started/viewing-api-usage-statistics
I can create my own for free, however we should look into generating one for production then.

benediktkaiser: In addition to this. Don't we want to implement this into `/core` as well then? Or is there some kind of caveat (aside from the rate limiting and generally us having to use more api calls)?

benediktkaiser:
> * [ ]  to specify `SUBSCRIPTIONS` via env variable (instead of hardcoding specific addresses)

We decided against this in the daily, as I later want to implement a feature that will let you customize the data shown in the email. 
As JSON formats don't support functions, this forces use to keep it in a typescript file.

LukSteib:
I was able to get the service started and receive notification mails. Looking good already 👍
I like to have the link to the specific event on etherscan included in the mail - so let's keep that for now.

To finish this first version I would like to include the requirement of event data parsing / fetch previous data into the scope of this PR. 
For the specific case of `UpdateVersion` event I would imagine to get the old contract address and the new contract address in the mail already. 
Since we are building with generalisability in mind we might think about how to properly make the html text customisable based on the event to be tracked (as you already pointed out in your comment above)

valiafetisov:
> As JSON formats don't support functions, this forces use to keep it in a typescript file

> we might think about how to properly make the html text customisable based on the event to be tracked

It's a harder question on how to generalise formatting, which needs to be conceptualised after implementing several types of formatting. I guess it should be possible to statically define it to some extend (eg compare with previous event or not, specify value type, format to number of decimal points, etc – sadly it can't be done completely automatically)

notification-service/README.md
valiafetisov: > Don't we want to implement this into `/core` as well then?

No, as specified in the issue discussion, we don't share code in this service with the `core`

notification-service/src/helpers/generateEmail.ts
valiafetisov: It's ok, I guess! Please link this repo in the code as a comment if we can't import this file directly

In the future, we may want to setup [vue + mjml](https://kevcodez.de/posts/2019-10-12-responsive-emails-vuejs-mjml/) for the email and preview it in storybook so we can do more complex formatting, while still working with the tools we used to work

benediktkaiser:
As discussed in the calls:
We decided to postpone the fetching of previous events due to the following scenario. 

We are looking into the `updateAddress` event. The `updateAddress` event always has a `key` and a `address`. If we were to only fetch the last event, the keys may not match up, therefore providing us no useful information. 

In this scenario, we would have to fetch the last event with the same key. However this would mean that the function to fetch the previous event would need to become hard coded for every event. 

Therefore it will not be implemented in version 1 of this service.

benediktkaiser:
I ran into an issue myself. 

If I set the event, I want to observe to `*`, everything works correctly. However if I set it to a specific name, say `Transfer`, the event is called, when a new event is created, however the event data passed back is `undefined`.

```console
📢  [event]: event "undefined" triggered in block "undefined". Attempting to send email. 

TypeError: Cannot read properties of undefined (reading 'src')
    at Object.formatData
    at FragmentRunningEvent.<anonymous>
    at Timeout._onTimeout 
    at listOnTimeout (node:internal/timers:564:17)
    at process.processTimers (node:internal/timers:507:7)
```

@valiafetisov do you have any idea, what might cause this? I did some googling online, however I could not find anything.

valiafetisov:
> do you have any idea, what might cause this? I did some googling online, however I could not find anything.

Well, I don't think you can google it, since it's most probably implementation problem. Please check the source code of the methods you're using or their documentation. Does the event name get passed to the provider(infura) or is it subscribes to all events and then parses only specified events?

Starting point: https://github.com/ethers-io/ethers.js/blob/master/packages/contracts/src.ts/index.ts#L1060-L1063

If it doesn't lead you anywhere the workaround would be to subscribe to all and filter on our side, right?

benediktkaiser:
> If it doesn't lead you anywhere the workaround would be to subscribe to all and filter on our side, right?

After looking into it, I sadly could not find a solution. Therefore I went with your purposel.  My websocket fires with every event, but I check if it matches. 

The only downside: 
Say I am listening to two separate events on the same contract: I will now be firing the websocket twice. Therefore I thought of the following optimisation:

I will loop through all the SUBSCRIPTIONS, but will loop through their contracts. Then I will only open one socket per contract and loop the SUBSCRIPTIONS inside that.

benediktkaiser:
I updated the system to use the following behavior: 

For every contract that needs to receive a web socket, one is opened. Every contract has a list of Subscriptions, that listen to it. These are then checked when an event is registered in the contract. Therefore the console messages now look like this:

Screenshot 2022-05-04 at 15 46 29

This should hopefully cost less resources, as we need far fewer web sockets.

valiafetisov:
Let's also get the vocabulary straight, I though:
 - we `open` only one `websocket connection` to infura. Then the log message should be:
   - `websocket connection opened to "kovan" network`
 - we `subscribe` to different events using the same `websocket connection` or websocket provider:
   - `subscribed to "XXX, EEE" events on "0xdA0Able0017DEbCd72Be8599041a2aa3bA7e740F" address`

Are those correct assumptions or what was your actual question?

Also, regarding `setting up custom mailing service` – I think it's more important to notify if the setup completed successfully rather than if it has been started (at least that's how the message reads now)

benediktkaiser:
> Are those correct assumptions or what was your actual question?

Yes, that is exactly what I was trying to say. Sorry, about my lack of correct terminology. 

> Also, regarding `setting up custom mailing service` – I think it's more important to notify if the setup completed successfully rather than if it has been started (at least that's how the message reads now)

I changed this! I also saw that the other email mocking site is now working again. 
If you do not add any SMTP settings it will automatically use it. Every time an Email is sent, the link to preview the mail will be printed in the console:
https://ethereal.email/message/YnKNaSkpyOuhP9GuYnKNejTRrgdcXRAPAAAAAsT9qbAosObElsikjcBead4

This is very nice for quick testing!

benediktkaiser:
I added CI steps now. This does not include any steps for development. As agreed on during the daily, this will be done by DevOps in a separate issue (#245).

The one thing that bothers me about my implementation of the CI is that one step always fails: https://drone.sidestream.tech/sidestream-tech/unified-auctions-ui/817/5/3. Can I fix this or does DevOps need to have a look?

valiafetisov:
> This does not include any steps for development

You mean the deployment?

> one step always fails

That step is to pull existing image to use it as a cache. Sine there was no image published yet it will fail, but that's expected! That's why the overall status of the CI is green.

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `ETHEREUM_NETWORK`: (optional, default `kovan`) – internal network name on which the bot poll for auctions. Available
+- `SMTP_*`: (required) - SMTP data for the service to send emails
+    - `SMTP_HOST`: (required) - SMTP host address
+    - `SMTP_PORT`: (optional) - SMTP port. Defaults to `465`
+    - `SMTP_USERNAME`: (required) - SMTP username
+    - `SMTP_PASSWORD`: (required) - SMTP password
+    - `SMTP_EMAIL`: (required) - The outgoing address which is displayed to the user
+- `RECEIVERS`: (required) - a json-formatted object with `email` as key and `Event Subscription Ids` as an array value, eg `RECEIVERS="{\"email@example.com\": [\"EVENT_1\", \"EVENT_2\"]}"`. The email will then become subscribed to the listed events
+
+Note: env variables are accessible via the `secret` command under `auction-ui/notification-service/${environment}`.
```
valiafetisov, line 39: The path to the secrets should be `auction-ui/${host}/notification-service`. 

But I couldn't find any secrets neither under `auction-ui/notification-service/localhost` or under `auction-ui/notification-service/dev`. Where are they currently stored?

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+    contract: '0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F',
+    eventName: 'UpdateVersion',
+    formatData: event => {
+      return `Key: ${event.args.key}<br /> Address: ${event.args.addr}`;
+    },
+  }
+]
+```
+
+- `id` - Unique identifier. Used for references in email and console logs
```
valiafetisov, line 59: Can we use `contract:eventName` as unique identifier in logs? Or is it actually used for some other purposes?

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `contract` - The Ethereum Address on which your Event is
```
valiafetisov, line 60: I would call it `address`
```suggestion
- `address` - The Ethereum address which expected to emit the event
```

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `eventName` - The name of the event you want to observe
```
valiafetisov, line 61: If you support wildcard, you should specify it here
```suggestion
- `eventName` - The name of the event you want to observe or `*`
```

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+- `formatData` - A function that takes the event output, can reformat and then output it in html. It will be displayed in the email.
```
valiafetisov, line 62:
```suggestion
- `formatData` - A function that takes the emitted event data and have to return an html string that will be displayed in the email.
```

notification-service/src/constants/SUBSCRIPTIONS.ts
```diff
@@ -0,0 +1,52 @@
+import axios from 'axios';
+import { EventSubscription } from '../types';
+import { ETHERSCAN_API_KEY, ETHEREUM_NETWORK } from '../variables';
+import { getEtherscanAPI, getEtherscanURL } from './NETWORKS';
+import { ethers } from 'ethers';
+
+export async function getAbiFromContractAddress(contract: string) {
```
valiafetisov, line 7: I suggest that you move this function to another `etherscan` file, since this one suppose to contain only configuration

notification-service/src/constants/SUBSCRIPTIONS.ts
```diff
@@ -0,0 +1,52 @@
...
+        contract: '0xdA0Ab1e0017DEbCd72Be8599041a2aa3bA7e740F',
+        eventName: 'UpdateAddress',
+        formatData: event => {
+            return `
+                Key: ${ethers.utils.formatBytes32String(event.args.key)}<br /> 
+                Address: <a href="${getEtherscanURL(ETHEREUM_NETWORK)}/address/${
+                event.args.addr
+            }" target="_blank" style="color: lightblue;">${event.args.addr}</a>
+            `;
+        },
```
valiafetisov, line 31: If you're using `getEtherscanURL` in every event, you can nicely provide it as a helper via function argument (with preset NETWORK) `getEtherscanURL(event.args.addr, 'address')`

```suggestion
        formatData: ({ event, getEtherscanURL }) => {
            const contractName = ethers.utils.formatBytes32String(event.args.key);
            const url = getEtherscanURL(event.args.addr, 'address')
            return `
                Contract: ${contractName}<br /> 
                Address: <a href="${url}" target="_blank" style="color: lightblue;">${event.args.addr}</a>
            `;
        },
```

notification-service/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ETHEREUM_NETWORK } from './variables';
+import { setupReceiverList } from './recievers';
+
+const start = async function () {
+    isNetworkSupported(ETHEREUM_NETWORK);
+    setupReceiverList();
+    const transporter = await setupMailer();
+    const wsProvider = setupWebSocket();
+
+    subscribe(wsProvider, mailData => sendMail(transporter, ETHEREUM_NETWORK, mailData));
```
valiafetisov, line 14: Why does subscriber already emits `mailData`? I though it would emit event + maybe a subscription id, which then will be transformed into email inside `sendMail` or some other third `generateEmail` function?

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+to get started with your development setup. Namely, you should:
+
+- have a minimum `node` and `npm` version (see `package.json` `engine` section)
+- have a certain formatting and linting setup
+- don't forget to create `./notification-service/.env` file
+
+Help on both things is given in the linked resources above.
+
+In addition to this please be aware of the following for this project:
+- By default, this project uses [Etheral Email](https://ethereal.email/) to mock emails. If Etheral is down and no custom SMTP server is set, the service will not run.
```
valiafetisov, line 76: What means by default? If no `SMTP_` variables are provided? Then why not add this note to the env variable description itself?

notification-service/README.md
```diff
@@ -0,0 +1,77 @@
...
+- We advise using [MailTrap](https://mailtrap.io/) for more extensive testing.
```
valiafetisov, line 77: I don't think we need to advertise any specific email service in our readme

notification-service/src/constants/SUBSCRIPTIONS.ts
```diff
@@ -0,0 +1,52 @@
+import axios from 'axios';
+import { EventSubscription } from '../types';
+import { ETHERSCAN_API_KEY, ETHEREUM_NETWORK } from '../variables';
+import { getEtherscanAPI, getEtherscanURL } from './NETWORKS';
+import { ethers } from 'ethers';
+
+export async function getAbiFromContractAddress(contract: string) {
+    if (!ETHERSCAN_API_KEY) {
+        throw new Error('subscriptions: please set a valid ETHERSCAN_API_KEY, in the environment files');
+    }
```
valiafetisov, line 10: Does it mean, it will only fail at runtime if no `ETHERSCAN_API_KEY` provided?

notification-service/README.md
benediktkaiser: I have not added any secrets yet, as we need to generate new API tokens for etherscan. Who should take care of this?

notification-service/README.md
benediktkaiser: It is primarily needed at the id that emails can subscribe to. But I will also use it in the logs!

notification-service/README.md
benediktkaiser: Using `*` does not work anymore, due to me changing the way the websockets are triggered. Therefore I removed this from the Readme.

context:
Before I listened to the event name, `*`, exists as an event name, but now I listen to all events be default and then compare their name (event.event) to the event listed in the Subscription. Therefore `*` would not work unless I support it again, which I don't want to do, as it could mess up with `formatData()` function, if events have different args.

notification-service/README.md
benediktkaiser: `nodemailer`, the package I use to send emails, comes with a testing SMTP server built in. 

See more here:
https://nodemailer.com/smtp/testing/

I thought it would be easy to include this, to do a quick test without having to setup your own SMTP server or MailTrap. However if you think it is not needed, I can remove the feature and have the service fail, if no valid SMTP server is provided.

notification-service/src/constants/SUBSCRIPTIONS.ts
benediktkaiser: I changed this, to verify it before hand!

notification-service/README.md
LukSteib: I took that one over and created an API key for us. You can find credentials in bitwarden under "SAS Etherscan Account"

benediktkaiser:
## Added new Discord notifier
Users can now be notified via Discord Web Hook. 

Screenshot 2022-05-05 at 15 40 15

In order to allow more notifiers I have also changed the `formatData()` function into one returning markdown rather then html. For the emails the markdown will now be parsed into html. 

## Emails
Emails now look like the following:

Screenshot 2022-05-05 at 15 41 50

I have also updated the ReadMe to inform on how to use the notifiers. It is now also very easy to add new ones, if we ever decide to add telegram, whatsapp or some other method.

notification-service/README.md
benediktkaiser: Not relevant anymore as I removed this feature!

.drone.yml
```diff
@@ -236,3 +253,14 @@ data:
...
         event:
           - pull_request
+    - name: notification-service-lint
+      type: kubernetes
+      depends_on: [publish-notification-service-on-pr-push]
+      image: ghcr.io/sidestream-tech/unified-auctions-ui/notification-service:pr-${DRONE_PULL_REQUEST}
+      workdir: /notification-service
+      steps_parallel:
+        - name: lint
+          command: npm run lint
```
valiafetisov, line 263: Do you also want to build it to validate types?

notification-service/README.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Adding a new "Receiver"
+You can set your receivers over the environment variable `RECEIVERS`. This is a json-formatted object. An example can be found below:
+
+```json
+{
+  "test@example.com": {
+    "type": "email",
+    "subscriptions": ["MCD_DAI_Transfer", "ChainLogUpdateAddress"]
+  },
```
valiafetisov, line 49: Suggested format was an _array_ of objects. Why? Because eventually this data will be stored in a database instead of the env variable

notification-service/src/etherscan.ts
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function getEtherscanAPI(network: string) {
+    return NETWORKS[network].etherscanAPI;
+}
+
+export function validateEtherscanAPIKey() {
+    if (!ETHERSCAN_API_KEY) {
+        throw new Error(``);
+    }
```
valiafetisov, line 16: Please provide readable error text

notification-service/src/websocket.ts
```diff
@@ -0,0 +1,50 @@
...
+    SUBSCRIPTIONS.forEach(eventSubscription => {
+        if (!contracts[eventSubscription.address]) {
+            contracts[eventSubscription.address] = [eventSubscription];
+        } else {
+            contracts[eventSubscription.address].push(eventSubscription);
+        }
+    });
+
+    Object.keys(contracts).forEach(async contractAddress => {
+        const contractAbi = await getAbiFromContractAddress(contractAddress);
```
valiafetisov, line 36: This function may fail because 
- network error
- no contract has no abi
- etc

Leaving the whole subscription service stale, right?

notification-service/tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+    "declaration": true,
+    "sourceMap": true,
+    "rootDir": "./src",
+    "strict": true,
+    "noImplicitReturns": true,
+    "noFallthroughCasesInSwitch": true,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "moduleResolution": "node",
+    "jsx": "react",
```
valiafetisov, line 17: That's some leftovers, should we better copy our existing ts config from another service?

notification-service/src/websocket.ts
```diff
@@ -0,0 +1,50 @@
...
+        const contract = new ethers.Contract(contractAddress, contractAbi, wsProvider);
+
+        console.info(
+            `${WEBSOCKET_PREFIX} registered listener for ${contracts[contractAddress].length} event(s) on contract "${contractAddress}"`
```
valiafetisov, line 40: As suggested before, please use this message format:
https://github.com/sidestream-tech/unified-auctions-ui/pull/240#issuecomment-1117367683 (or disagree with it 🙂)

notification-service/src/constants/SUBSCRIPTIONS.ts
```diff
@@ -0,0 +1,27 @@
...
+                    > Address: [${event.args.addr}](${formatEtherscanLink('address', event.args.addr)})`;
+        },
+    },
+    {
+        id: 'MCD_DAI_Transfer',
+        address: '0x4f96fe3b7a6cf9725f59d353f723c1bdb64ca6aa',
+        eventName: 'Transfer',
+        formatData: (event, formatEtherscanLink) => {
+            return `> From: [${event.args.src}](${formatEtherscanLink('address', event.args.src)})<br />
+                    > To: [${event.args.dst}](${formatEtherscanLink('address', event.args.src)})`;
```
valiafetisov, line 20: Invalid link
```suggestion
                    > To: [${event.args.dst}](${formatEtherscanLink('address', event.args.dst)})`;
```

notification-service/src/generators/generateDiscordMessage.ts
```diff
@@ -0,0 +1,44 @@
...
+    eventSubscription: EventSubscription,
+    event: any,
+    formattedData: string
+) {
+    return {
+        embeds: [
+            {
+                author: {
+                    name: 'Unified Auctions Service',
+                    url: 'https://github.com/sidestream-tech/unified-auctions-ui',
```
valiafetisov, line 15: I would also never hardcode those kind of values, but at least provide them as default via env variables

notification-service/src/notifiers.ts
```diff
@@ -0,0 +1,31 @@
...
+    const receivers = getReceiversForSubscriptionId(eventData.eventSubscription.id);
+
+    // Notify per mail
+    notifyPerMail(notifiers.mailer, eventData, getReceiversByType(receivers, 'email')).catch(error =>
+        console.error(error)
+    );
+    // Notify per discord
+    notifyPerDiscord(eventData, getReceiversByType(receivers, 'discord')).catch(error => {
+        console.error(error);
+    });
```
valiafetisov, line 30: What happens if there is no receivers of this type?

notification-service/README.md
benediktkaiser: Done! ✅ Also changed in the bot readme

.drone.yml
benediktkaiser: Does it not get built in this step? 

https://github.com/sidestream-tech/unified-auctions-ui/blob/6e76a858f1ddecb393a12b130d24d40fae9ed556/.drone.yml#L158

notification-service/README.md
benediktkaiser: I had actually wanted to do this but for some reason it was ingrained into my brain that the issue stated the other. Happy to change it tho!

notification-service/package.json
```diff
@@ -0,0 +1,57 @@
...
+  "description": "An automated blockchain subscription service.",
+  "module": "dist/notification-service.esm.js",
+  "main": "dist/index.js",
+  "typings": "dist/index.d.ts",
+  "files": [
+    "dist",
+    "src"
+  ],
+  "scripts": {
+    "dev": "tsdx watch",
```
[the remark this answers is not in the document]
benediktkaiser: It sadly only live complies. Therefore it is quite unnecessary in my opinion. 

I suggest the following:
Replace `tsdx watch` with `npm run build && node .`. This would rebuild the service and start it. I found that using this felt a lot more useful, as I did not need to run two commands. 

In addition to this: When developing the `bot`, I often found myself having to restart the entire bot during development anyway, as hot reloading works poorly with bots or these kinds of services.

notification-service/src/websocket.ts
benediktkaiser: I changed it to display the following message if the retrieving fails. I also display the Error of the function, so the user can see if it was because of one of your listed reasons.

Screenshot 2022-05-06 at 00 07 47

notification-service/src/websocket.ts
```diff
@@ -0,0 +1,50 @@
...
+export function setupWebSocket(): ethers.providers.WebSocketProvider {
+    if (!INFURA_PROJECT_ID) {
+        throw new Error(`${WEBSOCKET_PREFIX} please set a valid INFURA_PROJECT_ID, in the environment files`);
+    }
+    const wsProvider = new ethers.providers.WebSocketProvider(
+        `wss://${ETHEREUM_NETWORK}.infura.io/ws/v3/${INFURA_PROJECT_ID}`,
+        ETHEREUM_NETWORK
+    );
+
+    console.info(`${WEBSOCKET_PREFIX} websocket connection opened with network "${ETHEREUM_NETWORK}"`);
```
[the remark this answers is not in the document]
benediktkaiser: `.ready` does not appear to throw errors:

>  ignoring errors due to the target node not being active yet.
https://docs.ethers.io/v5/api/providers/provider/#Provider-ready

Therefore I decided to call:

```
await wsProvider.getNetwork();
```

which seems to do the trick 🙌

notification-service/src/websocket.ts
benediktkaiser: Sorry, completely overlooked it!

notification-service/src/generators/generateDiscordMessage.ts
benediktkaiser: Updated it!

notification-service/src/notifiers.ts
benediktkaiser: It now returns!

notification-service/package.json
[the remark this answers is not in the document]
valiafetisov: Well, according to the docs, you can provide `--onSuccess` command to run node after the compilation: https://github.com/jaredpalmer/tsdx#tsdx-watch

LukSteib (dismissed):
Was able to succesfully test notification for MCD Transfer event and received notification via both channels mail and discord

.drone.yml
valiafetisov: Good point, I guess it does, I hope it also results in CI error if it fails though

valiafetisov:
- We shouldn't open connection to RPC or subscribe to anything before we validate all env variables (eg mailer setup), [currently, according to the logs, it's done later on](https://user-images.githubusercontent.com/3393626/168275522-a2576486-0a2c-436e-8910-72609db44b48.png)
- We shouldn't subscribe to events if there is no notifiers for them, currently we do
- I received 403 error from etherscan API on the first run, but the app didn't crash / retried – this is ok for development, but if this happens on start in the cloud environment we would never learn about it
- For some reason we generate different text for different notifications (even different for "text-only email" vs "html email") we of course should have a central place to define it, this needs refactoring
- Please always store env variables used for testing in secrets, enforce their creation if needed @benediktkaiser

valiafetisov (approved):

PR #243, opened 2022-05-03
Added Simulation related UI

### LINKED ISSUE #241: Implement simulation-related UI
### Goal

Several components that is shown within simulation popup in storybook

### Context

In the https://github.com/sidestream-tech/unified-auctions-ui/issues/83#issuecomment-1064049565 we outlined how do we want the final simulation flow to look like. This issue is implementation of the UI needed for that:

- `simulation/SimulationForm`
  - props
    - `token`
    - `simulationId`
    - `isLoading`
    - `disabled`
  - displays
    - inputs for `token` and `simulationId` (with default values from props)
    - `isLoading` and `disabled` disables inputs changes state and text of the button (see mock)
    - when both values exists, the button becomes primary (if not disabled)
  - emits
    - `token`, `simulationId`
- `simulation/RpcForm`
  - same as above, but with `url` as a main prop/input
- `simulation/SimulationLogs`
  - props
    - `logs` – ordered array of strings
  - displays
    - title shows first log message (`log[0]`)
    - collapsed by default
    - uncollapses into a scrollable monospaced block on click (see how it looks [here](https://user-images.githubusercontent.com/3393626/166437413-c22c0180-7aa7-409e-8801-109ba4ff1c24.png), please copy it [from here](https://github.com/sidestream-tech/unified-auctions-ui/blob/f7c1dda8fd1750c50c2e6e3cf323673a79b221b7/frontend/containers/ChaoslabsContainer.vue#L15-L18))
    - collapses back on click to the title

Notes: 
 - all components should be responsive
 - form component should have `form` as root element, use `@submit.prevent`
 - some parts was implemented in the https://github.com/sidestream-tech/unified-auctions-ui/pull/89/files (before the mock was created, so please check it out before you start)

### Assets
The mock: https://github.com/sidestream-tech/unified-auctions-ui/issues/83#issuecomment-1064049565

### Tasks
- [ ] create 3 components
- [ ] create stories for each the states (isLoading / disabled), handle emits via actions

fegbert:
closes #241 

**Rpc Form**

**Simulation Form**

**Simulation Logs**


frontend/components/simulation/SimulationForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        },
+    },
+    data() {
+        return {
+            inputToken: this.token,
+            selectedSimulationId: this.simulationIds.split(',')[0],
+        };
+    },
+    computed: {
+        simulations(): Object {
```
valiafetisov, line 76: As far as I can see, `simulations` returns an array of [`SelectOption`s](https://github.com/sidestream-tech/unified-auctions-ui/blob/eed9564bc4b4d64afbe241f45cbcb690b55639ba/frontend/shims.d.ts#L25-L29). Please avoid using `Object` contractor as type for this exact reasons: in js arrays are also objects, as well as functions, etc

frontend/components/simulation/SimulationForm.vue
```diff
@@ -0,0 +1,101 @@
...
+                .split(',')
+                .map(id => ({
+                    label: id.trim(),
+                    value: id.trim(),
+                }))
+                .filter(simulation => simulation.label !== '');
+        },
+        noSimulationsAvailable(): boolean {
+            return Object.keys(this.simulations).length < 1;
+        },
```
valiafetisov, line 87: No need for keys, since it's an array
```suggestion
        noSimulationsAvailable(): boolean {
            return this.simulations.length === 0;
        },
```

frontend/components/simulation/SimulationForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        hasProperEnvVariables(): boolean {
+            return !!this.selectedSimulationId && !!this.inputToken;
+        },
```
valiafetisov, line 90:
```suggestion
        hasValidValues(): boolean {
```
+ you may want to trim them to be sure (here and during emit)

frontend/components/simulation/SimulationLogs.stories.js
```diff
@@ -0,0 +1,34 @@
+import { storiesOf } from '@storybook/vue';
+import faker from 'faker';
+import SimulationLogs from './SimulationLogs';
+
+const randomLogs = length => {
+    return faker.datatype.array(length).map(() => {
+        return `${faker.date.recent().toISOString()}: ${faker.random.words(3)}`;
```
valiafetisov, line 7: Can you please use `sentence` instead, to see how it handles long, multi-line log messages?

fegbert:
Mobile resolution to further show multi-line log messages:

image

frontend/components/simulation/SimulationLogs.stories.js
fegbert: Screenshot below 🙂

valiafetisov (dismissed):

LukSteib (changes requested):
- [ ] NIT: Pre-Input URL story for `RpcURL` should fill input with actual "http**s**:." address
- [ ] I would expect the `Connect` button of `RpcURL` to be disabled as long as no there is no input (similar to default state of the `SimulationForm`). Currently it's a clickable secondary  button
- [ ] I would expect the `Reset` button for the `SimulationForm` to be a secondary button (similar to the disconnect button of the `RpcURL`). Currently it is a primary type button

LukSteib (approved):

frontend/components/simulation/RpcForm.vue
```diff
@@ -0,0 +1,70 @@
...
+        return {
+            inputUrl: this.url || '',
+        };
+    },
+    methods: {
+        submitForm() {
+            if (!this.inputUrl.startsWith('https://')) {
+                message.error('RPC url should start with https://');
+                return;
+            }
```
valiafetisov, line 65: That's actually not a requirement, since someone can ethereum node locally

frontend/components/simulation/SimulationForm.vue
```diff
@@ -0,0 +1,101 @@
...
+            </label>
+            <label class="w-full md:w-1/2">
+                <TextBlock class="font-semibold text-xs">Simulation id</TextBlock>
+                <Select
+                    :value="selectedSimulationId"
+                    :disabled="isLoading || disabled || noSimulationsAvailable || isStarted"
+                    :loading="isLoading"
+                    :options="simulations"
+                    @change="selectedSimulationId = $event"
+                />
```
valiafetisov, line 16: Does it also let users type their own text or only allow to select pre-defined options? (we need to allow free-text input as well)

valiafetisov: Also `noSimulationsAvailable` should not disable the input

frontend/components/simulation/RpcForm.vue
```diff
@@ -0,0 +1,70 @@
...
+            this.$emit('submit', this.inputUrl);
```
valiafetisov, line 66: This is not yet trimmed, like requested here https://github.com/sidestream-tech/unified-auctions-ui/pull/243#discussion_r863927543

frontend/components/simulation/SimulationForm.vue
```diff
@@ -0,0 +1,101 @@
...
+        },
+        hasValidVariables(): boolean {
+            return !!this.selectedSimulationId && !!this.inputToken.trim();
+        },
+    },
+    methods: {
+        submitForm() {
+            this.$emit('submit', {
+                token: this.inputToken.trim(),
+                simulationId: this.selectedSimulationId,
```
valiafetisov, line 96: This should also be trimmed (as soon as it can be a free-text input)

valiafetisov (approved):

PR #244, opened 2022-05-04
Fix max gas

### LINKED ISSUE #213: Investigate gas prices
**Goal:**
Ensure that we properly use and set gas prices

**Context:**
During the last release we have reworked our gas price determination. 
Idea is to fetch fast gas price from chainlink oracle and set this one as `Max Fee Per Gas`. 
However while looking at the dashboard with different gas prices for `mainnet` (see screenshot in assets section below), I wondered:
If `base fee per gas` is way higher than `Max Fee Per Gas` that we set, wouldn't transactions fail again? I am hesitant and wanted to raise again whether the chainlink -> `Max Fee Per Gas` approach is correct. 

**Tasks:**
- [ ] Reconsider approach on how we determine the `Max Fee Per Gas` 

**Assets:**
image

valiafetisov:
Closes #213 

- `Max Fee Per Gas` is now `Base Fee Per Gas` * 2 + `Max Priority Fee Per Gas`
- In the  `SwapTransactionTable` we now display `Combined Transactions Fees`
  - with the value of `Base Fee Per Gas` + `Max Priority Fee Per Gas`, not `Max Fee Per Gas` as before

Screenshot 2022-05-04 at 11 11 01


LukSteib:
Looking good to me

Two small questions:
* Do we get the `base fee per gas` directly from the chain? So basically fetching the `base fee per gas` of the previous block? 
* Does the `auctionTransaction.totalFee` include auth fees if wallet / collateral haven't been authorised yet?

benediktkaiser (approved):

valiafetisov:
> So basically fetching the `base fee per gas` of the previous block?

Yes

> Does the `auctionTransaction.totalFee` include auth fees if wallet / collateral haven't been authorised yet?

Yes, `totalFee` includes all required fees for the swap transaction based on the connected wallet. When no wallet is connected, it assumes no authorizations were executed.

LukSteib (approved):

PR #249, opened 2022-05-05
Updated step to profitability calculation

### LINKED ISSUE #221: Improve profitability time calculation
### Goal

Profitability estimation use limited number of resources

### Context

Time till profitability calculation was initially implemented using a loop. That means the further the profit date, the more cycles the loop will take to compute it. Moreover, the loop is not limited, so if it's never going to be profitable, the number of loops can be quite big. Therefore, we want to reimplement this part using pure mathematical operations instead of the loop.

### Assets
Part of the code that needs to be rewritten:
https://github.com/sidestream-tech/unified-auctions-ui/blob/f2e8f318e20c28a2d469c90139ad67c0f8af2ab6/core/src/price.ts#L116-L119

### Tasks
- [x] Outline a formula here
- [x] Find and compare libraries that can help with computing it (popularity, support, package size, possibility to use/import only single function, precision, etc)
- [x] Propose a library candidate, get confirmation
- [x] Reimplement computation

benediktkaiser:
Closes #221

I managed to implement the new algorithm. After implementing it, I ran both calculation methods at the same time and made sure that all the results lined up:

Screenshot 2022-05-06 at 00 59 25

Afterwards, I ran the chrome performance tool on both algorithms to determine which one is more performant. The results are shown below:

**Old Algorithm**:
Screenshot 2022-05-06 at 01 08 44

**New Algorithm**:
Screenshot 2022-05-06 at 01 08 38

From what I can see the new algorithm and therefore this PR, perform slightly better than the old looping version.

----

Checklist:

### DIFF
--- a/core/src/price.ts
+++ b/core/src/price.ts
@@ -110,14 +110,12 @@ export const calculateTransactionGrossProfitDate = function (auction: Auction, c
         return undefined;
     }
 
-    let steps = 0;
-    let currentValue = new BigNumber(auction.approximateUnitPrice);
+    const stepNumber = Math.ceil(
+        Math.log(auction.marketUnitPrice.dividedBy(auction.approximateUnitPrice).toNumber()) /
+            Math.log(auction.priceDropRatio.toNumber())
+    );
 
-    while (currentValue.isGreaterThan(auction.marketUnitPrice)) {
-        steps += 1;
-        currentValue = currentValue.multipliedBy(auction.priceDropRatio);
-    }
     const secondsSinceLastPriceDrop = auction.secondsBetweenPriceDrops - auction.secondsTillNextPriceDrop;
-    const secondsTillProfitable = auction.secondsBetweenPriceDrops * steps - secondsSinceLastPriceDrop;
+    const secondsTillProfitable = auction.secondsBetweenPriceDrops * stepNumber - secondsSinceLastPriceDrop;
     return addSeconds(currentDate, secondsTillProfitable);
 };


valiafetisov:
> As far as I am aware this value would not change once the Auction started, or am I incorrect here?

It can change if the `market price`, `step` or `cut` has changed. If you would have a function that accepts those things as a parameter, then you can cache its output, right?

benediktkaiser:
> > As far as I am aware this value would not change once the Auction started, or am I incorrect here?
> 
> It can change if the `market price`, `step` or `cut` has changed. If you would have a function that accepts those things as a parameter, then you can cache its output, right?

I added a cache of 30 minutes. Do you think this is acceptable or should it be longer/shorter?

core/src/price.ts
```diff
@@ -110,14 +113,18 @@ export const calculateTransactionGrossProfitDate = function (auction: Auction, c
...
-    const secondsTillProfitable = auction.secondsBetweenPriceDrops * steps - secondsSinceLastPriceDrop;
+    const secondsTillProfitable = auction.secondsBetweenPriceDrops * stepNumber - secondsSinceLastPriceDrop;
     return addSeconds(currentDate, secondsTillProfitable);
 };
+
+export const calculateTransactionGrossProfitDate = memoizee(_calculateTransactionGrossProfitDate, {
+    maxAge: TRANSACTION_GROSS_PROFIT_DATE_CACHE_MS,
+    promise: true,
+    length: 2,
+});
```
valiafetisov, line 130: Since this function receives auction object which comes from the store, the auction object can potentially be:
- different object although the values are the same (especially since BigNumber values are objects themselves, recreated with `new` during the enrichment)
- or the same between the calls, but with different values

To avoid this problem of comparing objects by their reference, you need to serialise it:
https://github.com/medikoo/memoizee#cache-id-resolution-normalization

Then there will be no need for the `maxAge` since as soon as the values haven't changed, the profit time will be the same as well

core/src/price.ts
```diff
@@ -92,7 +95,7 @@ export const calculateTransactionCollateralOutcome = function (
     return potentialOutcomeCollateralAmount;
 };
 
-export const calculateTransactionGrossProfitDate = function (auction: Auction, currentDate: Date): Date | undefined {
+const _calculateTransactionGrossProfitDate = function (auction: Auction, currentDate: Date): Date | undefined {
```
valiafetisov, line 98: Oh, but you have a `currentDate` as a parameter! That means cache wouldn't make any sense, right?

core/src/price.ts
```diff
@@ -110,14 +113,18 @@ export const calculateTransactionGrossProfitDate = function (auction: Auction, c
         return undefined;
     }
 
-    let steps = 0;
-    let currentValue = new BigNumber(auction.approximateUnitPrice);
+    const stepNumber = Math.ceil(
+        Math.log(auction.marketUnitPrice.dividedBy(auction.approximateUnitPrice).toNumber()) /
+            Math.log(auction.priceDropRatio.toNumber())
+    );
```
valiafetisov, line 119: Do you think it's possible to use normal `Math` operations over those big numbers? Have you put any thought into it? (I can imagine it may be fine as soon as we fine with the approximate number here, but I haven't though about edge cases, have you?)

core/src/price.ts
valiafetisov: One way around this will be to split this function into two: one that calculates some static numbers based on the `initial price`, `step`, `cut` and `market price` and the other one which applies it. Then you can cache the first one

benediktkaiser: So I am having some issues dealing with this refractor. I currently have the following:

- Determine the initial Price per collaterals (take `initialPrice` and divide it by `collateralAmount`)
- Determine the steps needed to reach profitability based on the marketPrice
- Multiple the amount of steps by the amount of seconds between each step
- Add the number of seconds to the auction start time

Then this value would be cached unless: 
- Step cut/time change (which they should not)
- Market price is adjusted (can happen)

My issue is at the following part. We currently calculate the profitability time with the following function (at least in this PR):
```
const stepNumber = Math.ceil(
  Math.log(auction.marketUnitPrice.dividedBy(auction.approximateUnitPrice).toNumber()) /
  Math.log(auction.priceDropRatio.toNumber())
);
```

In order to cache the value correctly I replaced `auction.approximateUnitPrice` with the initial price per collateral determine before. Afterwards I only needed to change the following part:

```
return addSeconds(currentDate, secondsTillProfitable);
```

Where I changed `currentDate` to `auction.startTime`.

However after doing this I got some very weird numbers from the function. I assume one of the following is happening:
- `initialPrice` is somehow incorrect and therefore the base price per collateral is wrong
- I need to refractor the step count logic (I wouldn't know what tho)


@valiafetisov, maybe you could be so kind to give me some pointers, into which of these issue, it most likely is. Thanks in advance! 😄

valiafetisov: Ok, let's skip the cache from this PR and do it as a separate issue! Please open it, linking our discussion

benediktkaiser: I created a new issue under #268

valiafetisov (approved):

core/src/price.ts
valiafetisov: Another question is: can any of the numbers be 0? At `Math.log(0)` will return `-Infinity`

LukSteib (approved):

PR #251, opened 2022-05-06
Do not refetch all auctions, when one is opened

### LINKED ISSUE #223: Do not refetch all auctions when only one is opened
### Goal

Improve performance of the UI when user wants to participate in a single auction

### Context

Currently we're refetching all auctions every 30 seconds:
https://github.com/sidestream-tech/unified-auctions-ui/blob/f2e8f318e20c28a2d469c90139ad67c0f8af2ab6/frontend/store/auctions.ts#L19

And although it's not a simple interval, but 30 seconds between the end of the previous fetching and the next one (so that if the fetching itself takes more than 30 seconds, the UI is not overloaded), it's still too much in case there are a lot of active actions (eg > 100), which can compromise performance of the page itself, given all other enrichments we apply.

Instead, we should refetch only a single auction when the auction is "opened". This will also require us to better indicate what's happening, with both this auction and the auction table. Some initial ideas:
- Add indicator on the main page: above or below the table with auctions, when the auctions were fetched last time (and maybe even count down on when they will be refetched?)
- Move auctions table loading indicator from over the table to above/below around the last time fetch date (so it can also show refetching without disturbing the table itself)
- Darken the overlay or in other ways convey that auctions are not refetching
- Add indicator to the single auction page about the fetching date

### Implementation proposal
Implementation-wise this can be implemented completely in the store using `rootState.route.query` reactive state available in every store. The action `fetchWithoutLoading` can just check if `rootState.route.query.auction` is present and fetch only a single auction instead using something like `fetchAuctionByCollateralTypeAndAuctionIndex` + enrich it. We can also

### Tasks
- [x] create a mock
- [x] discuss, finalise
- [x] fetch only a single auction
- [x] implement UI improvements

benediktkaiser:
Closes #223.

I have changed the auction store to act with the following behaviour:
- Collateral Page Loads in, no selected auction -> All Auctions are fetched
  - Auction is selected. While Auction is open, only it will be refetched every 30 seconds
  - On closing of the Auction Panel, the `update` method is called, updating all auctions, to ensure accurate data
- Collateral Page Loads in, auction is selected -> Only the selected Auction loads!
  - On closing of the Auction Panel, the `update` method is called, updating all auctions, to ensure accurate data

Checklist:

### DIFF
--- a/frontend/containers/AuctionsContainer.vue
+++ b/frontend/containers/AuctionsContainer.vue
@@ -80,6 +80,7 @@ export default Vue.extend({
                 if (!newAuctionId) {
                     const network = this.$route.query.network;
                     this.$router.push({ query: { network } });
+                    this.update();
                 }
             },
         },
@@ -125,7 +126,7 @@ export default Vue.extend({
             'fetchWalletAuthorizationStatus',
             'fetchCollateralAuthorizationStatus',
         ]),
-        ...mapActions('auctions', ['bidWithCallee', 'bidWithDai', 'restart', 'fetchTakeEventsByAuctionId']),
+        ...mapActions('auctions', ['bidWithCallee', 'bidWithDai', 'restart', 'fetchTakeEventsByAuctionId', 'update']),
         ...mapActions('wallet', ['fetchWalletBalances', 'fetchCollateralVatBalance', 'withdrawAllCollateralFromVat']),
         openSelectWalletModal(): void {
             if (!this.hasAcceptedTerms) {
--- a/frontend/containers/AuctionsContainer.vue
+++ b/frontend/containers/AuctionsContainer.vue
@@ -80,7 +80,7 @@ export default Vue.extend({
                 if (!newAuctionId) {
                     const network = this.$route.query.network;
                     this.$router.push({ query: { network } });
-                    this.update();
+                    this.update(true);
                 }
             },
         },
--- a/frontend/containers/AuctionsContainer.vue
+++ b/frontend/containers/AuctionsContainer.vue
@@ -6,6 +6,7 @@
             :is-selected-auction-fetching="isSelectedAuctionFetching"
             :are-take-events-fetching="areTakeEventsFetching"
             :auctions-error="auctionsError"
+            :auction-errors="auctionErrors"
             :selected-auction-id.sync="selectedAuctionId"
             :is-explanations-shown.sync="isExplanationsShown"
             :wallet-address="walletAddress"
--- a/frontend/containers/AuctionsContainer.vue
+++ b/frontend/containers/AuctionsContainer.vue
@@ -128,7 +128,14 @@ export default Vue.extend({
             'fetchWalletAuthorizationStatus',
             'fetchCollateralAuthorizationStatus',
         ]),
-        ...mapActions('auctions', ['bidWithCallee', 'bidWithDai', 'restart', 'fetchTakeEventsByAuctionId', 'update', 'updateSingleAuction']),
+        ...mapActions('auctions', [
+            'bidWithCallee',
+            'bidWithDai',
+            'restart',
+            'fetchTakeEventsByAuctionId',
+            'update',
+            'updateSingleAuction',
+        ]),
         ...mapActions('wallet', ['fetchWalletBalances', 'fetchCollateralVatBalance', 'withdrawAllCollateralFromVat']),
         openSelectWalletModal(): void {
             if (!this.hasAcceptedTerms) {
--- a/frontend/containers/AuctionsContainer.vue
+++ b/frontend/containers/AuctionsContainer.vue
@@ -3,6 +3,7 @@
         <MainFlow
             :auctions="auctions"
             :are-auctions-fetching="areAuctionsFetching"
+            :is-selected-auction-fetching="isSelectedAuctionFetching"
             :are-take-events-fetching="areTakeEventsFetching"
             :auctions-error="auctionsError"
             :selected-auction-id.sync="selectedAuctionId"
@@ -49,6 +50,7 @@ export default Vue.extend({
             auctions: 'listAuctionTransactions',
             takeEvents: 'getTakeEventStorage',
             areAuctionsFetching: 'getAreAuctionsFetching',
+            isSelectedAuctionFetching: 'getIsSelectedAuctionFetching',
             areTakeEventsFetching: 'getAreTakeEventsFetching',
             isAuctionBidding: 'getIsBidding',
             auctionsError: 'getError',
@@ -80,7 +82,7 @@ export default Vue.extend({
                 if (!newAuctionId) {
                     const network = this.$route.query.network;
                     this.$router.push({ query: { network } });
-                    this.update(true);
+                    this.updateSingleAuction(newAuctionId);
                 }
             },
         },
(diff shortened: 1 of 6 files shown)


fegbert (dismissed):

frontend/store/auctions.ts
```diff
@@ -42,6 +46,7 @@ export const state = (): State => ({
     error: null,
     restartingAuctionsIds: [],
     lastUpdated: undefined,
+    selectedAuctionId: undefined,
```
valiafetisov, line 49: We shouldn't store this value, since it can be derived from the router

frontend/store/auctions.ts
```diff
@@ -147,25 +155,46 @@ export const mutations = {
...
+                commit('setError', error.message);
+            } finally {
+                commit('refreshLastUpdated');
+                commit('setAreAuctionsFetching', false);
+            }
+        } else {
+            commit('setAreAuctionsFetching', true);
+            try {
+                const auctionInfo = parseAuctionId(selectedAuctionId);
+                const auction = await fetchAuctionByCollateralTypeAndAuctionIndex(
```
valiafetisov, line 185: Can you please instead create `fetchAuctionByAuctionId` which does parsing of the id inside of it?

frontend/store/auctions.ts
```diff
@@ -147,25 +155,46 @@ export const mutations = {
...
+                    network,
+                    auctionInfo.collateralType,
+                    auctionInfo.index
+                );
+                const enrichedAuction = await enrichAuction(network, auction);
+                commit('setAuction', enrichedAuction);
+            } catch (error) {
+                console.error('fetch auction error', error);
+                commit('setError', error.message);
```
valiafetisov, line 194: This also sets error into the table, not auction page! Instead, we should store/display it in relation to the refetched auction

valiafetisov (commented):
Does this PR still fetch all auctions on page load? Because it shouldn't, in case user comes from the twitter, for example

benediktkaiser:
> - [ ] Does this PR still fetch all auctions on page load? Because it shouldn't, in case user comes from the twitter, for example

if you open the page via a link that selects an auction, the others will not be loaded until you exit the auction view.

frontend/store/auctions.ts
benediktkaiser: This was left over! My bad 😄

frontend/containers/AuctionsContainer.vue
```diff
@@ -80,6 +82,7 @@ export default Vue.extend({
                 if (!newAuctionId) {
                     const network = this.$route.query.network;
                     this.$router.push({ query: { network } });
+                    this.updateSingleAuction(newAuctionId);
```
valiafetisov, line 85: According to the `if` statement on line 82, `newAuctionId` is falsy, but you still call `updateSingleAuction` with it?

frontend/store/auctions.ts
```diff
@@ -165,9 +182,26 @@ export const actions = {
...
+        if (!network) {
+            return;
+        }
+        commit('setIsSelectedAuctionFetching', true);
+        try {
+            const auction = await fetchSingleAuctionById(network, auctionId);
+            commit('setAuction', auction);
+        } catch (error) {
+            console.error('fetch auction error', error);
+            commit('setError', error.message);
```
valiafetisov, line 200: [This comment](https://github.com/sidestream-tech/unified-auctions-ui/pull/251#discussion_r869107949) was not resolved: fetching a single auction will still set error over the whole table in the inactive panel (not even visible on mobile) while the auction page will display nothing. 

The error should be stored either to `selectedAuctionError` hack  or via `auction.error` hack or better via `auctionErrorsStore`. Which is then should be passed to the auction page alert

frontend/containers/AuctionsContainer.vue
benediktkaiser: Oops! Oversaw this.

frontend/store/auctions.ts
benediktkaiser: I made a new auctions error store, where we now push errors according to auctionId.

LukSteib (changes requested):
* when on a single auctions page only this auction is continuously fetched and the overall table not anymore ✔️  
* when directly entering a single auction via dedicated auction URL, only this auction is continuously fetched ✔️ 
- [x] Whenever switching back from single auction page to auction table page I would expect to immediately fetch auction data. This is currently not handled properly - in the bad case scenario a user has to wait for 29 sec in order to see that there are actually more auctions

---
Sidenote: I was not able to test the scenario of an error during fetching. Tried it via blocking network requests for infura URL but then we just do not fetch any longer without explicitly throwing an error.

benediktkaiser:
> * [x]  Whenever switching back from single auction page to auction table page I would expect to immediately fetch auction data. This is currently not handled properly - in the bad case scenario a user has to wait for 29 sec in order to see that there are actually more auctions

This works now! I called the wrong function before 🤪 

> Sidenote: I was not able to test the scenario of an error during fetching. Tried it via blocking network requests for infura URL but then we just do not fetch any longer without explicitly throwing an error.

Lukas and I tested this via Tandem by introducing an artificial error. When first loading the auction the page looks as follows:

Screenshot 2022-05-18 at 12 59 58

If the auction was previously fetched the user still sees the old data along with the error message. The buttons also get deactivated, to ensure the user cannot participate while seeing outdated information.

benediktkaiser:
> >  introducing an artificial error. When first loading the auction the page looks as follows
> 
> 
> 
> What is the error message that we are going to display in the error banner? 

the parsed error message coming from the fetching.

LukSteib (dismissed):

frontend/containers/AuctionsContainer.vue
```diff
@@ -125,7 +130,14 @@ export default Vue.extend({
...
             'fetchCollateralAuthorizationStatus',
         ]),
-        ...mapActions('auctions', ['bidWithCallee', 'bidWithDai', 'restart', 'fetchTakeEventsByAuctionId']),
+        ...mapActions('auctions', [
+            'bidWithCallee',
+            'bidWithDai',
+            'restart',
+            'fetchTakeEventsByAuctionId',
+            'updateAllAuctions',
+            'updateSingleAuction',
```
valiafetisov, line 139: `updateSingleAuction` seems to be unused here, right?

frontend/store/auctions.ts
```diff
@@ -26,9 +27,11 @@ interface State {
     auctionStorage: Record<string, AuctionTransaction>;
     takeEventStorage: Record<string, TakeEvent[]>;
     areAuctionsFetching: boolean;
+    isSelectedAuctionFetching: boolean;
     areTakeEventsFetching: boolean;
     isBidding: boolean;
     error: string | null;
+    auctionErrors: Record<string, string>;
```
valiafetisov, line 34: The type suppose to be `Record<string, string | undefined>` since some keys may be missing

frontend/store/auctions.ts
```diff
@@ -161,13 +186,31 @@ export const actions = {
...
+    async updateSingleAuction({ commit, rootGetters }: ActionContext<State, State>, auctionId: string) {
+        const network = rootGetters['network/getMakerNetwork'];
+        if (!network) {
+            return;
+        }
+        commit('setIsSelectedAuctionFetching', true);
+        try {
+            const auction = await fetchSingleAuctionById(network, auctionId);
+            commit('setAuction', auction);
+            commit('setErrorByAuctionId', { auctionId, error: null });
```
valiafetisov, line 206: We should use null only if we need to distinguish from "no value" state (here and everywhere in our code). Here it's more appropriate to store the same value as the default one, ie `undefined`

PR #257, opened 2022-05-10
Remove minimum bid limit

valiafetisov:
This pr doesn't close any specific issue, but tackles the minimum limit of the DAI the user can bid.

- Limit removed
- Error message changed
- `Set minimum bid` removed from the table

Screenshot 2022-05-10 at 15 54 20
Screenshot 2022-05-10 at 15 54 33
Screenshot 2022-05-10 at 15 49 44

### DIFF
--- a/frontend/components/utils/BidInput.vue
+++ b/frontend/components/utils/BidInput.vue
@@ -62,10 +62,8 @@ export default Vue.extend({
                 return;
             }
             const bidTopLimit = maxValue?.minus(minValue);
-            if (currentValue?.isGreaterThan(bidTopLimit) || currentValue?.isLessThan(minValue)) {
-                throw new Error(
-                    `The value can only be between ${minValue.toFixed(2)} and ${bidTopLimit.toFixed(2)} or the maximum`
-                );
+            if (currentValue?.isGreaterThan(bidTopLimit)) {
+                throw new Error(`The value can only be between 0 and ${bidTopLimit.toFixed(2)} or the maximum`);
             }
             if (maxValue?.isLessThan(minValue.multipliedBy(2))) {
                 throw new Error('The value can not be changed since the leftover part will be too small');
--- a/frontend/components/utils/BidInput.vue
+++ b/frontend/components/utils/BidInput.vue
@@ -63,9 +63,9 @@ export default Vue.extend({
             }
             const bidTopLimit = maxValue?.minus(minValue);
             if (currentValue?.isGreaterThan(bidTopLimit)) {
-                throw new Error(`The value can only be between 0 and ${bidTopLimit.toFixed(2)} or the maximum`);
+                throw new Error(`The value can only be less than ${bidTopLimit.toFixed(2)} or the maximum`);
             }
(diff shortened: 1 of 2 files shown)


LukSteib (changes requested):
* Was able to bid a small amount (100 Dai) on the ETH-C auction and everything worked fine (see recording below for the whole process)
- [ ] For the `ETH-B` auction I am not able to alter the input value and we are not able to display the tx outcome. What is the reason behind that? (see screeshot below) 
recording:
https://www.loom.com/share/ef872dbb5fa445ec940013b4d4ff894d


screenshot: 
image

valiafetisov:
> For the ETH-B auction I am not able to alter the input value and we are not able to display the tx outcome. What is the reason behind that? Also the tooltip text is kind of missleading if there is no minimum anymore (see screeshot below)
recording:

Good catch! Previously, the value that disabled the input was `debtDAI < minimumBidDai * 2`, I changed it to `debtDAI < minimumBidDai`. Now this should edge case should never happen, unless the chost was increased after the auction was created.

- We can reintroduce `minumum bid` (ie chost) as `minimum leftover` to the table for this logic to make sense, otherwise right now it only shown (in a way) after you try to input a little less than total

LukSteib:
Now the input is no longer disabled but there seems to be a problem with the computation of our `bidTopLimit` as it is indicated as 0.00 and prevents any input other than the maximum bid. Maximum value which is our `debtDai` seems to be correct though. 
Do you have an idea?


image

valiafetisov:
Now we're talking edge cases! This message is shown only when `debtDai` is equal to `minimumBidDai`. I now disable input for this case as well. Although if the difference is very small (eg `debtDai - minimumBidDai = 0.0001`) the message will still stay the same. We can maybe disable it input difference is smaller than 1 DAI?

frontend/components/utils/BidInput.vue
```diff
@@ -62,10 +62,8 @@ export default Vue.extend({
...
             const bidTopLimit = maxValue?.minus(minValue);
-            if (currentValue?.isGreaterThan(bidTopLimit) || currentValue?.isLessThan(minValue)) {
-                throw new Error(
-                    `The value can only be between ${minValue.toFixed(2)} and ${bidTopLimit.toFixed(2)} or the maximum`
-                );
+            if (currentValue?.isGreaterThan(bidTopLimit)) {
+                throw new Error(`The value can only be between 0 and ${bidTopLimit.toFixed(2)} or the maximum`);
             }
             if (maxValue?.isLessThan(minValue.multipliedBy(2))) {
                 throw new Error('The value can not be changed since the leftover part will be too small');
```
LukSteib, line 69: I assume this one needs adjustment as well since we also apply this multiplication by 2

Otherwise I am running into the problem indicated below:
image

image

valiafetisov:
> We can maybe disable it input difference is smaller than 1 DAI?

Instead I change the error message to be `The value can only be less than ${bidTopLimit.toFixed(2)} or the maximum`

LukSteib (approved):

benediktkaiser (approved):

PR #260, opened 2022-05-13
Count callee transactions

valiafetisov:
This PR is a quick and dirty implementation of a script to fetch information about finished auctions and export it into xlsx.

Note: not ready for review.


valiafetisov:
Won't merge until we have a clear system on where to store scripts and how to document them

PR #266, opened 2022-05-16
Add error parser

### LINKED ISSUE #259: Improve error message parsing
### Goal
Collect error message and ensure they are parsed correctly in the future 

### Context
We encountered that not all error messages are parsed in a user friendly way in our UI, sometimes resulting in notification `Error: unknown` 

Errors from the RPC come in unstructured form: text + json in one variable.

This issue should be used to collect a list of known errors and create a story to display error message + result parsed by the helper 

### Tasks
- [x] Collect known error message (see first comment) 
- [x] create a story for `./helpers/parseMetamaskError` 
   - [x] display error message + result parsed by helper

benediktkaiser:
Closes #259.

Checklist:

Todo:
- [ ] Add remaining error messages


core/src/constants/METAMASK_ERRORS.ts
```diff
@@ -0,0 +1,6 @@
...
...000000000000000000000000200000000000000000000000000000000000000000000000ff4327f3f09430c010000000000000000000000000000000000000004aaebc81e9fe122a2aab7215400000000000000000000000074893c37beacf205507ea794470b13de0629422000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000180000000000000000000000000b2e1284fb685dbdcc4378cf5d4e9515679527a42000000000000000000000000f11a98339fe1cde648e8d1463310ce3ccc3d7cc1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000000030000000000000000000000001f9840a85d5af5bf1d1762f925bdaddc4201f984000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f","type":2,"accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.4.5)':
```
valiafetisov, line 2: - [ ] Let's not store it as part of our core, but only as part of our frontend storybook file
- [ ] The error thrown [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/b0970c58b7258335f59403194821bbadd621339b/frontend/store/auctions.ts#L207) so the actual error message thrown by the metamask starts after `Error: ` – this is what you need to give as input to the function
- [ ] no need to store output here

core/src/helpers/parseMetamaskError.ts
```diff
@@ -6,9 +7,16 @@ const parseMetamaskError = function (errorMessage = ''): string {
     try {
         const jsonString = errorMessage.substring(jsonFirstIndex, jsonLastIndex + 1);
         const metamaskError = JSON.parse(jsonString);
-        return truncateText(metamaskError?.value?.data?.message || 'unknown');
+
+        const rawErrorMessage = metamaskError?.value?.data?.message || 'unknown';
+        const parsedErrorMessage = METAMASK_ERRORS[rawErrorMessage] || rawErrorMessage;
```
valiafetisov, line 12: This issue is about improving how we parse error messages! We don't have to store a list or exact strings that should display something else instead

valiafetisov: In other words, you need to change the logic above those lines, where we try to `JSON.parse` the message based on the hacked together `jsonFirstIndex, jsonLastIndex` which doesn't always work as it seems!

benediktkaiser:
Closing this and reopening a new PR!

PR #269, opened 2022-05-17
Enabled keeper pre-authorisation for wallet and collaterals

### LINKED ISSUE #264: Enable keeper pre-authorisation for wallet and collaterals
**Goal:**
Keeper implementation allows to pre-authorise wallet and collaterals

**Context:**
Currently the keeper is implemented in a way that once an active auction is fetched the keeper checks whether the two required authorisation steps (ie. wallet auth and collateral auth) need to be done in order to participate in the auction. 
There are two reasons that require us to change that behavior and enable the authorisation steps in advance of an active auction:
* speed of execution matters when there is a profitable auction
* authorisation tx can be timed in order to use low gas fees

With this issue we want to introduce the capability to do authorisations without the need for an active auction.

_Proposal:_
* Introduce a new env variable that allows to enable or disable pre-authorisation steps
  * if pre-authorisation is enabled on startup the wallet auth state is checked and authorised if not done yet
  * if pre-authorisation is enabled on startup the collateral auth state (! for withelisted collaterals only !) is checked and authorised if not done yet 
  * if pre-authorisation is disabled, the authorisation is only done once the keeper is about to participate in an auction (ie. current behavior) 
* Reasoning for this proposal: 
  * there should be an active way for a keeper operator to decide on authorisation since the related transactions incur fees that at this point in time are not covered by the resulting profit from an auction. Hence it needs to be a conscious decision to do the auth steps in advance with the risk of not amortsing them in the future

**Tasks:**
- [ ] Decide on implementation approach
- [ ] Implement the approach
   - [ ] adjust keeper related readme accordingly 
- [ ] Test pre-authorisation on testnet

benediktkaiser:
Closes #264

Tested the new feature on `hardhat` setup. I also verified that the string to boolean parsing is working correctly, ignoring case and spaces.

On first run:
Screenshot 2022-05-18 at 00 26 21

On second run:
Screenshot 2022-05-18 at 00 27 23

Checklist:

### DIFF
--- a/bot/src/authorisation.ts
+++ b/bot/src/authorisation.ts
@@ -8,7 +8,7 @@ import {
 import { ETHEREUM_NETWORK, KEEPER_PREAUTHORIZE, WHITELISTED_COLLATERALS } from './variables';
 import { parseCollateralWhitelist } from './whitelist';
 
-export async function authorizeKeeperWallet(authorizeCallback?: () => void) {
+export async function checkAndAuthorizeWallet() {
     const signer = await getSigner(ETHEREUM_NETWORK);
     const walletAddress = await signer.getAddress();
     const isWalletAuth = await getWalletAuthorizationStatus(ETHEREUM_NETWORK, walletAddress);
@@ -17,17 +17,13 @@ export async function authorizeKeeperWallet(authorizeCallback?: () => void) {
         console.info(`keeper: wallet "${walletAddress}" has not been authorized yet. Attempting authorization now...`);
         const transactionHash = await authorizeWallet(ETHEREUM_NETWORK, walletAddress, false);
         console.info(`keeper: wallet "${walletAddress}" successfully authorized via "${transactionHash}" transaction`);
-        if (authorizeCallback) {
-            authorizeCallback();
-        }
+        return true;
     }
+    console.info(`keeper: wallet is already authorized`);
+    return false;
 }
 
-export async function checkAndAuthorizeKeeperCollateral(
-    walletAddress: string,
-    collateralType: string,
-    authorizeCallback?: () => void
-) {
+export async function checkAndAuthorizeCollateral(walletAddress: string, collateralType: string) {
     // get collateral authorization status
     const isCollateralAuth = await getCollateralAuthorizationStatus(ETHEREUM_NETWORK, collateralType, walletAddress);
 
@@ -45,10 +41,10 @@ export async function checkAndAuthorizeKeeperCollateral(
         console.info(
             `keeper: collateral "${collateralType}" successfully authorized on wallet "${walletAddress}" via "${collateralTransactionHash}" transaction`
         );
-        if (authorizeCallback) {
-            authorizeCallback();
-        }
+        return true;
     }
+    console.info(`keeper: collateral "${collateralType}" has already been authorized on wallet "${walletAddress}".`);
+    return false;
 }
 
 export async function setupCollateralAuthorizations() {
@@ -57,7 +53,7 @@ export async function setupCollateralAuthorizations() {
     }
 
     // check if the wallet is authorized
-    await authorizeKeeperWallet();
+    await checkAndAuthorizeWallet();
 
     const collaterals = parseCollateralWhitelist(WHITELISTED_COLLATERALS);
     console.info(
@@ -68,6 +64,6 @@ export async function setupCollateralAuthorizations() {
     const walletAddress = await signer.getAddress();
 
     for (const collateral of collaterals) {
-        await checkAndAuthorizeKeeperCollateral(walletAddress, collateral);
+        await checkAndAuthorizeCollateral(walletAddress, collateral);
     }
 }
--- a/bot/src/authorisation.ts
+++ b/bot/src/authorisation.ts
@@ -8,61 +8,61 @@ import {
 import { ETHEREUM_NETWORK, KEEPER_PREAUTHORIZE } from './variables';
 import { getWhitelistedCollaterals } from './whitelist';
 
-export async function checkAndAuthorizeWallet() {
-    const signer = await getSigner(ETHEREUM_NETWORK);
-    const walletAddress = await signer.getAddress();
+export async function checkAndAuthorizeWallet(walletAddress: string): Promise<boolean> {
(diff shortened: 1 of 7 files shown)


bot/src/authorisation.ts
```diff
@@ -0,0 +1,73 @@
...
+        return;
+    }
+
+    // check if the wallet is authorized
+    await authorizeKeeperWallet();
+
+    const collaterals = parseCollateralWhitelist(WHITELISTED_COLLATERALS);
+    console.info(
+        `keeper: "KEEPER_PREAUTHORIZE" is true attempting to authorize collaterals: '${collaterals.toString()}'`
+    );
```
valiafetisov, line 65: In case there is no `WHITELISTED_COLLATERALS` this will fail right? But the issue expect preauth to work even if no whitelisted collaterals specified

bot/src/authorisation.ts
```diff
@@ -0,0 +1,73 @@
...
+} from 'auctions-core/src/authorizations';
+import { ETHEREUM_NETWORK, KEEPER_PREAUTHORIZE, WHITELISTED_COLLATERALS } from './variables';
+import { parseCollateralWhitelist } from './whitelist';
+
+export async function authorizeKeeperWallet(authorizeCallback?: () => void) {
+    const signer = await getSigner(ETHEREUM_NETWORK);
+    const walletAddress = await signer.getAddress();
+    const isWalletAuth = await getWalletAuthorizationStatus(ETHEREUM_NETWORK, walletAddress);
+
+    if (!isWalletAuth) {
```
valiafetisov, line 16: `else` should print "wallet already authorized", right?

bot/src/authorisation.ts
```diff
@@ -0,0 +1,73 @@
...
+        console.info(`keeper: wallet "${walletAddress}" has not been authorized yet. Attempting authorization now...`);
+        const transactionHash = await authorizeWallet(ETHEREUM_NETWORK, walletAddress, false);
+        console.info(`keeper: wallet "${walletAddress}" successfully authorized via "${transactionHash}" transaction`);
+        if (authorizeCallback) {
+            authorizeCallback();
+        }
```
valiafetisov, line 22: Why do you need a callback? I though we're using promises (:

bot/src/authorisation.ts
```diff
@@ -0,0 +1,73 @@
...
+            walletAddress,
+            collateralType,
+            false
+        );
+        console.info(
+            `keeper: collateral "${collateralType}" successfully authorized on wallet "${walletAddress}" via "${collateralTransactionHash}" transaction`
+        );
+        if (authorizeCallback) {
+            authorizeCallback();
+        }
```
valiafetisov, line 50: Same here:
- callback is unnecessary 
- else should print "... already authorized ..."

bot/src/keeper.ts
```diff
@@ -93,38 +89,16 @@ const checkAndParticipateIfPossible = async function (auction: AuctionInitialInf
 
     // try to authorize the wallet then return
     if (!isWalletAuth) {
-        console.info(`keeper: wallet "${walletAddress}" has not been authorized yet. Attempting authorization now...`);
-        const transactionHash = await authorizeWallet(ETHEREUM_NETWORK, walletAddress, false);
-        console.info(`keeper: wallet "${walletAddress}" successfully authorized via "${transactionHash}" transaction`);
-        await checkAndParticipateIfPossible(auction);
-        return;
+        await authorizeKeeperWallet(async () => await checkAndParticipateIfPossible(auction));
```
valiafetisov, line 92: Why do you need to check `isWalletAuth` here if you will check it within the function? Should function then be called `checkAnd...`?

bot/src/authorisation.ts
benediktkaiser: Yes, it returns if either `WHITELISTED_COLLATERALS` is not set or `KEEPER_PREAUTHORIZE` is false. 

As this part of the function is also ran after validating the whitelist, we can be sure, that the collaterals entered exist and are correct.

bot/src/authorisation.ts
benediktkaiser: We did not have this beforehand, but I can add it now!

bot/src/authorisation.ts
benediktkaiser: I needed it to recall `checkAndParticipateIfPossible`, however I changed it to return 

- `true` if the collateral or wallet was newly authorised
- `false` if the collateral or wallet was previously already authorised

When calling the functions I then check the return value and can then reload the entire function if needed.

bot/src/authorisation.ts
benediktkaiser: See above.

bot/src/keeper.ts
benediktkaiser: Removed this!

bot/src/authorisation.ts
valiafetisov: I meant that pre-authorisation should happen on **all collaterals** if no `WHITELISTED_COLLATERALS` are specified. 

I also suggest that instead of parsing `WHITELISTED_COLLATERALS` every time you need them, you should have one place where you define something like `getCurrentCollaterals` (that does not require passing env variable) and receive list of collaterals that are either all or whitelisted.

LukSteib:
- [ ] See @valiafetisov comment [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/269#discussion_r879228584). If no collateral is whitelisted but pre-auth set to `true` let's authorise all of the collateral types + wallet. 

Other than that looking good.

bot/src/authorisation.ts
```diff
@@ -0,0 +1,69 @@
...
+import {
+    authorizeCollateral,
+    authorizeWallet,
+    getCollateralAuthorizationStatus,
+    getWalletAuthorizationStatus,
+} from 'auctions-core/src/authorizations';
+import { ETHEREUM_NETWORK, KEEPER_PREAUTHORIZE } from './variables';
+import { getWhitelistedCollaterals } from './whitelist';
+
+export async function checkAndAuthorizeWallet() {
```
valiafetisov, line 11: I'm not sure why typescript return type is not required here and in other functions 🙁

PR #272, opened 2022-05-22
Refactored how meta mask errors are parsed and added stories

### LINKED ISSUE #259: Improve error message parsing

benediktkaiser:
Closes #259 

Checklist:

Try number two! I saw that all Meta mask errors have a small explanation before the first `(`, therefore I split the message there and display the front message. Is this what we want?

I also added stories for all errors logged by @LukSteib


core/src/helpers/parseMetamaskError.ts
```diff
@@ -1,15 +1,8 @@
...
-    try {
-        const jsonString = errorMessage.substring(jsonFirstIndex, jsonLastIndex + 1);
-        const metamaskError = JSON.parse(jsonString);
-        return truncateText(metamaskError?.value?.data?.message || 'unknown');
-    } catch {
-        return truncateText(errorMessage || 'unknown');
-    }
+const parseMetamaskError = function (errorMessage = ''): unknown {
+    const splitError = errorMessage.split('(')[0];
+    return truncateText(splitError);
```
valiafetisov, line 5: This will display only the first part of the error, while the most important one, as already said in the issue and in the previous PR would be the part which is inside `value.data.message`. As suggested before the message displayed to the user should be `${metamaskError.value.data.message} (${metamaskErrorMessage})`. But you currently only display `metamaskErrorMessage`

frontend/components/ErrorMessages.stories.js
```diff
@@ -0,0 +1,52 @@
...
...0000000000000000000000200000000000000000000000000000000000000000000000ff4327f3f09430c010000000000000000000000000000000000000004aaebc81e9fe122a2aab7215400000000000000000000000074893c37beacf205507ea794470b13de0629422000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000180000000000000000000000000b2e1284fb685dbdcc4378cf5d4e9515679527a42000000000000000000000000f11a98339fe1cde648e8d1463310ce3ccc3d7cc1000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000a0000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000000030000000000000000000000001f9840a85d5af5bf1d1762f925bdaddc4201f984000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f0000000000000000000000000000000000000000000000000000000000000002000000000000000000000000c02aaa39b223fe8d0a0e5c4f27ead9083c756cc20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f","type":2,"accessList":null}, code=UNPREDICTABLE_GAS_LIMIT, version=providers/5.4.5)\n',
```
valiafetisov, line 20: As noted before: the actual error message thrown by metamask **will not** contain `Bidding error Error:`. Explanation can be found in the other pr. Please remove it from here and from other error messages

benediktkaiser:
I updated the parsing method and added the stories. You can preview the outputs in Storybook!

core/src/helpers/parseMetamaskError.ts
```diff
@@ -1,12 +1,20 @@
...
+        return truncateText(errorMessage || 'unknown');
+    }
     try {
-        const jsonString = errorMessage.substring(jsonFirstIndex, jsonLastIndex + 1);
-        const metamaskError = JSON.parse(jsonString);
-        return truncateText(metamaskError?.value?.data?.message || 'unknown');
+        const firstTrim = errorMessage.substring(jsonFirstIndex + 11);
+        const jsonLastIndex = firstTrim.indexOf('"');
+        const trimmedErrorMessage = firstTrim
+            .substring(0, jsonLastIndex)
```
benediktkaiser, line 12: I don't like that I have to call so many `replace` methods. Maybe @valiafetisov you can help me out and generate (or teach me) a rejex expresion to make it a one liner?

core/src/helpers/parseMetamaskError.ts
```diff
@@ -1,12 +1,20 @@
...
+const parseMetamaskError = function (errorMessage = ''): unknown {
+    const jsonFirstIndex = errorMessage.indexOf('"message":"');
+    if (jsonFirstIndex === -1) {
+        return truncateText(errorMessage || 'unknown');
+    }
     try {
-        const jsonString = errorMessage.substring(jsonFirstIndex, jsonLastIndex + 1);
-        const metamaskError = JSON.parse(jsonString);
-        return truncateText(metamaskError?.value?.data?.message || 'unknown');
+        const firstTrim = errorMessage.substring(jsonFirstIndex + 11);
```
valiafetisov, line 9: Please don't hardcode `11`, but derive it from the string you're using

core/src/helpers/parseMetamaskError.ts
valiafetisov: I'm not sure, what all those replacements do, why do we need them?

benediktkaiser: I wanted to clean up the final error message to remove characters that do not contain any valuable information. 

The first error message would look like this:
`[ethjs-query] while formatting outputs from RPC '{\`, so I removed the special characters at the end to clean it up.

valiafetisov:
Final parsing results looks like this (black is the original error, red is the parsed message that will be displayed to the user):

screencapture-localhost-8025-iframe-html-2022-06-09-18_14_48

LukSteib (approved):

LukSteib:
> @LukSteib, do you think it's good to go?

Looking good

PR #273, opened 2022-05-23
Fixed HeapIO Integration

### LINKED ISSUE #267: Fix heap io integration
**Goal:**
Ensure to receive user data for staging and production

**Context:**
Currently we only receive user data for development environment. 

Staging console shows the following error:
`heapIO initialization failed. Analytics are disabled`
Stack trace:
`loadAnalytics	@	5733c0d.js:1`

Production console shows the following error:
`HEAPIO_ID is not defined. Analytics are disabled`
Stack trace:
`loadAnalytics | @ | 50a2c41.js:1`

For all three environments the correct `heapio_id` is set via env variables (double checked that)

**Tasks:**
- [x] Investigate heap problems on staging and production
   - [x] document reasons in this issue
- [x] Fix potential problems

fegbert:
Closes #267 

Checklist:

fixes are:
- Added HeapIO ID to drone file to env variable gets properly set in environments now
- Disabled static prefix for nuxt to page will always look at `~/static` instead of relative to the page.
  - Locally doesn't break anything, checked with T&S file and heapIO.

### DIFF
--- a/frontend/nuxt.config.js
+++ b/frontend/nuxt.config.js
@@ -34,7 +34,7 @@ export default {
             { hid: 'description', name: 'description', content: '' },
         ],
         link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
-        script: [{ src: 'js/HeapIO.js' }],
+        script: [{ src: '/js/HeapIO.js' }],
     },
 
     // Global CSS: https://go.nuxtjs.dev/config-css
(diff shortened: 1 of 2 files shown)


frontend/nuxt.config.js
```diff
@@ -71,6 +71,11 @@ export default {
         domain: process.env.PRODUCTION_DOMAIN,
     },
 
+    // Disable Static Asset Path Prefix: https://nuxtjs.org/docs/directory-structure/static/#static-asset-prefixs
+    static: {
+        prefix: false,
+    },
```
valiafetisov, line 77: Why do you need to disable it? I don't think we can do that, since this folder contains favicon and Terms and Conditions pdf

fegbert: You're right, I thought the access to the folder was the problem, but I think it was only the path to the file, since those 2 things work in staging & prod.

The problem might've been that I used `src: 'js/HeapIO.js'` instead of `src: '/js/HeapIO.js'`

valiafetisov (approved):

PR #274, opened 2022-05-23
Fixed Profit Calculation

### LINKED ISSUE #232: Investigate profit calculation
**Context:**
We experienced our profit calculation to not be accurate all the time. 
See recording [here](https://github.com/sidestream-tech/auction-ui/issues/407#issue-1204556894) -> LOCAL TESTING video

We saw that similar problems were tackled in the auction demo keeper implementation [here](https://github.com/makerdao/auction-demo-keeper/commit/74bfb2d34dbc58f819293e099050fed2ce687525)

First conclusion by @valiafetisov: 
> We should definitely use this sophisticated calculateTransactionCollateralOutcome mirroring of the contract to calculate our profit in swap transaction as well and also account for the states when auction can't be executed for some strange reason

**Tasks:**
- [ ] Look into profit calculation (especially for swap transaction) and potential change to the more sophisticated `calculateTransactionCollateralOutcome` approach

fegbert:
Closes #232 

Checklist:

### DIFF
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -44,12 +44,30 @@ const enrichAuctionWithActualNumbers = async function (
     };
 };
 
+const calculateCollateralToCoverDebt = async function (network: string, auction: Auction): Promise<BigNumber> {
+    const transactionCollateralOutcome = calculateTransactionCollateralOutcome(
+        auction.debtDAI,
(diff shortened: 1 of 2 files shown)


valiafetisov:
Closing since the try seems to be unsuccessful

PR #281, opened 2022-05-31
Disabled refresh on network switch

### LINKED ISSUE #263: Disable page refresh on network switch
**Goal:**
We no longer refresh the whole page when switching the selected network

**Context:**
Currently there are three ways to switch the network: 
a) via the network dropdown in the header of the UI
b) via metamask extension
c) directly passing the desired network via URL parameter
In case there is a mismatch between UI and metamask a popup is shown to either adjust UI or to switch wallet. No matter which selection is made to align UI and wallet we hard refresh the whole page. 
This should be prevented in the future. 

**Tasks:**
- [ ] Prevent hard refresh of the application when switching network

fegbert:
Closes #263 

**current flow**

---

Checklist:


fegbert:
## Remaining Problems

- I have to manually wait for the walletNetwork to update
  - I am awaiting the [setWalletNetwork]() setter in the network store
  - After this resolves I would expect the `getWalletNetwork` getter to return the new network.
    - It doesn't 😢
  - The getter is only being updated when the `walletChainId` is switched.
    - This is done by calling the `setWalletChainId` setter of the network store
  - It gets called by the [networkChangedHandler](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/lib/wallets/MetaMask.ts#L73), which gets called by [MetaMasks chainChanged](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/lib/wallets/MetaMask.ts#L92) event.
    - The event gets triggered "long" after `setWalletNetwork` has been resolved, **so this is what is causing the delay**

**TLDR:** `chainChanged` event gets triggered after setWalletNetwork is already resolved, so getter doesn't update in time

frontend/components/common/Select.vue
```diff
@@ -108,7 +112,11 @@ export default Vue.extend({
     watch: {
         localVisible: {
             handler(newVisible) {
-                this.$emit('update:visible', newVisible);
+                if (this.disabled) {
+                    this.localVisible = false;
+                } else {
```
valiafetisov, line 117: Can you disable the actual selects / modals instead of silently skipping emit here? I would even say that we should add isLoading animation to the network label in the header (so probably you should pass it as `isLoading` to the `Select`)

valiafetisov:
[`setWalletNetwork`](https://github.com/sidestream-tech/unified-auctions-ui/blob/c4b90b6246bbdd02e1e09e5ef7a7e4d33e8e6219/frontend/store/network.ts#L90-L98) is an action (not setter). It resolves as soon as `provider.send('wallet_switchEthereumChain', [{ chainId }])` gets resolved, not when the network change happens, so it's not surprising. We can not change this behaviour since it's implemented by Metamask. We could of course set it ourselves, but sending `wallet_switchEthereumChain` without an exception doesn't mean it will actually change the network, so I would keep current implementation and wait for the `chainChanged`.

But instead of using `setInteral` loop to watch for the state change, can you maybe use watcher either via little-known [vuex `watch` method](https://vuex.vuejs.org/api/#watch) or a normal vue watcher somewhere in the layout / header / network switch? Or would it become even more hacky?

frontend/store/authorizations.ts
```diff
@@ -243,4 +243,10 @@ export const actions = {
             commit('setIsAllowanceAmountLoading', false);
         }
     },
+    async setup({ commit, dispatch }: ActionContext<State, State>) {
+        commit('reset');
+
+        await dispatch('refetch');
+        await dispatch('fetchAllowanceAmount');
```
valiafetisov, line 250: I believe you also need to fetch authorizations for wallet and specific opened collateral. This logic is already implemented for "wallet switch" so you would need to refactor it into a single place

frontend/store/collaterals.ts
```diff
@@ -59,6 +67,9 @@ export const mutations = {
     setCollateralStatus(state: State, collateralStatus: CollateralStatus) {
         Vue.set(state.collateralStatusesStorage, collateralStatus.type, collateralStatus);
     },
+    reset(state: State) {
+        Object.assign(state, getInitialState());
```
valiafetisov, line 71: Since getInitialState returns a fresh object, can't you just:
```suggestion
        state = getInitialState();
```
? (here and everywhere else)

frontend/store/auctions.ts
```diff
@@ -352,4 +357,9 @@ export const actions = {
             commit('setAreTakeEventsFetching', false);
         }
     },
+    async setup({ commit, dispatch }: ActionContext<State, State>) {
+        commit('reset');
+
+        await dispatch('fetch');
```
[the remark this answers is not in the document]
fegbert: Yep you're right. This should be `update` not `fetch`

frontend/store/collaterals.ts
fegbert: Yes, but then eslint starts screaming about no unused vars 😅

fegbert:
I already like it a lot more with using the watcher, but there are still some problems / there's still some hacky stuff:
1. Handling of the `isChangingNetwork` state needs to be done in 2 different actions
1.1 Gets set to `true` in the `setPageNetwork` action, this is good
1.2 Needs to get set to `false` in the setup action because otherwise it's too early, this is bad
2. When only switching page network the `walletChainId` never changes => watcher doesn't trigger => setup doesn't get called so I have to call it manually every time after calling `setPageNetwork`.

frontend/components/common/Select.vue
fegbert: The popover / tooltip doesn't seem to have a disabled property, so this was my attempt to do that

For the animation, I already added a loader when `isChangingNetwork` is true which replaces the network icon for that time.

fegbert:
> And only when page network === wallet network, we should call "setup" to refetch all data.

This gave me a good idea. Instead of watching the `walletChainId` which only changes when switching the wallet network, I'm now watching the `getMakerNetwork` which is only not undefined if your proposed check applies. This works perfectly, thanks! 🙂 

> Maybe isChangingNetwork can be derived from the state when page network !== wallet network? Or is it exactly to prevent popup to be shown while wallet is applying changes?

Yea this is used to prevent the fix network popup from showing so using this check wouldn't really work for it.

valiafetisov:
> Yea this is used to prevent the fix network popup from showing in only that period of time, so using this check wouldn't really work for it.

Then, maybe it's fine?

> Handling of the `isChangingNetwork` state needs to be done in 2 different actions

Can the process be stopped in-between those two actions? Eg what would happen if the user rejects network change?

fegbert:
After testing around a bit, I don't think there is a way to stop in between. If the user rejects / the changing fails, `isChangingNetwork` also gets set back to `false`

frontend/components/common/Select.vue
valiafetisov: > The popover / tooltip doesn't seem to have a disabled property

- If no content is provided they are not shown
- You can disable all buttons within the popover
- You can replace the network label with `Loading...` which doesn't have popover over it at all

I would go with the last option, but silently skipping user action is an anti-pattern

frontend/store/auctions.ts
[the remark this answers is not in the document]
valiafetisov: Does running `update` restart/cancels previously set loop? Or is it the addition to the other requests?

frontend/store/authorizations.ts
valiafetisov: Yes, seems like it. Do you want to move `fetchAllowanceAmount` call to the `refetch` as well?

frontend/layouts/default.vue
```diff
@@ -115,6 +118,13 @@ export default Vue.extend({
             return this.$nuxt?.context?.isDev;
         },
     },
+    watch: {
+        getMakerNetwork(newValue) {
+            if (newValue) {
+                this.$store.dispatch('network/setup');
+            }
+        },
```
valiafetisov, line 126: Would this also be called on the page load (after persistent values are inserted into the store)? – This will result in doubling the amount of requests.

Should we call the same thing `network/setup` initially? Currently [something else is called here](https://github.com/sidestream-tech/unified-auctions-ui/blob/8d976c1c74618c3179059c311d232d2dcf2ee949/frontend/plugins/vuex-persist.client.js#L18-L21) (because of the problems with persistent storage which get initialised sometimes before, sometimes after the layout mount)

LukSteib (changes requested):
- [x] switching to `localhost:8545` network ends up in never ending loading state (for both ways, switching from UI or switching from Metamask) see [1]. The auction table is not updated



[1]
image

frontend/store/authorizations.ts
fegbert: Sure, I think that makes sense

frontend/components/common/Select.vue
fegbert: Yep, the last one sounds the best to me too. Will add that!

frontend/store/auctions.ts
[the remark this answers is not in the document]
fegbert: Now I'm confused 😅 

We probably want to "restart" the looping on switch, right? So stop the existing loop and replace it with a new one?

`fetch` does that. It first checks if there's already intervals running and then stops them if so before starting new ones.
`update` does not touch the intervals at all

fegbert:
> switching to localhost:8545 network ends up in never ending loading state

This should only happen when hardhat isn't running. Then of course it is loading infinitely since the network will never successfully switch.

Maybe we could add a timeout or something?

LukSteib:
> This should only happen when hardhat isn't running. Then of course it is loading infinitely since the network will never successfully switch.

> Maybe we could add a timeout or something?

Makes sense and then it is of course also not a big issue. If setting a timeout for these cases is easy to implement. Let's do it. If it's hard let's not optimise at this point. 
What happens after the timeout? From user experience would be fine to just fall back to prev or default network

fegbert:
Added timeout (currently 8 seconds, don't know how long something like this should be 😅 )
When timed out, the page just goes back to the previous network.

LukSteib (changes requested):
- [ ] the fallback logic somehow breaks the overall network switching functionality. No longer able to switch between networks when there is no wallet connected and when there is a wallet connected (see recording below) 

https://www.loom.com/share/86dbbaa627f34609a0895d74e3a51a39

LukSteib (dismissed):

valiafetisov:
We also have `fetchRelatedData ` function within the wallet modal, which should also be refetched on the network switch. https://github.com/sidestream-tech/unified-auctions-ui/blob/bc6bf2310c4997ed22145c6d399196f1f969674c/frontend/containers/WalletModalContainer.vue#L103-L108

Please move the content of it to the related setup actions and replace its content with the setup action(s)

frontend/store/network.ts
```diff
@@ -60,31 +64,59 @@ export const getters = {
...
     ): Promise<void> {
         const oldNetwork = getters.getPageNetwork;
         if (newNetwork === oldNetwork) {
             return;
         }
+        commit('setIsChangingNetwork', true);
         try {
             await dispatch('setWalletNetwork', newNetwork);
-            window.location.href = `${rootState.route.path}?network=${newNetwork}`;
+            await window.$nuxt.$router.replace({
```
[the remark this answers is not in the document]
fegbert: > to give the user possibility to go back to the previous network

Yes you're right, that wouldn't be possible with `replace`. Replaced it 🙂

frontend/store/network.ts
```diff
@@ -60,31 +64,59 @@ export const getters = {
...
+            networkChangeTimeoutId = setTimeout(() => {
+                if (getters.getMakerNetwork === undefined) {
+                    message.error(`Failed to change network to ${newNetwork}`);
+                    commit('setIsChangingNetwork', false);
+                    window.$nuxt.$router.replace({
+                        query: {
+                            network: oldNetwork,
+                        },
+                    });
+                }
```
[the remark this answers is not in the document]
fegbert: > Please check if you need to clear networkChangeTimeoutId in some other actions (not sure about this one)

Nice call! After testing some things, I've found that switching the network, and then, while the timeout is waiting, switching it again but with the MetaMask extension, the timeout doesn't get, reset leading to unwanted behavior. 

**reference**



I think also clearing it in the `setup` action is the best way to fix this so that whenever the `walletChainId` changed, the timeout gets cleared.

frontend/store/auctions.ts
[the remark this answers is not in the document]
fegbert: `fetch` also calls `updateAuctionsPrices` (since it starts its loop), but I think that's fine, if not even wanted, right?

fegbert:
> We also have fetchRelatedData function within the wallet modal, which should also be refetched on the network switch.

Should I also do this for the `AuctionsContainers` `fetchRelatedData`?

valiafetisov:
> Should I also do this for the `AuctionsContainers` `fetchRelatedData`?

Yes, that would be nice to universalise those calls via central `setup` method(s) that you've created. Or do you see any problems with that?

fegbert:
Nope, only had to add one more fetch to the wallets setup 🙂 

I did notice though that [after restarting auctions](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/store/auctions.ts#L308) we are calling a `fetchWithoutLoading` action which doesn't seem to exist anymore.

valiafetisov:
> we are calling a `fetchWithoutLoading` action which doesn't seem to exist anymore

Good catch! Do you also want to fix it in this PR? I guess you would need to call `update` instead?

valiafetisov (approved):

PR #284, opened 2022-06-08
Support custom rpc urls

### LINKED ISSUE #222: Support custom RPC url(s)
### Goal

The user is able to supply RPC url via environment variable instead of infura project id

### Context

1. Currently we only let users (who wants to start our UI or the bot locally) provide `INFURA_PROJECT_ID`. But this is limiting, since users may use different providers as well as custom / local nodes
2. Currently the user is only able to connect to a set of networks from the list of supported ones and there is no possibility to dynamically add new networks to our UI
  - Use case: user is able to use custom PRC endpoint of their own
  - Use case: we're able to integrate chaoslabs forks

Note: most probably, those two scenarios require two different solutions. This issue outlines them here because we don't want them to be in conflict with each other. 

### Tasks

- [x] Come up with a solution(s), outline it here
- [x] Discuss
- [x] Implement

benediktkaiser:
Closes #222

Checklist:

### DIFF
--- a/bot/src/index.ts
+++ b/bot/src/index.ts
@@ -1,5 +1,4 @@
 import { setTimeout as delay } from 'timers/promises';
-import { getNetworkConfigByType } from 'auctions-core/src/constants/NETWORKS';
 import { getAllAuctions, getNewAuctionsFromActiveAuctions } from './auctions';
 import notify from './notify';
 import participate, { setupKeeper } from './keeper';
@@ -28,7 +27,6 @@ const loop = async function (): Promise<void> {
 
 const start = async function (): Promise<void> {
     await delay(SETUP_DELAY);
-    getNetworkConfigByType(ETHEREUM_NETWORK);
     setupWhitelist();
     await setupTwitter();
     await setupKeeper();
--- a/bot/src/keeper.ts
+++ b/bot/src/keeper.ts
@@ -17,7 +17,7 @@ export const setupKeeper = async function () {
         return;
     }
     try {
-        setSigner(ETHEREUM_NETWORK, createSigner(ETHEREUM_NETWORK, KEEPER_WALLET_PRIVATE_KEY));
+        setSigner(ETHEREUM_NETWORK, createSigner(KEEPER_WALLET_PRIVATE_KEY));
         const signer = await getSigner(ETHEREUM_NETWORK);
         const address = await signer.getAddress();
         isSetupCompleted = true;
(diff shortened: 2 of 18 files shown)


bot/src/index.ts
```diff
@@ -1,9 +1,8 @@
 import { setTimeout as delay } from 'timers/promises';
-import { getNetworkConfigByType } from 'auctions-core/src/constants/NETWORKS';
+import setupNetworks, { getNetworkConfigByType } from 'auctions-core/dist/src/networks';
```
valiafetisov, line 2: Please make sure to never use `dist` folder, since this folder is only updated on the `npm run build` executed inside core folder.

```suggestion
import setupNetworks, { getNetworkConfigByType } from 'auctions-core/src/networks';
```

That's the reason you were stack with outdated code. Here and everywhere else

benediktkaiser:
What has been done:
- Updated core functions to use dynamic network system
- Updated `bot` to use RPC URLs

What has to be done:
- Update the frontend to use the new dynamic system
  - I already created a new store, that automatically updates the network list according to the core functions, however the auction retrieving does not wait for the networks to be loaded, therefore a delay or await needs to be implemented. 
  - I currently retrieve the networks and then set the selected network, however this poses new issues when a network is defined in the url, so that will need to be handled. 
  - I also need to wait for #281 in order to avoid endless refresh loops due to me calling our `setPageNetwork`, which currently means that the page loads, then refreshes once it sets the selected network and there is a url mismatch. 
  - some component stories are also still broken, as we used the static networks to test storybook
- Update the `bot` and `frontend` ReadMes to explain the new RPC system and remove old references to the infura project id.

benediktkaiser:
Brief summary of call with Valia about this PR:

- We will pause all work until #281 is merged, as it changes some things about the network store, which would result in a lot of merge conflicts with this PR. 
- We concluded, we need to create the providers for the networks before adding them, as we need to fetch the ChainID. This poses a new issue, as we do not know how long this fetching may take (if a invalid URL is entered, it can take until our timeout time to resolve). Therefore we need to refactor our auction fetching logic in the stores to await the network setup. Where can I find the exact logic for this? I would then purpose that I move the network setup into the first call of the Auctions Fetching, in which I can then await the network setup and then fetch the auctions.
- I needed to rewrite multiple components, to receive networks as a prop, rather then a direct import as before. This will require the stories to be rewritten and a helper function to be created, that can mock network configs.

benediktkaiser:
So I changed the order of the fetching however I have the following issues:

- Before networks are loaded I get this error message: `Error while fetching tokenAddressDai: No network found with name "goerli"`. Where is this triggered/What is this for? 
- After loading the networks, the change network modal is triggered and it says, that the network I am on is not supported (even tho it should be)
- I still have to look into the url switching, however I will do this after we merge #281, as it changes some of the logic for it.

LukSteib:
> I still have to look into the url switching, however I will do this after we merge https://github.com/sidestream-tech/unified-auctions-ui/pull/281, as it changes some of the logic for it

The respective PR (page refresh on network switch) has been merged now

PR #287, opened 2022-06-09
feat: e2e cypress test for collaterals

### LINKED ISSUE #277: Investigate end-to-end testing
### Goal

Outlines of different end-to-end testing possibilities

### Context

We want to introduce some high-level testing scenarios where the main features of our main UI are tested. For example: that list of auctions is displayed and a user is able to participate in one of them (eg execute both auth steps and the transaction). The test should also independently check that blockchain have expected changes (eg `Take` event has been created)

I expect both should be possible:
- a) on a highest-level, ie via headless browser-based testing tool that installs metamask, spawns hardhat simulation and "click" buttons to reach desired final state
- b) via execution of specific (core) functions (or using a bot against hardhat)

Note: this feature most probably needs #222 

### Tasks

- [ ] Investigate feasibility of both approaches
- [ ] Propose testing approach
- [ ] Propose concrete testing tools/frameworks

### LINKED ISSUE #282: Implement end-to-end tests
### Goal

Implement end-to-end test

### Context
Implement end-to-end test based on the extensive investigation and discussion here: #277 

### Tasks

- [ ] Implement end-to-end test

KirillDogadin-std:
Closes #282
closes #277 

Checklist:


KirillDogadin-std:
The feature though lacks the automatic reset of the hardhat. Within the cypress test the hardhat related function call results in the error 

```
Module not found: Error: Can't resolve 'async_hooks' in '~~~~/unified-auctions-ui/frontend/node_modules/undici/lib/api'
```

which apparently (judging by https://stackoverflow.com/questions/69460177/cypress-error-cant-resolve-async-hooks) is due to the cypress and hardhat expecting different contexts. I am not sure whether at this stage we should bother with making this work.

LukSteib (commented):
Unsure what to pay attention to during functional test to be honest. 
Execute the relevant commands and the run passed (see screenshot below).
Is there a cypress UI that I have/can double-check or anything else? 

- [x] one remark: cypress folder is currently inside frontend. So either wrong place or you have to update the readme section accordingly.

image

frontend/README.md
```diff
@@ -42,3 +42,33 @@ Help on both things is given in the linked resources above.
...
+> npm run dev
+> cd ../hardhat
+> npm run dev
+> cd ../cypress
+> npx cypress run
+```
+
+`cypress run` by default records a video. The video recording is [disabled](https://docs.cypress.io/guides/guides/screenshots-and-videos#Videos) in `cypress/cypress.json` config.
+
+The e2e tests should be executed on the freshly started hardhat network - this is due to the fact that the test changes the state of the chain, so the state has to be reset on every run of the e2e suite. That can be achieved either via programmatic means or manual restarts (manual restart is less preferred.)
```
valiafetisov, line 65: - What exactly happens when tests are executed second time on the same hardhat fork?
- Do you want to remove sentence about `programmatic means`?

KirillDogadin-std:
Is there a cypress UI that I have/can double-check or anything else?

https://github.com/sidestream-tech/unified-auctions-ui/pull/287/files#diff-35aa98d245b201cfc92c74313abcf8fa0f6069fc750f05e7a755665360a0e543R73

if that line was noticed but interpreted incorrectly, let me know.

---

> So either wrong place or you have to update the readme section accordingly.

in the examples i've seen cypress directory is in the service's direcroy. I've messed up the `cd`s in the readme :(

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
+import { BigNumber } from 'ethers';
+import { expect } from 'chai';
+
+describe('Collateral auctions', function () {
+    // Set the private key to one of the deterministic hardhat keys.
+    // see https://hardhat.org/hardhat-network#running-stand-alone-in-order-to-support-wallets-and-other-software
+    const KEEPER_WALLET_PRIVATE_KEY = '0x701b615bbdfb9de65240bc28bd21bbc0d996645a3dd57e7b12bc2bdf6f192c82';
+    const URL_TO_VISIT = `${process.env.FRONTEND_URL || 'localhost:3000'}/collateral?network=localhost`;
```
valiafetisov, line 8: If you introduce new env variable, you need to:
- set it in the nuxt config and move default value there
- update dockerfile

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+    let daiBalanceBefore: Record<string, BigNumber>;
+    let daiBalanceAfter: Record<string, BigNumber>;
```
valiafetisov, line 10: DAI balance is a BigNumber, not record. Please either rename it to `walletBalancesBefore` or store only DAI balance

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    function testSwapProfit() {
+        cy.visit(URL_TO_VISIT);
+        cy.window()
+            .its('$nuxt')
+            .then(function (nuxt) {
+                nuxt.$store.dispatch('wallet/createWalletFromPrivateKey', KEEPER_WALLET_PRIVATE_KEY);
+            });
+        cy.get('table').first({ timeout: 140000 }).should('contain.text', 'ETH');
```
valiafetisov, line 19: Please move this number into a variable, something like `PAGE_LOAD_TIMEOUT_MS` and please use multiplication to increase readability, eg `140 * 1000` or `2.5 * 60 * 1000`

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+
+        cy.window()
+            .its('$nuxt')
+            .then(function (nuxt) {
+                daiBalanceBefore = nuxt.$store.getters['wallet/walletBalances'];
+            });
+
+        cy.get('a').eq(1).should('contain.text', 'Participate').click();
+        cy.get('button.Button').eq(3).should('contain.text', 'Directly swap').click();
```
valiafetisov, line 28: So adding a new unrelated button to the page can break the test?  That's called fragile tests. We don't want that. Good test should be as readable as specification of what should happen without any magic in-depth knowledge of selectors involved.

Based on this: please use more specific selectors or add new classes to the buttons in question, eg `SwapButton`

here and everywhere else

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+        cy.get('button.Button')
+            .eq(5)
+            .should('contain.text', 'Authorize DAI Transactions')
+            .click()
+            .then(() =>
+                cy
+                    .get('button.Button')
+                    .eq(6)
+                    .should('contain.text', 'Authorize ETH-C Transactions')
+                    .click({ timeout: 19000 })
```
valiafetisov, line 38: Based on what are those hardcoded timeouts? I would suggest to have a single one per type

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+            });
+
+        cy.get('a').eq(1).should('contain.text', 'Participate').click();
+        cy.get('button.Button').eq(3).should('contain.text', 'Directly swap').click();
+        cy.get('button.Button')
+            .eq(5)
+            .should('contain.text', 'Authorize DAI Transactions')
+            .click()
+            .then(() =>
+                cy
```
valiafetisov, line 34: Why do you sometimes need to put another chain of executions inside then? Can't it be moved to the new line, since cypress still magically "awaits" every line

frontend/cypress/integration/basic-spec.ts
valiafetisov: Another approach will be to use contains directly, without specifying a selector, [according to the docs](https://docs.cypress.io/api/commands/contains#Syntax):

```suggestion
        cy.contains('Directly swap').click();
```

But it doesn't really work for all buttons, since some buttons contains content-specific values, so adding readable class names is more generic approach

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+                    .click({ timeout: 19000 })
+            );
+
+        cy.get('button.Button')
+            .eq(8)
+            .should('contain.text', 'Execute')
+            .click({ timeout: 25000 })
+            .then(() => {
+                cy.get('svg.CloseIcon').eq(1, { timeout: 10000 }).click();
+                cy.get('svg.CloseIcon').first({ timeout: 10000 }).click();
```
valiafetisov, line 47: What does this do? Can't you just have a greater parent timeout?

frontend/cypress/integration/basic-spec.ts
valiafetisov: By the way, does timeout here is the _maximum_ timeout or is it awaited anyway? With `puppeteer` the common pattern is `waitForSelector('selector', { timeout: amount_ms })` where it is awaited maximum for the `amount_ms`

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+            });
+    }
+    it('participates with swap and with direct bid', function () {
+        testSwapProfit();
+        testBidWithDai();
+        cy.window()
```
valiafetisov, line 76: Currently, you check `daiBalanceAfter` after you executed both transactions, which means
- `testBidWithDai` is not actually checked (dai is increased after first transaction, but should be decreased after second, it can not be decreased above 0)
- It would not always be the case that you will have enough DAI to bid from the first transaction

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,84 @@
...
+            .should('contain.text', 'Manage DAI in VAT')
+            .click()
+            .then(() => {
+                cy.get('svg.animate-spin').eq(3, { timeout: 140000 }).should('not.be.visible');
+                cy.contains('Allow unlimited access to DAI').click();
+                cy.get('button').eq(30).should('contain.text', 'deposit').click();
+                cy.get('span.ant-modal-close-x').first().click();
+            });
+        cy.get('input.Input').first().type('100');
+        cy.contains('Authorize WSTETH-A Transactions').click();
```
valiafetisov, line 64: I would suggest to not write fragile tests that break if we move the block number. `WSTETH-A` should not be part of this test

frontend/README.md
```diff
@@ -42,3 +42,33 @@ Help on both things is given in the linked resources above.
 - `DEFAULT_ETHEREUM_NETWORK`: (optional, default `mainnet`) Can be set to change the default ethereum network
 - `MAX_PRIORITY_FEE_PER_GAS_WEI`: (optional, default can be found in core/src/gas.ts) – [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) `max_priority_fee_per_gas` value
 - `HEAPIO_ID`: (optional) [HeapIO analytics](https://heapanalytics.com/) Project's Environment ID. Required for tracking analytics.
+
+## End-to-end testing
+
+**Running tests locally**
+
+The e2e testing is achieved via `cypress` tool. The directory that contains the tests is, correspondingly, `cypress/`.
```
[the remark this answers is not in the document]
KirillDogadin-std: The pattern i've observed throughout various unofficial and [official] sources is to name directory with cypress related logic `cypress/`.

  - it contains a lot of additional functionality
  - runs the tests in browser context and not in node.

so naming this `tests` and putting the cypress there would then remove the segreagation between usual tests and the cypress ones.

frontend/README.md
KirillDogadin-std: > What exactly happens when tests are executed second time on the same hardhat fork?

they fail because they expect the authorizations to not be present (and the ui thus has the corresponding buttons disabled and with different text)

frontend/cypress/integration/basic-spec.ts
KirillDogadin-std: this is the maximum. as soon as the expectation is met, the flow continues.

frontend/cypress/integration/basic-spec.ts
KirillDogadin-std: Turns out that `contains` accepts regexps which makes selection easier 

( initial problem that caused the hack was that it selects the first element that contains the argument )

frontend/cypress/integration/basic-spec.ts
KirillDogadin-std: this closes the popups to return to the main page, added comment.

actually, these timouts are not needed at all and adjustments to the parent timout is also not required

hardhat/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM node:16.9.1 as dependency-base
+
+ENV APP_ROOT /app
+
+RUN mkdir ${APP_ROOT}
+WORKDIR ${APP_ROOT}
+ADD . ${APP_ROOT}
+
+RUN npm ci
```
valiafetisov, line 9: Please:
- first COPY `package.json` and `package-lock.json`
- then install dependencies (`RUN npm ci`)
- only then COPY the rest `COPY . ${APP_ROOT}`

This will let docker keep cache of the installed packages and not rerun `npm ci` in case `package-lock.json` hasn't changed

Also, [prefer `COPY` above `ADD`](https://stackoverflow.com/a/24958548)

frontend/Dockerfile
```diff
@@ -48,6 +48,8 @@ ARG DEFAULT_ETHEREUM_NETWORK
 ENV DEFAULT_ETHEREUM_NETWORK=${DEFAULT_ETHEREUM_NETWORK}
 ARG HEAPIO_ID
 ENV HEAPIO_ID=${HEAPIO_ID}
+ARG FRONTEND_URL
+ENV FRONTEND_URL=${FRONTEND_URL}
```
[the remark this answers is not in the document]
KirillDogadin-std: Sorry, i don't observe the line where it is supplied to the frontend. Could you point this?

or do you imply that one should make them supplied and then derive? I will go with this option for now.

frontend/cypress/plugins/index.js
```diff
@@ -0,0 +1,22 @@
+/// <reference types="cypress" />
```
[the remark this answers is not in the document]
KirillDogadin-std: those are autogenerated though, would you prefer them removed+ignored?

frontend/Dockerfile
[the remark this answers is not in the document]
valiafetisov: Hmm, true, they are not supplied in the dockerfile atm, maybe they are supplied at runtime? If you need to make it run in the docker, you will need to set `HOST` to `0.0.0.0`. 

But then, `0.0.0.0:3000` wouldn't be the correct url for cypress, right? Maybe it's confusing that e2e tests are part of the frontend, while actually not running frontend itself. I guess just naming `FRONTEND_URL` in a better way and adding it to readme with explanation would help. Eg `E2E_FRONTEND_URL`?

frontend/cypress/integration/basic-spec.ts
```diff
@@ -0,0 +1,66 @@
...
+    it('participates with swap and with direct bid', function () {
+        testSwapProfit();
+        testBidWithDai();
+        cy.window()
+            .its('$nuxt')
+            .then(function (nuxt) {
+                walletBalanceAfter = nuxt.$store.getters['wallet/walletBalances'];
+                // eslint-disable-next-line no-unused-expressions
+                expect(walletBalanceAfter.walletVatDAI.gt(walletBalanceBefore.walletVatDAI)).to.be.true;
+            });
```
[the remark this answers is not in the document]
KirillDogadin-std: That i've tried to tackle for quite some time. The reason it is done this way is that it 'at least works somehow':

i've searched for how to perform that properly. see the commit [6306b2b](https://github.com/sidestream-tech/unified-auctions-ui/pull/287/commits/6306b2b86e441c26df813db484714087f3277679) . This is beyond the point of being uncomfy now and starts to transfer into frustration:

- the problem is that the balance is not updated in the timely manner even if i invoke `wallet/fetchWalletBalances` via store before making the assertations. So with the non-upto-date balance the test fails.
- the execution is so slow and instable that i can't reach the point where i place the `debugger`;
  - moreover with the debugging being practically inpractical i can't even be sure that the first point holds true.
- ideally the check should be done in blockchain an that itself is a whole different task, so even if i spend more time here i do not see how it is a "good way"

I would appreciate an advice here on whether to/how to proceed

LukSteib:
> When opening cypress and logging in with my git credentials it seems that it's the default state without any tests yet.

Was able to start test via cypress UI and watch test execution. Ran into timeout. However, orally discussed with @KirillDogadin-std that this is expected. 

Running the e2e test via terminal is successful as already reported [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/287#pullrequestreview-1001387590). So from my side this is fine. 

Next step: Include e2e test into our CI. Which will be handled in separate PR.

valiafetisov:
> But then is there a way to interact with this UI (ie. observe the runs of our e2e tests)?

> However, orally discussed with @KirillDogadin-std that this is expected.

@KirillDogadin-std, can you please always answer those and further questions in a written form, so we have documented reasons, commands, etc at least in the pull requests, if those are not already specified in the code/readme.

valiafetisov:
Closing this since Cypress turns out to be very slow and unpredictable in the CI

PR #305, opened 2022-06-15
Added SurplusEventTable

### LINKED ISSUE #303: Create `SurplusEventsTable`
### Goal
The user has transparency on the different activity for the auction at hand via a dedicated `SurplusEventsTable`

### Context

This issue covers the creation of the `SurplusEventsTable`. It will later on be displayed on the auction page of one specific surplus auction. The table covers the basic events of a surplus auction. These are "start" , several "bid" events and a closing "collect" event

### Tasks
- [ ] Create the `SurplusEventsTable` (see mock below)
- [ ] Transaction column should include a clickable transaction hash that links to etherscan `https://etherscan.io/tx/[tx-hash]`
- [ ] type column -> string
- [ ] wallet -> wallet address with a clickable etherscan link `https://etherscan.io/address/[address]`
   - [ ] if the wallet address matches the wallet that is currently connected we will display "You"
- [ ] amount is empty for start and collect and contains number in MKR for the bid type entries
- [ ] When column provides counter that points back to the event  
- [ ] Create stories
   - [ ] error state in case there was a problem fetching the events
   - [ ] loading state

### Assets
`SurplusEventsTable`:

Broader context:

benediktkaiser:
Closes #303

Screenshot 2022-06-15 at 13 09 06

Checklist:

### DIFF
--- a/core/src/types.ts
+++ b/core/src/types.ts
@@ -153,10 +153,40 @@ export declare interface CollateralStatus {
     balance?: BigNumber;
 }
 
-export declare interface SurplusEvent {
-    transactionAddress?: string;
-    type: string;
-    walletAddress?: string;
-    amount?: BigNumber;
-    date: Date;
+export type SurplusAuctionStates =
+    | 'just-started'
+    | 'have-bids'
+    | 'ready-for-collection'
+    | 'collected'
+    | 'requires-restart';
+
+export type SurplusAuctionEventTypes = 'start' | 'bid' | 'collect';
+
+export declare interface SurplusAuctionData {
+    bidAmountMKR?: BigNumber;
+    receiveAmountDAI: BigNumber;
+    receiverAddress: string;
+    auctionEndDate: Date;
+    bidEndDate?: Date;
+    earliestEndDate: Date;
+    state: SurplusAuctionStates;
+}
+
+export declare interface SurplusEvent extends SurplusAuctionData {
+    type: SurplusAuctionEventTypes;
+    address?: string | null;
+    transactionHash: string;
+    transactionDate: Date;
+}
+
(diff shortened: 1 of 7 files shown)


fegbert (dismissed):

fegbert (dismissed):

LukSteib (changes requested):
- [ ] The "Default" story shows values in the "Amount" column for all event types. `start` and `collect` types however shouldn't have an MKR value
- [ ] For the indication on transaction execution we show rough time estimates (e.g. "about 5 hours ago) -> For this type of auctions specific times are more crucial. So I would prefer to always have detailed time indications like "XX hours YY minutes ZZ seconds ago"
- [x] Why do we show the sentence "This is a demo error message, no real error occurred." for the error state? Since this story should reflect error behavior in production let's display the actual error message. Going with "Error while retrieving events. " alone would be fine

valiafetisov:
Seeing the table in action, made me realise that we can combine `Type` and `Amount` columns to make it more readable:

MakerDAO surplus@2x (1)

benediktkaiser:
> * [ ]  The "Default" story shows values in the "Amount" column for all event types. `start` and `collect` types however shouldn't have an MKR value

From what @KirillDogadin-std told me, the start event is automatically also the first bid. Meaning that when someone starts the Auction, they must bid on it. I don't know if this information is correct or not. If we find out this is not true, then it can be easily changed. However this data is optional, meaning that is no amount is set, the Table will correctly reflect it.

benediktkaiser:
Screenshot 2022-06-20 at 12 45 50

Updated to new Design and changed generation of fake Auctions.

LukSteib:
Based on latest information regarding not being able to fetch history of events for an auction let's pause the work in this component for now. It is no longer part of the latest mock

Issue #188, opened 2022-04-05
Investigate surplus auctions logic


valiafetisov:
### Goal

Step-by-step explanation how it works, implementation proposal

### Context

Besides collateral auctions that are already supported by the UI and the keeper, we also want to support Maker surplus auctions ([or `flap` auctions using maker terminology](https://docs.makerdao.com/smart-contract-modules/system-stabilizer-module/flap-detailed-documentation)). It is currently supported by the https://auctions.makerdao.com/flap platform ([source code](https://github.com/makerdao/auctions-ui)), but our UI should eventually support all auction types. There are other projects like [flap-auctions-api](https://github.com/makerdao/flap-auctions-api) that may implement desired functionality in other languages that can also be used for reference.

The main goal of the issue is to investigate what smart contracts are responsible for those types of auctions, how existing tools deal with it. This should be done by documenting it in the issue using english terms (instead of maker terms).

For the example of this kind of research issue, you can have a look at the recent `Implement wallet-related logic` issue https://github.com/sidestream-tech/unified-auctions-ui/issues/124 and its investigation result https://github.com/sidestream-tech/unified-auctions-ui/issues/124#issuecomment-1064277892 as well as the PR https://github.com/sidestream-tech/unified-auctions-ui/pull/143 itself.

### Tasks
- [x] investigate how it's done, document here with links to the source code
  - [x] insure step-by-step process is documented including steps that are unknown at the moment (eg authorization step)
- [ ] Discuss the scope
- [ ] Propose clear implementation approach

KirillDogadin-std:
Here're some (ongoing) investigation records

contract name: `MCD_FLAP`  https://github.com/makerdao/auctions-ui/blob/master/constants.js#L39

Probably will need the following fucntions in the core for the feature:
    - transferMKRtoAddress - `tend` gets realised and the user can bet MKR on the auction
    - confirmWinningBid - `deal` is realised and the user can withdraw DAI from the auction
    - https://github.com/makerdao/auctions-ui/blob/master/hooks/useAuctionActions.js#L46-L62

additional functions that might be related.

https://github.com/makerdao/auctions-ui/blob/master/hooks/useAuctionActions.js#L46-L62
https://github.com/makerdao/auctions-ui/blob/master/hooks/useAllowances.js#L86-L96

So far I did not manage to figure out how the auth should look like in the upcoming implementation but just to keep track, here's some relevant info:  
[...]

PR #306, opened 2022-06-15
Added `SurplusAuctionsTable`

### LINKED ISSUE #302: Create `SurplusAuctionsTable`
### Goal
User is provided with surplus auctions via dedicated table

### Context

This issue covers the creation of the `SurplusAuctionTable`. Similar to the collateral auction table this table will later on be displayed on the main `/surplus` page

### Tasks
- [x] Create `SurplusAuctionTable` by adjusting the existing the existing `AuctionsTable`
- [x] Use the columns shown in the Mock
   - [x] Id -> integer 
   - [x] auction amount -> number in DAI
   - [x] highest bid -> number in MKR
   - [x] auction price -> number in DAI per MKR 
   - [x] market value -> percentage in relation (ie. with "above" or "below" indicator)
      - [x] twist to the mock: this time the "above" values should be marked in green
   - [x] state
      - [x] differentiate: "ends in [counter]", "expires in [counter]", "expired [counter]", "ended [counter]", "Collected [counter]"
   - [x] action column
       - [x] differentiate: "Participate", "Restart", "Collect DAI", "See details" 
       - [x] match the actions to the different states as shown in the mock 
- [x] Apply existing styling (hovering, greyed out rows if inactive)
   - [x] greyed out for the following action/states: Restart, See details 
   - [x] normal white for the following action/states: Participate, Collect DAI  
- [x] create stories 
   - [x] ensure that different states and actions are included
   - [x] include error state 
   - [x] include loading state  

### Assets
image

zoey-kaiser:
Closes #302 

Screenshot 2022-06-16 at 00 43 25

Checklist:


benediktkaiser:
I finished everything else we need for this PR. The default sorting is currently as follows:

- All Active Auctions (internally sorted by market price ascending)
- All Auctions that need a restart
- All Auctions ready for collection
- All Collected and finished Auctions

Is this a good order or do we need to change something @LukSteib?

I also added the ability to sort by Auction States. Do we want this or should I remove it?

frontend/helpers/generateFakeSurplusAuction.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    if (surplusAuctionData.state === 'collected') {
+        events.push(generateFakeSurplusEvent('collect'));
+    }
+
+    const latestBids = events.filter(event => {
+        return event.type === 'bid';
+    });
+    const highestBid = latestBids.length > 0 ? latestBids[latestBids.length - 1].bidAmountMKR : undefined;
```
benediktkaiser, line 58: I currently calculate the highest bid and auction price in the helper function. Meaning we will later need to recalculate this every X seconds, like with the step/cut price of normal auctions. 

Do we want to do this, or should I have this number be calculated inside the component. My thought process of calculating it externally was, that we want to show this data in other parts of our UI as well, meaning I personally belive it makes more sense to calculate it once, globally rather then multiple times in seperate components. However, this could also be changed if we belive it is better to do it on a component level.

core/src/types.ts
```diff
@@ -152,3 +152,41 @@ export declare interface CollateralStatus {
...
+    id: number;
+    network: string;
+    events: SurplusEvent[];
+}
+
+export declare interface SurplusAuction extends InitialSurplusAuction {
+    marketUnitPrice: BigNumber;
+    highestBid?: BigNumber;
+    auctionPrice?: BigNumber;
+}
```
KirillDogadin-std, line 192: regarding the hierarchy:

- i don't really understand why there's a duplicative information in various places:
  - the event has the auction data. although i would expect it to be the other way round: the events happen for the particular auction, so having the events listed as part of auction data makes more sense to me from the generic perspective.
  - if i think about it extensively: i fail to see the purpose of the initial auction here. Moreover i only can vaguely guess what "initial" implies in the naming.

benediktkaiser: > * the event has the auction data. although i would expect it to be the other way round: the events happen for the particular auction, so having the events listed as part of auction data makes more sense to me from the generic perspective.

I took this directly from @valiafetisov type mocking (https://github.com/sidestream-tech/unified-auctions-ui/issues/188#issuecomment-1152161469). I understand your point, however I was thinking that maybe there was a reason behind this. But I agree it is generally not needed and could be removed.

> * if i think about it extensively: i fail to see the purpose of the initial auction here. Moreover i only can vaguely guess what "initial" implies in the naming.

I modeled this after how we currently calculate "normal auctions". Initial implies that is only contains the data directly fetched from the blockchain. The full object then adds any additional data, we calculate on our end. In this case we would calculate, who the last bidder was and then we set the highest bid and auction price according to this data. We also do this with the normal auctions, when we calculate the next price drop according to the step/cut parameters, or add the market value (which is not initially included and needs to be enriched afterwards).

KirillDogadin-std: Verbal discussion result:

- dont use generics
- do not store auction data in the events
- store only relevant information in the events
- move the initial surplus aucion data pieces to the base interface: `id, network`

Rationales:

- most basic & simple, makes sense to not store auction data in the events.

benediktkaiser: We agreed on the following types in call:

```js

export declare interface GasParameters {
    maxFeePerGas?: string;
    maxPriorityFeePerGas?: string;
    gasPrice?: string;
}

export declare interface AuctionInitialInfo {
    network: string;
    id: string;
    index: number;
    collateralType: string;
    collateralSymbol: string;
    collateralAmount: BigNumber;
    debtDAI: BigNumber;
    startDate: Date;
    endDate: Date;
    vaultAddress: string;
    isActive: boolean;
    isFinished: boolean;
    isRestarting: boolean;
    marketUnitPrice?: BigNumber;
    marketUnitPriceToUnitPriceRatio?: BigNumber;
    transactionGrossProfit?: BigNumber;
    transactionAddress?: string;
    initialPrice: BigNumber;
    fetchedAt: Date;
}

export declare interface AuctionStatus {
    isActive: boolean;
    debtDAI: BigNumber;
    collateralAmount: BigNumber;
    unitPrice: BigNumber;
    totalPrice: BigNumber;
}

export declare interface Auction extends AuctionInitialInfo {
    unitPrice: BigNumber;
    totalPrice: BigNumber;
    approximateUnitPrice: BigNumber;
    secondsBetweenPriceDrops?: number;
    secondsTillNextPriceDrop?: number;
    priceDropRatio?: BigNumber;
    transactionGrossProfitDate?: Date;
    minimumBidDai: BigNumber;
}

export declare interface TransactionFees {
    bidTransactionFeeETH: BigNumber;
    bidTransactionFeeDAI: BigNumber;
    swapTransactionFeeETH: BigNumber;
    swapTransactionFeeDAI: BigNumber;
    authTransactionFeeETH: BigNumber;
    authTransactionFeeDAI: BigNumber;
    restartTransactionFeeETH: BigNumber;
    restartTransactionFeeDAI: BigNumber;
}

export declare interface CollateralRow extends CollateralConfig, Partial<MakerParams> {
    marketUnitPrice?: BigNumber | string;
    tokenAddress?: string;
    tokenAddressError?: string;
}

export declare interface AuctionTransaction extends Auction, TransactionFees {
    transactionNetProfit: BigNumber;
    combinedBidFeesDAI: BigNumber;
    combinedBidFeesETH: BigNumber;
    combinedSwapFeesDAI: BigNumber;
    combinedSwapFeesETH: BigNumber;
}

export declare interface RegularCalleeConfig {
    callee: 'UniswapV2CalleeDai' | 'WstETHCurveUniv3Callee' | 'CurveLpTokenUniv3Callee' | 'UniswapV3Callee';
    route: string[];
}

export declare interface UniswapV2LpTokenCalleeConfig {
    callee: 'UniswapV2LpTokenCalleeDai';
    token0: string;
    token1: string;
}

export declare interface CollateralConfig {
    title: string;
    ilk: string;
    symbol: string;
    decimals: number;
    exchange: RegularCalleeConfig | UniswapV2LpTokenCalleeConfig;
}

export declare interface NetworkConfig {
    chainId: string;
    title: string;
    url: string;
    gasPrice?: number;
    etherscanUrl: string;
    isFork: boolean;
}

export declare interface CalleeAddresses {
    UniswapV2CalleeDai: string;
    UniswapV2LpTokenCalleeDai: string;
    WstETHCurveUniv3Callee?: string;
    CurveLpTokenUniv3Callee?: string;
    UniswapV3Callee?: string;
}

export type CalleeNames = keyof CalleeAddresses;

export declare interface CalleeFunctions {
    getCalleeData: (network: string, collateral: CollateralConfig, profitAddress: string) => Promise<string>;
    getMarketPrice: (network: string, collateral: CollateralConfig, amount: BigNumber) => Promise<BigNumber>;
}

export declare interface MakerParams {
    secondsBetweenPriceDrops: number;
    priceDropRatio: BigNumber;
}

export declare interface MessageContent {
    key: string;
    content: string;
    duration: number;
}

export declare interface Notifier {
    (messageType: 'loading' | 'error' | 'success' | 'info', messageContent: MessageContent): void;
}

export declare interface TakeEvent {
    transactionHash: string;
    blockNumber: number;
    transactionDate?: Date;
}

export declare interface WalletBalances {
    walletETH: BigNumber;
    walletDAI: BigNumber;
    walletVatDAI: BigNumber;
    walletLastUpdatedDate: Date;
}

export declare interface CollateralStatus {
    type: string;
    symbol: string;
    isAuthorized: boolean;
    isAuthorizing: boolean;
    isDepositingOrWithdrawing: boolean;
    address?: string | null;
    balance?: BigNumber;
}

export type SurplusAuctionStates =
    | 'just-started'
    | 'have-bids'
    | 'ready-for-collection'
    | 'collected'
    | 'requires-restart';

export type SurplusAuctionEventTypes = 'start' | 'bid' | 'collect';

export declare interface SurplusAuctionData {
    id: number;
    network: string;
    receiveAmountDAI: BigNumber;
    receiverAddress: string;
    auctionEndDate: Date;
    bidEndDate?: Date;
    earliestEndDate: Date;
    state: SurplusAuctionStates;
}

export declare interface SurplusEvent {
    type: SurplusAuctionEventTypes;
    bidAmountMKR?: BigNumber;
    address: string;
    transactionHash: string;
    transactionDate: Date;
}

export declare interface InitialSurplusAuction extends SurplusAuctionData {
    events: SurplusEvent[];
}

export declare interface SurplusAuction extends InitialSurplusAuction {
    marketUnitPrice: BigNumber;
    highestBid?: BigNumber;
    auctionPrice?: BigNumber;
}

```

frontend/helpers/generateFakeSurplusAuction.ts
fegbert: I'm on the same side here. I think calculating inside components would just lead to duplicate code, and I don't see any upsides compared to calculating externally.

core/src/types.ts
LukSteib: >     receiverAddress: string;

NIT: I think the receiver address should be optional since for the state `just-started` (ie. auction without bids) we are not able to define it yet.

LukSteib (changes requested):
> Is this a good order or do we need to change something @LukSteib?

Yes, I like the default sorting

---
- [ ] why does this one collected auction have a different color coding (not hovered)? (see s1 below)
- [ ] NIT: For the state filter can we sort the selection options by the default sorting order (ie. just started, have bids, ready to collect, requires restart, collected)
- [ ] The auctions of the just started state in the default story do display a market difference. But as long as there are no bids no market difference can be displayed. Let's account for that (see s2 below)


s1:
image

s2: 
image

core/src/types.ts
valiafetisov: > I think the receiver address should be optional since for the state just-started (ie. auction without bids) we are not able to define it yet.

Technically, in this case, `receiverAddress` would contain the address of the wallet that kicked the auction, so it doesn't need to be optional. But maybe there is a better name for this field

benediktkaiser:
> * [ ]  why does this one collected auction have a different color coding (not hovered)? (see s1 below)

This comes from the fact, that it is the currently selected Auction. When we see this in the actual UI, this alternative color, will only be set when the Auction Panel is opened, to help indicate, which Auction from the table was selected. We have the same behaviour in the normal Auctions Table. 

> * [ ]  NIT: For the state filter can we sort the selection options by the default sorting order (ie. just started, have bids, ready to collect, requires restart, collected)

Sure!

> * [ ]  The auctions of the just started state in the default story do display a market difference. But as long as there are no bids no market difference can be displayed. Let's account for that (see s2 below)

I will change the helper functions!

core/src/types.ts
benediktkaiser: > Technically, in this case, `receiverAddress` would contain the address of the wallet that kicked the auction, so it doesn't need to be optional. But maybe there is a better name for this field

As discussed in the Daily, we need to investigate the return values from the events. @valiafetisov will post any updates on the investigation. Until then the types will stay the same. Once we have more information we will adjust them accordingly.

LukSteib (changes requested):
- [ ] We decided to change the denomination of the auction price to `MKR per DAI` -> can you please account for that change in the table
   - [ ] consequently the logic for market difference follows the old approach (green color coding for relative value below (and not the other way round)

frontend/helpers/generateFakeSurplusAuction.ts
valiafetisov: `bidAmountMKR` is actually the field that always contains the highest bid

Also, this PR should remove events as we had to change implementation proposal based on investigation.

benediktkaiser: Do we have any central place, where we defined the types? I did some looking and I saw that you updated the types in this issue: https://github.com/sidestream-tech/unified-auctions-ui/issues/188#issuecomment-1161592228

However, the types in the PR #317 are slightly different. Which one should I use?

valiafetisov: Feel free to use any of those, since ones in the #317 might not be final as well

benediktkaiser: What is missing in the types in #188 is the address of the highest bidder. We want to display in the UI if the current user is the highest bidder and to make this possible we need to store the address.

Is this technological possible?

LukSteib (approved):

fegbert (dismissed):

core/src/types.ts
```diff
@@ -159,19 +159,21 @@ export declare interface SurplusAuctionBase {
...
+    earliestEndDate?: Date;
 }
 
-export type SurplusAuction = SurplusAuctionActive | SurplusAuctionCollected;
 export type SurplusAuctionStates = SurplusAuction['state'];
+
+export declare interface SurplusAuctionTransaction extends SurplusAuction {
+    marketUnitPrice?: BigNumber | string;
+    marketUnitPriceToUnitPriceRatio?: BigNumber;
+    unitPrice?: BigNumber;
```
KirillDogadin-std, line 178: not sure what's the difference between this one and the `marketUnitPrice`. Could you clarify? or perhaps @LukSteib could?

frontend/components/utils/FormatMarketValue.vue
```diff
@@ -19,6 +21,10 @@ export default Vue.extend({
             type: [Number, Object] as Vue.PropType<Number | BigNumber>,
             default: undefined,
         },
+        inverted: {
```
KirillDogadin-std, line 24: i don't get the purpose of this variable. could you perhaps come up with a clearer name? (nit)

KirillDogadin-std (changes requested):

core/src/types.ts
zoey-kaiser: marketUnitPrice is how much 1 MKR costs on UniSwap
unitPrice is how much 1 MKR costs in the current bidding cycle. 

We use these two values to then calculate the marketUnitPriceToUnitPriceRatio. It helps the user see if the current bid is worth it.

frontend/components/utils/FormatMarketValue.vue
zoey-kaiser: This used to be how SurplusAuctions should work. Lukas had originally asked me to invert the colors (red and green) when displaying market difference, however this request we scrapped. I will remove this prop! 

Good catch!

core/src/types.ts
KirillDogadin-std: ok... so

if i buy 1 mkr for 2 dai on uniswap, and on auction i buy 1 mkr for 1,5 dai, then
```
    marketUnitPrice: 2
    marketUnitPriceToUnitPriceRatio: 1.42
    unitPrice: 1.5
```
right?

zoey-kaiser: This is the logic we currently use for collateral auctions:

```ts
const marketUnitPrice = await getMarketPrice(network, collateralSymbol, collateralAmount);
const marketUnitPriceToUnitPriceRatio = approximateUnitPrice
            .minus(marketUnitPrice)
            .dividedBy(marketUnitPrice);
``` 

UnitPrice is fetched from the auction directly, so I assume it should be the same as `totalAmount` / `totalPrice`.

Therefore in your example it should be as follows:

1.5 - 2 = -0.5
-0.5 / 2 = -0.25

Meaning that the current unit price of the auction is -0.25 under the price on Uniswap. I hope this explantation made a bit more sense 🤗

KirillDogadin-std: ok yes, thanks!

KirillDogadin-std (approved):

PR #307, opened 2022-06-15
Integrate `Surplus` into Frontend

### LINKED ISSUE #301: Implement /surplus page
### Goal
Basic layout of surplus auction page in place

### Context

With this issue we want to create create the basic page for interaction with surplus auctions similar but not identical to the existing `/collateral` page

### Tasks
- [ ] Create a new subpage for surplus auctions that will be available under the route `.../surplus` (see Mock below)
- [ ] Adjust existing header 
   - [ ] replace "Collateral auctions" with "Surplus Auctions"
   - [ ] logo should route to the `.../surplus` 
- [ ] use existing footer (no adjustment needed)
- [ ] Adjust existing `LandingBlock` 
   - [ ]  heading: "Surplus auctions" 
   - [ ] same buttons to show/hide explanations
- [ ] Include TextBlocks
   - [ ] "What is the Maker Protocol?" (can be re-used one to one)
   - [ ] "What are the surplus auctions?" (use lorem ipsum text for now -> text will be provided later on)
   - [ ] "Why would I participate?" (use lorem ipsum text for now -> text will be provided later on)
   - [ ] "What is the catch?" (use lorem ipsum text for now -> text will be provided later on)
- [ ] Ensure that TextBlocks can be shown/hidden with explanation toggle (similar to existing logic)

fegbert:
Closes #301 
Closes https://github.com/sidestream-tech/unified-auctions-ui/issues/340
Closes https://github.com/sidestream-tech/unified-auctions-ui/issues/346

I've also done some changes to the header in this PR, since we had to create a new layout file every time we needed something changed in the header, which then lead to layouts like `unified` and `minimal` whose only difference was a single prop of the header. Now the layout is deciding what to use depending on the page name.

**Please let me know if this approach makes sense / is better** 🙂 

Current categories are:
- Auctions Portal Pages - `/surplus` and `/collateral`, header name is based on page name
  - This one I'm not too sure about, since with this approach only these pages can have the respective name in the header. So for example if we add a subpage for the surplus actions, this would need to be changed.
  - The positive is that if we add for example debt auctions, nothing would need to be changed other than adding it to the list.
  - Alternative would be splitting in groups `surplus` and `collateral`
- Minimal - Only dark mode toggle
- Unified -  Dark & Expert mode toggles

Also added prop `title` to the `LandingBlock` for changing the text but keeping the buttons 

**preview**
image

Checklist:


frontend/components/surplus/SurplusText.vue
```diff
@@ -0,0 +1,70 @@
...
+            <Explain text="stable currency">
+                <a href="https://makerdao.com/en/whitepaper#the-dai-stablecoin" target="_blank">
+                    The DAI Stablecoin
+                </a>
+                is a decentralized, unbiased, collateral-backed cryptocurrency soft-pegged to the US Dollar </Explain
+            >, which can be used to borrow money over a longer period without being affected by unpredictable exchange
+            rates.
+        </TextBlock>
+        <TextBlock v-if="isExplanationsShown" title="What are surplus auctions?" class="TextBlock">
+            Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore
```
valiafetisov, line 21: @LukSteib, please provide texts for this page (I can also help with editing/proofreading)

frontend/components/surplus/SurplusFlow.vue
```diff
@@ -0,0 +1,58 @@
+<template>
+    <div :class="isStagingEnvironment ? 'SplitLayoutStagingContainer' : 'SplitLayoutContainer'">
+        <SplitLayout :step.sync="step">
```
valiafetisov, line 3: Does the page switching with the panel works already?

valiafetisov (commented):
In general, this issue will be the last one we merge, since if we merge it already, there will be a broken state on staging 🙁

aomafarag:
I addressed all requested changes, except for the split layout one, as I do not understand the context pertaining to its implementation, and whether or not I should modify something.

frontend/components/surplus/SurplusText.vue
```diff
@@ -3,43 +3,52 @@
         <TextBlock v-if="isExplanationsShown" title="What is the Maker Protocol?" class="TextBlock">
```
valiafetisov, line 3: Since we're reusing "What is the Maker Protocol?" text block and "What is the catch?", can you please turn them into a separate components and reuse between pages those two pages?

frontend/components/surplus/SurplusText.vue
```diff
@@ -3,43 +3,52 @@
...
-            sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam
-            voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea
-            takimata sanctus est Lorem ipsum dolor sit amet.
+            A
+            <a
+                href="https://docs.makerdao.com/smart-contract-modules/system-stabilizer-module/flap-detailed-documentation"
+                target="_blank"
+            >
+                surplus auction
+            </a>
```
valiafetisov, line 25: Let's not have links directly in the text, but only inside `Explain` texts. I would suggest to wrap `surplus auction` into `Explain` with the text: `In Maker terms surplus auctions are called [flap auctions](https://docs.makerdao.com/smart-contract-modules/system-stabilizer-module/flap-detailed-documentation) and are operated via the [flap.sol contract](https://github.com/makerdao/dss/blob/master/src/flap.sol)`

@LukSteib do you agree?

frontend/components/layout/Header.vue
```diff
@@ -119,6 +120,10 @@ export default Vue.extend({
             type: Boolean,
             default: false,
         },
+        pageName: {
+            type: String,
+            default: '',
+        },
```
valiafetisov, line 126: You created a new `pageName` prop, but you
- don't provide `this.$route?.name` to it via the parent
- didn't add a new story providing this prop

If you would mark this prop as required, you would immediately see the errors that would hint that something is off

valiafetisov (commented):
> as I do not understand the context pertaining to its implementation, and whether or not I should modify anything.

When you don't understand something, you should ask specific questions directly in the comment thread that asks for the change, [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/307#discussion_r914543278) in this case

frontend/components/surplus/SurplusFlow.vue
valiafetisov: `SplitLayout` is the layout that shows two different layouts/pages at the same time. Clicking a link in the surplus table should open a panel from the right sight. Currently it doesn't work and is not testable. Since this PR will be the last one to be merged, we should wait for both:
- table with surplus auctions
- surplus auction "page"
to be completed and merged to continue working on this PR

frontend/components/surplus/SurplusText.vue
LukSteib: Agreed

LukSteib (changes requested):
Reviewing based on flapper simulation: 
- [ ] I receive a fetch surplus auction error (see logs1 below)



<details><summary>logs1</summary>
<p>

```
fetch surplus auction error Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6","data":"0xf7729d430000000000000000000000009f8f72aa9304c8b593d555f12ef6589cc3a579a20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f0000000000000000000000000000000000000000000000000000000000000bb800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000","accessList":null}, error={"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":53,\"error\":{\"code\":-32603,\"message\":\"Error: VM Exception while processing transaction: reverted with reason string 'AS'\"}}","error":{"code":-32603},"requestBody":"{\"method\":\"eth_call\",\"params\":[{\"to\":\"0xb27308f9f90d607463bb33ea1bebb41c27ce5ab6\",\"data\":\"0xf7729d430000000000000000000000009f8f72aa9304c8b593d555f12ef6589cc3a579a20000000000000000000000006b175474e89094c44da98b954eedeac495271d0f0000000000000000000000000000000000000000000000000000000000000bb800000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\"},\"latest\"],\"id\":53,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","url":"http://127.0.0.1:8545"}, code=CALL_EXCEPTION, version=providers/5.6.8)
    at checkError (json-rpc-provider.js?9d4a:64:1)
    at StaticJsonRpcProvider.eval (json-rpc-provider.js?9d4a:533:1)
```

</p>
</details>

Reviewing based on hardhat:
- [ ] The time till auction end calculation seems to be off (see s1 below). An "... ago" indicator although auction is active (but probably due to hardhat, right?)
- [ ] the market difference displayed in overall auctions table on main page is different than the one displayed in single auction table
- [ ] the MKRAllowanceCheck panel is in disabled state (indicating no wallet connected) although a wallet is connected and the MKR in wallet is displayed properly

s1:
image

valiafetisov:
> The time till auction end calculation seems to be off (see s1 below). An "... ago" indicator although auction is active (but probably due to hardhat, right?)

Yes, that's the reason. The first part of the phrase is based on the auction state, which is determined based on the chain time. But the time is displayed based on the browser time. Won't fix it in this PR.

valiafetisov:
> I receive a fetch surplus auction error (see logs1 below)

There is a problem with market price fetching on the chaoslabs network, couldn't figure out why, fixed it by just not displaying it. Also, on this network no transactions (eg swap to WETH via metamask) are going through. Seems like a problem on their side.

valiafetisov:
Please also review normal auctions fetching/refetching/authorisations since this PR affects some of that logic

valiafetisov:
> only thing I am not able (not knowing how) to test: Collecting a finished auction where I am the highest bidder (ie. withdrawing DAI from VAT)

I've been able to test it via warping time, via curl, but also created a function:
```
await $nuxt.$store.dispatch('network/warpTime');
```

> question related to that: We do not show the DAI withdrawal panel on the auctionTransaction page. Intended?

Not intended. It's actually part of the mock, but wasn't added earlier to the transaction panel since we were not sure if it will be needed. We can add it next week during polishing, since withdrawal is still possible via the wallet popup, right?

LukSteib:
> We can add it next week during polishing, since withdrawal is still possible via the wallet popup, right?

Agree

LukSteib (approved):
Surplus auction test via hardhat: 
* texts looking good 
* auctions are fetched
* MKR balance fetched correctly 
* was able to give allowance
* was able to place a bid
  * lowest next possible bid updated accordingly
* was able to collect auction (after using timeWarp function)
* DAI ended up in VAT (checked via wallet Popup)
  * was able to withdraw DAI from VAT to wallet

 ---

Collateral auction test via hardhat:
* auth steps working
* swap execution working
* DAI deposit and withdraw working
* DAI bidding working
* collateral withdrawal from VAT working

This looks good to me - hence approved

PR #311, opened 2022-06-17
Added SurplusAuctionTransactionTable

### LINKED ISSUE #309: Create `SurplusAuctionTransactionTable`
### Goal
The user gets an overview on different transactions of one auction

### Context

This issue covers the creation of the `SurplusAuctionTransactionTable` . It will later on be displayed on the surplus auction transaction page inside one of the panels where the user will finally place a bid.

### Tasks
- [ ] Create the `SurplusAuctionTransactionTable`  (see mock below)
- [ ] Transaction column -> include a clickable transaction hash that links to etherscan `https://etherscan.io/tx/[tx-hash]`
   - [ ] note: compared to the `SurplusAuctionEventTable` this one only displays transactions of type `bid` (no start and collect) 
- [ ] Bidder column -> wallet address as a clickable etherscan link `https://etherscan.io/address/[wallet-address]`
   - [ ] if the bidder address matches the connected wallet we also display "(you)" after the link
- [ ] Amount column -> number in MKR
- [ ] When column -> counter that points back to the moment of the bid
- [ ] Create stories
   - [ ] error state in case there was a problem fetching the events
   - [ ] loading state
   - [ ] No data state (see Mock below)

### Assets
`SurplusAuctionTransactionTable` (with transactions):

`SurplusAuctionTransactionTable` (not data state):

Broader context:

benediktkaiser:
Closes #309.

Screenshot 2022-06-17 at 10 49 43

_Base forked from #303_ 

Checklist:


LukSteib (changes requested):
- [ ] Same as for events table. Let's be more detailed for the "When" column and show hours, minutes, seconds
- [x] "This is a demo error message, no real error occurred." Demo message sentence in error story not needed
- [x] The "With start and collect events" story doesn't make much sense, right?
- [x] Can we show the latest bid at the top of the table (ie. invert the current order)

benediktkaiser:
> * [ ]  The "With start and collect events" story doesn't make much sense, right?

This is more a story to test my filtering functionality. For the case that we accidentally also pass `start` and `collect`. This story ensures, that these events are correctly filtered out and not displayed in the table. 

> * Can we show the latest bid at the top of the table (ie. invert the current order)

I think this makes sense. Another idea of mine: What if I only show the last 5 bids and then. has a button at the bottom entitled: "View all bids", which then extends the table, showing all the previous bids. This:

a.) Makes the UI cleaner on first load, without removing any vital information
b.) Makes sure the user understands even clearer, that the top bid is the last one (as it would make even less sense to hide the latest bid)

benediktkaiser:
> * [x]  Same as for events table. Let's be more detailed for the "When" column and show hours, minutes, seconds

I added the same logic as I did in the Event Table. It now shows the exact time with seconds

> * [x]  The "With start and collect events" story doesn't make much sense, right?

I decided to remove this story, after verifying it still worked. It will probably cause more confusion, then that it will be helpful.

> [x] Can we show the latest bid at the top of the table (ie. invert the current order)

I inverted the order and also made the table cut of after the last 5 auctions. If this is the case, a new row is shown, which the user can click to expand and show all previous bids. The design and functionality is the same as with the Auctions Table we are already using. Is 5 a good cut off point or should I increase/decrease it?

Screenshot 2022-06-20 at 15 53 57

PR #314, opened 2022-06-17
Added `WalletMkrDepositCheckPanel`

### LINKED ISSUE #310: Create `WalletMkrDepsoitCheckPanel`
### Goal
The user is able to deposit MKR to the VAT if needed to make the desired bid

### Context

This issue covers the creation of the `WalletMkrDepsoitCheckPanel` . It follows the same structure as depositing of DAI via the wallet popup that was already implemented. 

### Tasks
- [x] Create `WalletMkrDepsoitCheckPanel` (see Mock below)
- [x] Account for the different states of the panel and sub-panels
   - [x] states are similar to the depositDaiPanel that we currently use for the wallet popup (states are also covered in the Mocks below)
   - [x] apply existing rule of collapsing and expanding based on panel state
- [x] Create stories and cover the different possible panel states
   - [x] Include states seen in mock 
   - [x] include unknown state if no wallet is connected (not in mock but also needed)
 
### Assets
MKR token address (mainnet): 0x9f8f72aa9304c8b593d555f12ef6589cc3a579a2

<details><summary>Mocks of `WalletMkrDepsoitCheckPanel`:</summary>
<p>

</p>
</details>

fegbert:
Closes #310 

- Added `WalletMkrDepositCheckPanel`, which basically checks the same thing as its DAI counterpart but shows different things (different text + wallet info table)
  - The upper inner panel
- Added a `currency` prop to the `AllowanceAmountCheckPanel` and changed its text to be more general
- Added `WalletVatMkrDepositCheckPanel`, which combines the two panels mentioned and, if all checks pass, allows to deposit to the VAT. 

**preview**

image

Checklist:


LukSteib (dismissed):

frontend/components/panels/WalletMkrDepositCheckPanel.vue
```diff
@@ -0,0 +1,152 @@
+<template>
```
valiafetisov, line 1: May I ask why have you reused `AllowanceAmountCheckPanel.vue` component, but not `WalletVatDaiBalanceCheckPanel.vue`? Can's you also simply `currency` prop there (and possibly rename the component)?

frontend/components/panels/WalletVatMkrDepositCheckPanel.vue
```diff
@@ -0,0 +1,152 @@
+<template>
```
valiafetisov, line 1: I would also generalise this component as well, based on the `currency` prop

frontend/components/panels/WalletVatMkrDepositCheckPanel.vue
```diff
@@ -0,0 +1,152 @@
...
+            :token-address-mkr="tokenAddressMkr"
+            :disabled="disabled"
+            :is-loading="isLoading"
+            :is-explanations-shown="isExplanationsShown"
+            @refresh="$emit('refresh')"
+        />
+        <AllowanceAmountCheckPanel
+            :is-correct.sync="isAmountWithinAllowance"
+            :allowance-amount="allowanceAmount"
+            :desired-amount="desiredAmount"
```
valiafetisov, line 19: Also based on the screenshot: here we should pass the same value as displayed in `Minimum to deposit` field

frontend/components/panels/WalletVatMkrDepositCheckPanel.vue
```diff
@@ -0,0 +1,152 @@
...
+            to be deposited there before the participation can happen.
+        </TextBlock>
+        <Button
+            type="primary"
+            class="w-full mt-2"
+            :disabled="disabled || !isAmountWithinAllowance || !isEnoughMkrInWallet || isEnoughDeposited"
+            :is-loading="isLoading"
+            @click="$emit('deposit')"
+        >
+            Deposit <format-currency :value="desiredAmount" currency="MKR" class="ml-1" />
```
valiafetisov, line 37: Same `Minimum to deposit` value should also be displayed here.
As well as emitted as value together with `deposit` event

LukSteib: Agree - missed that in my review :( 

Thx for pointing it out!

frontend/components/panels/WalletDepositCheckPanel.vue
```diff
@@ -0,0 +1,166 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+        <div class="flex flex-col gap-2">
+            <TextBlock v-if="currency === 'MKR'">
```
valiafetisov, line 5: Why does it only shown for the MKR? If it's about the table with details, you should better introduce `isTableShown` prop for example

frontend/components/panels/WalletVatDepositCheckPanel.stories.js
```diff
@@ -0,0 +1,133 @@
+import { storiesOf } from '@storybook/vue';
+import { action } from '@storybook/addon-actions';
+import faker from 'faker';
+import BigNumber from 'bignumber.js';
+import WalletVatDepositCheckPanel from './WalletVatDepositCheckPanel';
```
valiafetisov, line 5: Both components `WalletVatDepositCheckPanel` and `WalletDepositCheckPanel` deposit to VAT, so it's confusing naming. I propose
- `WalletVatCheckPanel` that checks
- `WalletDepositFlowCheckPanel` that also allow to execute the action

frontend/components/panels/WalletVatDepositCheckPanel.vue
```diff
@@ -0,0 +1,164 @@
...
+            return BigNumber.maximum(0, this.desiredAmount.minus(this.walletVatAmount));
+        },
+        isDesiredAmountValid(): boolean {
+            return this.desiredAmount && !this.desiredAmount.isNaN();
+        },
+        isEnoughDeposited(): boolean {
+            if (!this.walletVatAmount || !this.isDesiredAmountValid || this.desiredAmount.isLessThan(0)) {
+                return false;
+            }
+            return this.desiredAmount.isLessThan(this.walletVatAmount);
```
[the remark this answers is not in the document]
valiafetisov: I am not sure why do you compute it again if you already have `minimumDepositAmount` which you can check to be 0 here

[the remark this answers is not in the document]
fegbert: Yea, good question 😅

frontend/components/panels/WalletDepositFlowCheckPanel.vue
```diff
@@ -116,10 +116,7 @@ export default Vue.extend({
             return this.desiredAmount && !this.desiredAmount.isNaN();
         },
         isEnoughDeposited(): boolean {
-            if (!this.walletVatAmount || !this.isDesiredAmountValid || this.desiredAmount.isLessThan(0)) {
-                return false;
-            }
-            return this.desiredAmount.isLessThan(this.walletVatAmount);
+            return !this.minimumDepositAmount?.isGreaterThan(0);
```
valiafetisov, line 119: In case `this.minimumDepositAmount` is null, `!this.minimumDepositAmount?.isGreaterThan(0)` will evaluate into `true`. Is it expected? Otherwise I would suggest to use `.isZero()` without negation to avoid this edge case

frontend/components/panels/WalletDepositFlowCheckPanel.vue
```diff
@@ -0,0 +1,161 @@
...
+            if (!this.desiredAmount || !this.walletVatAmount) {
+                return undefined;
+            }
+            return BigNumber.maximum(0, this.desiredAmount.minus(this.walletVatAmount));
+        },
+        isDesiredAmountValid(): boolean {
+            return this.desiredAmount && !this.desiredAmount.isNaN();
+        },
+        isEnoughDeposited(): boolean {
+            return this.minimumDepositAmount?.isZero() || false;
```
valiafetisov, line 119: In the future, please prefer using [`??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) when dealing with null produced by `?.`

valiafetisov (approved):

PR #317, opened 2022-06-21
Fetch surplus auction by id

### LINKED ISSUE #295: Create function to fetch auction by index
### Goal
Fetch surplus auction information based on index

### Context
The issue is a subset of the implementation proposal you can find [here](https://github.com/sidestream-tech/unified-auctions-ui/issues/188#issuecomment-1152161469)

### Tasks
- [ ] Create new file in core `surplus.ts`
- [ ] Create function to fetch auction by index `fetchSurplusAuctionByIndex(network, index)`
    - [ ] fetch events related to the auction, filtering by its index
    - [ ] returns renamed and transformed auction with events array
    - [ ]  in case the `currentState === 'just-started'`, we should additionally fetch `flap.kicks(index)` to check that the auction was not restarted (because calling `flap.tick(index)` [resets `auctionEndDate`, but emits no events](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/flap.sol#L133-L137)) – this should also be fetched after the restart transaction execution

### Assets

 Proposal:
```typescript
// NB: this is only a proposal
interface SurplusAuctionData {
    bidAmountMKR: BigNumber | undefined; // amount of MRK (bid)
    receiveAmountDAI: BigNumber; // amount of DAI to be received (lot)
    receiverAddress: string; // the address of the wallet who receives DAI (guy)
    auctionEndDate: Date; // final date when auction will end (now + `flap.tau()` – 2 days atm) (end)
    bidEndDate: Date; // (tic)
    earliestEndDate: Date; // earliest of auctionEndDate and auctionEndDate
    state: 'just-started' | 'have-bids' | 'ready-for-collection' | 'collected' | 'requires-restart'; // auction state determined from fetched events
}
interface SurplusBid extends SurplusAuctionData {
    type: 'start' | 'bid' | 'collect'; // as 'Kick', 'Tend', 'Deal'
    transactionHash: string;
    transactionDate: Date; // determined from block number (see `fetchDateByBlockNumber`)
    // other SurplusAuctionData fields
}
interface SurplusAuction extends SurplusAuctionData {
    events: SurplusEvent[];
}
```

KirillDogadin-std:
Closes #295 

Checklist:

___

the events and state of the auction are a bit in the different shape as initially intended because of the complications with the event extraction from the blockchain and due to the fact that the surplus auction start event has to be extracted anyway to gain the complete info.

manual test was happening on block `14078340` and with the latest auction `2328` (the greatest `kick` value)

### DIFF
--- a/core/src/contracts.ts
+++ b/core/src/contracts.ts
@@ -14,8 +14,6 @@ import WSTETH from './abis/WSTETH.json';
 import getSigner from './signer';
 import memoizee from 'memoizee';
 
-const CACHE_EXPIRY_MS = 24 * 60 * 60 * 1000;
-
 export const getClipperNameByCollateralType = function (collateralType: string): string {
     const suffix = collateralType.toUpperCase().replace('-', '_');
     return `MCD_CLIP_${suffix}`;
@@ -70,7 +68,6 @@ const _getContract = async function (network: string, contractName: string, useS
 };
 
 const getContract = memoizee(_getContract, {
-    maxAge: CACHE_EXPIRY_MS,
     promise: true,
     length: 3,
 });
--- a/core/src/contracts.ts
+++ b/core/src/contracts.ts
@@ -9,6 +9,7 @@ import MCD_JOIN from './abis/MCD_JOIN.json';
 import MCD_CLIP_CALC from './abis/MCD_CLIP_CALC.json';
 import MCD_CLIP from './abis/MCD_CLIP.json';
 import MCD_DOG from './abis/MCD_DOG.json';
+import MCD_FLAP from './abis/MCD_FLAP.json';
 import WSTETH from './abis/WSTETH.json';
 import getSigner from './signer';
 
@@ -35,6 +36,9 @@ const getContractInterfaceByName = async function (contractName: string): Promis
     }
     if (contractName === 'MCD_DOG') {
         return MCD_DOG;
+    }
+    if (contractName === 'MCD_FLAP') {
+        return MCD_FLAP;
     }
     if (contractName === 'WSTETH') {
         return WSTETH;
--- a/core/src/contracts.ts
+++ b/core/src/contracts.ts
@@ -12,6 +12,9 @@ import MCD_DOG from './abis/MCD_DOG.json';
 import MCD_FLAP from './abis/MCD_FLAP.json';
 import WSTETH from './abis/WSTETH.json';
 import getSigner from './signer';
+import memoizee from 'memoizee';
+
+const CACHE_EXPIRY_MS = 24 * 60 * 60 * 1000;
 
 export const getClipperNameByCollateralType = function (collateralType: string): string {
     const suffix = collateralType.toUpperCase().replace('-', '_');
@@ -58,12 +61,18 @@ const getContractInterfaceByName = async function (contractName: string): Promis
     throw new Error(`No contract interface found for "${contractName}"`);
 };
 
-const getContract = async function (network: string, contractName: string, useSigner = false): Promise<Contract> {
+const _getContract = async function (network: string, contractName: string, useSigner = false): Promise<Contract> {
     const contractAddress = await getContractAddressByName(network, contractName);
     const contractInterface = await getContractInterfaceByName(contractName);
     const signerOrProvider = useSigner ? await getSigner(network) : await getProvider(network);
     const contract = await new ethers.Contract(contractAddress, contractInterface, signerOrProvider);
     return contract;
 };
+
+const getContract = memoizee(_getContract, {
+    maxAge: CACHE_EXPIRY_MS,
+    promise: true,
+    length: 3,
+});
 
 export default getContract;
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -1,4 +1,4 @@
-import type { SurplusAuction, SurplusAuctionStates, SurplusAuctionUnexistant, SurplusAuctionBaseData } from './types';
+import type { SurplusAuction, SurplusAuctionUnexistant, SurplusAuctionBaseData } from './types';
 import { getEarliestDate } from './helpers/getEarliestDate';
 import BigNumber from './bignumber';
 import getContract from './contracts';
@@ -7,13 +7,35 @@ import { Contract } from 'ethers';
 const surplusAuctionLastIndex = async (contract: Contract) => {
     const auctionsQuantityBinary = await contract.kicks();
     return new BigNumber(auctionsQuantityBinary._hex);
-}
+};
+
+const getAuctionState = (earliestEndDate: Date, greatestBid: number) => {
+    const isBidExpired = new Date() > earliestEndDate;
+    const haveBids = greatestBid === 0;
+    if (haveBids) {
+        if (isBidExpired) {
+            return 'ready-for-collection';
+        }
+        return 'have-bids';
+    }
+    if (isBidExpired) {
+        return 'requires-restart';
+    }
+    return 'just-started';
+};
 
 export const fetchSurplusAuctionByIndex = async function (
     network: string,
-    auctionIndex: number
+    auctionIndex: number,
+    contractOrNull: Contract | null = null
 ): Promise<SurplusAuction | SurplusAuctionUnexistant> {
-    const contract = await getContract(network, 'MCD_FLAP');
+    let contract: Contract;
+    if (!!contractOrNull) {
+        contract = contractOrNull;
+    } else {
+        contract = await getContract(network, 'MCD_FLAP');
+    }
+
     const auctionData = await contract.bids(auctionIndex);
     const isAuctionDeleted = new BigNumber(auctionData.end).eq(0);
     const baseAuctionInfo: SurplusAuctionBaseData = {
@@ -22,24 +44,17 @@ export const fetchSurplusAuctionByIndex = async function (
     };
 
     if (isAuctionDeleted) {
-        const auctionLastIndex = await surplusAuctionLastIndex(contract)
+        const auctionLastIndex = await surplusAuctionLastIndex(contract);
         if (auctionLastIndex.lt(auctionIndex)) {
             throw new Error('No active auction found with this id');
         }
-        return { ...baseAuctionInfo, state: 'collected'};
+        return { ...baseAuctionInfo, state: 'collected' };
     }
 
     const auctionEndDate = new Date(auctionData.end * 1000);
     const bidEndDate = new Date(auctionData.tic * 1000);
     const earliestEndDate = getEarliestDate(auctionEndDate, bidEndDate);
-
-    const isBidExpired = new Date() > earliestEndDate;
-
-    const haveBids = auctionData.tic === 0;
-    const [stateIfExpired, stateIfNotExpired] = haveBids
-        ? ['ready-for-collection', 'have-bids']
-        : ['requires-restart', 'just-started'];
-    const state: SurplusAuctionStates = (isBidExpired ? stateIfExpired : stateIfNotExpired) as SurplusAuctionStates;
+    const state = getAuctionState(earliestEndDate, auctionData.tic);
 
     return {
         ...baseAuctionInfo,
@@ -51,4 +66,30 @@ export const fetchSurplusAuctionByIndex = async function (
         bidEndDate,
         state,
     };
+};
+
+const getActiveSurplusAuctionOrNull = async (
+    network: string,
+    auctionIndex: number,
+    contractOrNull: Contract | null = null
+) => {
+    const auction = await fetchSurplusAuctionByIndex(network, auctionIndex, contractOrNull);
+    if (auction.state === 'collected') {
+        return null;
+    }
+    return auction;
+};
+
+export const fetchActiveSurplusAuctions = async function (network: string): Promise<SurplusAuction[]> {
+    const contract = await getContract(network, 'MCD_FLAP');
+    const auctionLastIndex = (await surplusAuctionLastIndex(contract)).toNumber();
+
+    let auctionIndexToFetch = auctionLastIndex;
+    let surplusAuctions: SurplusAuction[] = [];
+    let currentSurplusAuction;
+    while ((currentSurplusAuction = await getActiveSurplusAuctionOrNull(network, auctionIndexToFetch, contract))) {
+        surplusAuctions.push(currentSurplusAuction);
+        --auctionIndexToFetch;
+    }
+    return surplusAuctions;
 };
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -41,7 +41,7 @@ export const fetchSurplusAuctionByIndex = async function (
     if (isAuctionDeleted) {
         const auctionLastIndex = await getSurplusAuctionLastIndex(contract);
         if (auctionLastIndex < auctionIndex) {
-            throw new Error('No active auction exists this id');
+            throw new Error('No active auction exists with this id');
         }
         return { ...baseAuctionInfo, state: 'collected' };
     }
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -1,4 +1,4 @@
-import type { SurplusAuction, SurplusAuctionCollected, SurplusAuctionBase } from './types';
+import type { SurplusAuction, SurplusAuctionBase } from './types';
 import { getEarliestDate } from './helpers/getEarliestDate';
 import BigNumber from './bignumber';
 import getContract from './contracts';
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -4,6 +4,7 @@ import BigNumber from './bignumber';
 import getContract from './contracts';
 import { Contract } from 'ethers';
 import getNetworkDate from './date';
+import { RAD, WAD } from './constants/UNITS';
 
 const getSurplusAuctionLastIndex = async (contract: Contract): Promise<number> => {
     const auctionsQuantityBinary = await contract.kicks();
@@ -60,8 +61,8 @@ export const fetchSurplusAuctionByIndex = async function (
     return {
         ...baseAuctionInfo,
         earliestEndDate,
-        bidAmountMKR: new BigNumber(auctionData.bid._hex),
-        receiveAmountDAI: new BigNumber(auctionData.lot._hex),
+        bidAmountMKR: new BigNumber(auctionData.bid._hex).div(WAD),
+        receiveAmountDAI: new BigNumber(auctionData.lot._hex).div(RAD),
         receiverAddress: auctionData.guy,
         auctionEndDate,
         bidEndDate,
@@ -69,7 +70,7 @@ export const fetchSurplusAuctionByIndex = async function (
     };
 };
 
-const getActiveSurplusAuctionOrNull = async (network: string, auctionIndex: number, contract?: Contract) => {
+const getActiveSurplusAuctionOrUndefined = async (network: string, auctionIndex: number, contract?: Contract) => {
     const auction = await fetchSurplusAuctionByIndex(network, auctionIndex, contract);
     if (auction.state === 'collected') {
         return;
@@ -81,11 +82,10 @@ export const fetchActiveSurplusAuctions = async function (network: string): Prom
     const contract = await getContract(network, 'MCD_FLAP');
     const auctionLastIndex = await getSurplusAuctionLastIndex(contract);
 
-    const auctionIndexToFetch = auctionLastIndex;
     const surplusAuctions: SurplusAuction[] = [];
     let currentSurplusAuction;
     for (let i = auctionLastIndex; i > 0; i--) {
-        currentSurplusAuction = await getActiveSurplusAuctionOrNull(network, auctionIndexToFetch, contract);
+        currentSurplusAuction = await getActiveSurplusAuctionOrUndefined(network, i, contract);
         if (!currentSurplusAuction) {
             break;
         }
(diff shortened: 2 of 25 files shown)


core/src/surplus.ts
```diff
@@ -0,0 +1,52 @@
...
+    network: string,
+    auctionIndex: number
+): Promise<SurplusAuctionData> {
+    const contract = await getContract(network, 'MCD_FLAP');
+
+    const auctionsQuantityBinary = await contract.kicks();
+    const auctionsQuantity = new BigNumber(auctionsQuantityBinary._hex);
+    if (auctionsQuantity.lt(auctionIndex)) {
+        throw new Error('No active auction found with this id');
+    }
```
valiafetisov, line 16: I don't think you need to do extra call to check again that index doesn't exceed `kicks`. You can at least move few lines down, after you already know that there is no auction with this id and you need to check if auction never existed. 

I would also move `auctionLastIndex` logic into a separate function, since we will need to reuse it later on

core/src/surplus.ts
```diff
@@ -0,0 +1,52 @@
...
+import getContract from './contracts';
+
+export const fetchSurplusAuctionByIndex = async function (
+    network: string,
+    auctionIndex: number
+): Promise<SurplusAuctionData> {
+    const contract = await getContract(network, 'MCD_FLAP');
+
+    const auctionsQuantityBinary = await contract.kicks();
+    const auctionsQuantity = new BigNumber(auctionsQuantityBinary._hex);
```
valiafetisov, line 13: I would say it's the `auctionLastIndex`

core/src/surplus.ts
```diff
@@ -0,0 +1,52 @@
...
+    if (auctionsQuantity.lt(auctionIndex)) {
+        throw new Error('No active auction found with this id');
+    }
+
+    const auctionData = await contract.bids(auctionIndex);
+    const isAuctionDeleted = new BigNumber(auctionData.lot).eq(0);
+    const baseAuctionInfo: SurplusAuctionData = {
```
valiafetisov, line 20: - I would suggest to maybe have a different type for the collected auction? Otherwise, types doesn't really make sense as most of the parameters are optional. Would it make sense?
- I would check that `end` date is 0 instead of `lot` to be on the safe side

core/src/surplus.ts
```diff
@@ -0,0 +1,52 @@
...
+    const state: SurplusAuctionStates = (isBidExpired ? stateIfExpired : stateIfNotExpired) as SurplusAuctionStates;
+
+    return {
+        ...baseAuctionInfo,
+        earliestEndDate,
+        bidAmountMKR: new BigNumber(auctionData.bid._hex),
+        receiveAmountDAI: new BigNumber(auctionData.lot._hex),
+        receiverAddress: auctionData.guy,
+        auctionEndDate: expirationTimeAuction,
+        bidEndDate: expirationTimeBid,
```
valiafetisov, line 49: Do you want to rename the type or the vars for consistency?

core/src/surplus.ts
```diff
@@ -0,0 +1,52 @@
...
+    const expirationTimeBid = new Date(auctionData.tic * 1000);
+    const earliestEndDate = getEarliestDate(expirationTimeAuction, expirationTimeBid);
+
+    const isBidExpired = new Date() > earliestEndDate;
+
+    const haveBids = auctionData.tic === 0;
+    const [stateIfExpired, stateIfNotExpired] = haveBids
+        ? ['ready-for-collection', 'have-bids']
+        : ['requires-restart', 'just-started'];
+    const state: SurplusAuctionStates = (isBidExpired ? stateIfExpired : stateIfNotExpired) as SurplusAuctionStates;
```
valiafetisov, line 40: Can we please make it a little more readable? I think simple `if/else` would work here. You can also move it to a separate function eg `determineAuctionState`

I also think that typecasting via `as` would be enough for ts to figure out the type.

valiafetisov (commented):
I would suggest to also implement fetching of all active auctions in the same PR since it may influence the structure of your functions

valiafetisov:
I would also advice to add store-related actions+storage to frontend in the same PR, so that its functionally testable

KirillDogadin-std:
> How is it related to this PR?

the proposal of the datastructure for the auctions includes market price. So asking in case they have to be also included here.

core/src/surplus.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export const fetchSurplusAuctionByIndex = async function (
+    network: string,
+    auctionIndex: number,
+    contractOrNull: Contract | null = null
+): Promise<SurplusAuction | SurplusAuctionUnexistant> {
+    let contract: Contract;
+    if (!!contractOrNull) {
+        contract = contractOrNull;
+    } else {
```
valiafetisov, line 35: - Why do you need to pass contract as a parameter to this function?
- When and why can contract become null?

In short, you don't need to provide it via parameter, please use `getContract` directly

core/src/surplus.ts
```diff
@@ -0,0 +1,95 @@
...
+    const isAuctionDeleted = new BigNumber(auctionData.end).eq(0);
+    const baseAuctionInfo: SurplusAuctionBaseData = {
+        network,
+        id: auctionIndex,
+    };
+
+    if (isAuctionDeleted) {
+        const auctionLastIndex = await surplusAuctionLastIndex(contract);
+        if (auctionLastIndex.lt(auctionIndex)) {
+            throw new Error('No active auction found with this id');
```
valiafetisov, line 49: I would make it clear that auction with this id never existed in the first place, not just not found?

core/src/surplus.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const getActiveSurplusAuctionOrNull = async (
+    network: string,
+    auctionIndex: number,
+    contractOrNull: Contract | null = null
+) => {
+    const auction = await fetchSurplusAuctionByIndex(network, auctionIndex, contractOrNull);
+    if (auction.state === 'collected') {
+        return null;
+    }
```
valiafetisov, line 79: Please prefer using `undefined` over `null` in our codebase. null should only be used if you need to have extra state, distinct from `undefined`
```suggestion
    if (auction.state === 'collected') {
        return;
    }
```

core/src/surplus.ts
```diff
@@ -0,0 +1,95 @@
...
+    const contract = await getContract(network, 'MCD_FLAP');
+    const auctionLastIndex = (await surplusAuctionLastIndex(contract)).toNumber();
+
+    let auctionIndexToFetch = auctionLastIndex;
+    const surplusAuctions: SurplusAuction[] = [];
+    let currentSurplusAuction;
+    while ((currentSurplusAuction = await getActiveSurplusAuctionOrNull(network, auctionIndexToFetch, contract))) {
+        surplusAuctions.push(currentSurplusAuction);
+        --auctionIndexToFetch;
+    }
```
valiafetisov, line 93: I think it would be more readable with a `for` loop and a `break`

core/src/types.ts
```diff
@@ -152,3 +152,29 @@ export declare interface CollateralStatus {
     address?: string | null;
     balance?: BigNumber;
 }
+
+export type SurplusAuctionStates = 'just-started' | 'have-bids' | 'ready-for-collection' | 'requires-restart';
+
+export type SurplusAuctionEventTypes = 'start' | 'bid' | 'collect';
```
valiafetisov, line 158: Please remove unused types

frontend/store/auctions.ts
```diff
@@ -358,6 +359,11 @@ export const actions = {
             commit('setAreTakeEventsFetching', false);
         }
     },
+    async fetchSurplusAuctions({ rootGetters }: ActionContext<State, State>) {
+        const network = rootGetters['network/getMakerNetwork'];
+        const auctions = await fetchActiveSurplusAuctions(network);
+        return auctions;
+    },
```
valiafetisov, line 366: Let's create another `surplus` store instead and not mix collateral with surplus auctions

core/src/surplus.ts
```diff
@@ -0,0 +1,95 @@
+import type { SurplusAuction, SurplusAuctionUnexistant, SurplusAuctionBaseData } from './types';
+import { getEarliestDate } from './helpers/getEarliestDate';
+import BigNumber from './bignumber';
+import getContract from './contracts';
+import { Contract } from 'ethers';
+
+const surplusAuctionLastIndex = async (contract: Contract) => {
+    const auctionsQuantityBinary = await contract.kicks();
+    return new BigNumber(auctionsQuantityBinary._hex);
```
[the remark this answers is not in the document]
KirillDogadin-std: > why doesn't ts require you typing output of the function?

it can infer it apparently.

core/src/surplus.ts
KirillDogadin-std: if the contract has already been fetched on the previous level ( the function that fetches all the active auction ), then re-getting the contract every time seems inefficient operation. that's why here it's optional and is set if was not provided.

KirillDogadin-std: still changed the arg defenition to the ts format

core/src/surplus.ts
KirillDogadin-std: well, my preference is to the while but i dont mind changing

core/src/contracts.ts
```diff
@@ -54,12 +61,18 @@ const getContractInterfaceByName = async function (contractName: string): Promis
...
+const _getContract = async function (network: string, contractName: string, useSigner = false): Promise<Contract> {
     const contractAddress = await getContractAddressByName(network, contractName);
     const contractInterface = await getContractInterfaceByName(contractName);
     const signerOrProvider = useSigner ? await getSigner(network) : await getProvider(network);
     const contract = await new ethers.Contract(contractAddress, contractInterface, signerOrProvider);
     return contract;
 };
 
+const getContract = memoizee(_getContract, {
+    maxAge: CACHE_EXPIRY_MS,
```
valiafetisov, line 73: Does it need to expire?

core/src/types.ts
```diff
@@ -152,3 +152,27 @@ export declare interface CollateralStatus {
...
+    auctionEndDate: Date;
+    bidEndDate: Date;
+    earliestEndDate: Date;
+}
+
+export declare interface SurplusAuction extends SurplusAuctionFullData {
+    state: SurplusAuctionStates;
+}
+
+export declare interface SurplusAuctionUnexistant {
```
valiafetisov, line 176:
```suggestion
export declare interface SurplusAuctionCollected {
```
?

core/src/types.ts
```diff
@@ -152,3 +152,27 @@ export declare interface CollateralStatus {
...
+    receiveAmountDAI: BigNumber;
+    receiverAddress: string;
+    auctionEndDate: Date;
+    bidEndDate: Date;
+    earliestEndDate: Date;
+}
+
+export declare interface SurplusAuction extends SurplusAuctionFullData {
+    state: SurplusAuctionStates;
+}
```
valiafetisov, line 174: Should `SurplusAuction` be `SurplusAuctionBase | SurplusAuctionCollected`?

KirillDogadin-std:
To manually test:

1. set the value of `blockNumber` in `hardhat/hardhat.config.js` to `14078340`.
2. run the frontend
3. run the hardhat
4. open browser, visit the app
5. open the console in the browser, 
6. call `auc = await this.$nuxt.$store.dispatch('surplus/fetchSurplusAuctions')`
7. wait....
8. something pops up in the console, interact and see the values
9. you can select specific objects in the list via `auc[<your-desired-index>]`
10. if the value is of type `BigNumber` such as the `bidAmountDai` you can do `auc[<your-desired-index>].bidAmountDai.toNumber()`

core/src/types.ts
```diff
@@ -152,3 +152,27 @@ export declare interface CollateralStatus {
     address?: string | null;
     balance?: BigNumber;
 }
+
+export type SurplusAuctionStates = 'just-started' | 'have-bids' | 'ready-for-collection' | 'requires-restart';
```
valiafetisov, line 156: If you export this type, `SurplusAuctionStates` should also include `collected`, I believe. Otherwise, you can maybe do it the other way around and derive it from `SurplusAuction.state`?

core/src/surplus.ts
```diff
@@ -0,0 +1,88 @@
...
+    const isAuctionDeleted = new BigNumber(auctionData.end).eq(0);
+    const baseAuctionInfo: SurplusAuctionBaseData = {
+        network,
+        id: auctionIndex,
+    };
+
+    if (isAuctionDeleted) {
+        const auctionLastIndex = await getSurplusAuctionLastIndex(contract);
+        if (auctionLastIndex < auctionIndex) {
+            throw new Error('No active auction exists this id');
```
valiafetisov, line 44:
```suggestion
            throw new Error('No active auction exists with this id');
```
?

core/src/surplus.ts
```diff
@@ -0,0 +1,88 @@
...
+        receiveAmountDAI: new BigNumber(auctionData.lot._hex).div(RAD),
+        receiverAddress: auctionData.guy,
+        auctionEndDate,
+        bidEndDate,
+        state,
+    };
+};
+
+const getActiveSurplusAuctionOrUndefined = async (network: string, auctionIndex: number) => {
+    const auction = await fetchSurplusAuctionByIndex(network, auctionIndex);
```
valiafetisov, line 67: Can there be a case that it throws an error?

frontend/store/surplus.ts
```diff
@@ -0,0 +1,38 @@
...
+    takeEventStorage: {},
+    areAuctionsFetching: false,
+    isSelectedAuctionFetching: false,
+    areTakeEventsFetching: false,
+    isBidding: false,
+    error: null,
+    auctionErrors: {},
+    restartingAuctionsIds: [],
+    lastUpdated: undefined,
+});
```
valiafetisov, line 29: - Please remove unused properties
- You can already save auctions into `auctionStorage` and provide a getter

core/src/types.ts
KirillDogadin-std: I do not understand what you're suggesting here exactly.

core/src/surplus.ts
KirillDogadin-std: Within this context there're no errors i am aware of that we expect and have to handle.

core/src/types.ts
valiafetisov: I suggest to have one main `SurplusAuction` type that is a tuple of all other types:
```ts
export declare interface SurplusAuctionBase {
    id: number;
    network: string;
}

export declare interface SurplusAuctionCollected extends SurplusAuctionBase {
    state: 'collected';
}

export declare interface SurplusAuctionActive extends SurplusAuctionBase {
    state: 'just-started' | 'have-bids' | 'ready-for-collection' | 'requires-restart';
    bidAmountMKR: BigNumber;
    receiveAmountDAI: BigNumber;
    receiverAddress: string;
    auctionEndDate: Date;
    bidEndDate: Date;
    earliestEndDate: Date;
}

export type SurplusAuction = SurplusAuctionActive | SurplusAuctionCollected;

export type SurplusAuctionStates = SurplusAuction['state'];
```
This will 
- let us use the same type everywhere and let typescript make most of the work (check that if you access `receiverAddress`, you had checked that `state` is not `collected`)
- also make types directly reusable (you can use `SurplusAuctionCollected` in places where you don't expect generic auction)

core/src/surplus.ts
valiafetisov: 1. `fetchSurplusAuctionByIndex` throws an error
2. underlying provider may also throw and error on the network timeout

frontend/store/surplus.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const getters = {
+    auctionStorage(state: State) {
+        return state.auctionStorage;
+    },
+};
+
+export const mutations = {
+    addAuctionToStorage(state: State, auction: SurplusAuctionFullData) {
+        state.auctionStorage[auction.id] = auction;
```
valiafetisov, line 23: Adding new key to the store wouldn't reactively propagate the changes. You need to either:
- use `Vue.set` (preferred) https://github.com/sidestream-tech/unified-auctions-ui/blob/d295a3e6879b6b49348d3e5e8c84865e4444af29/frontend/store/auctions.ts#L125
- replace the whole object https://github.com/sidestream-tech/unified-auctions-ui/blob/d295a3e6879b6b49348d3e5e8c84865e4444af29/frontend/store/auctions.ts#L119-L122

core/src/types.ts
valiafetisov: This comment hasn't been addressed (despite being liked)

core/src/surplus.ts
valiafetisov: > Usually (to my xp) there's something one has to do if the error rises.

Yes, my question was, if there is an edge case where this error expected

core/src/surplus.ts
```diff
@@ -0,0 +1,88 @@
...
+    if (isBidExpired) {
+        return 'requires-restart';
+    }
+    return 'just-started';
+};
+
+export const fetchSurplusAuctionByIndex = async function (
+    network: string,
+    auctionIndex: number
+): Promise<SurplusAuction | SurplusAuctionCollected> {
```
valiafetisov, line 32: As suggested above, this function should return `SurplusAuction` which itself is a `SurplusAuctionActive | SurplusAuctionCollected`

core/src/surplus.ts
```diff
@@ -0,0 +1,88 @@
...
+    const contract = await getContract(network, 'MCD_FLAP');
+    const auctionData = await contract.bids(auctionIndex);
+    const isAuctionDeleted = new BigNumber(auctionData.end).eq(0);
```
valiafetisov, line 35:
```suggestion
    const isAuctionCollected = new BigNumber(auctionData.end).eq(0);
```

frontend/store/surplus.ts
```diff
@@ -0,0 +1,35 @@
...
+export const state = (): State => getInitialState();
+
+export const getters = {
+    auctionStorage(state: State) {
+        return state.auctionStorage;
+    },
+};
+
+export const mutations = {
+    addAuctionToStorage(state: State, auction: SurplusAuctionFullData) {
```
valiafetisov, line 23: Wouldn't auction added to the storage contain `state`? Then, the type should be `SurplusAuction`

frontend/store/surplus.ts
```diff
@@ -0,0 +1,35 @@
...
+        Vue.set(state.auctionStorage, auction.id, auction);
+    },
+};
+
+export const actions = {
+    async fetchSurplusAuctions({ rootGetters, commit }: ActionContext<State, State>) {
+        const network = rootGetters['network/getMakerNetwork'];
```
valiafetisov, line 30: You should check if `network` is not undefined before calling the rest

core/src/types.ts
KirillDogadin-std: oops, wrong branch, cherry picked.

core/src/types.ts
```diff
@@ -152,3 +152,25 @@ export declare interface CollateralStatus {
...
+}
+
+export declare interface SurplusAuctionActive extends SurplusAuctionBase {
+    bidAmountMKR: BigNumber;
+    receiveAmountDAI: BigNumber;
+    receiverAddress: string;
+    auctionEndDate: Date;
+    bidEndDate: Date;
+    earliestEndDate: Date;
+    state: 'just-started' | 'have-bids' | 'ready-for-collection' | 'requires-restart' | 'collected';
```
valiafetisov, line 168: The `SurplusAuctionActive` can not have state `collected`

core/src/surplus.ts
```diff
@@ -0,0 +1,88 @@
...
+        const auctionLastIndex = await getSurplusAuctionLastIndex(contract);
+        if (auctionLastIndex < auctionIndex) {
+            throw new Error('No active auction exists with this id');
+        }
+        return { ...baseAuctionInfo, state: 'collected' };
+    }
+
+    const auctionEndDate = new Date(auctionData.end * 1000);
+    const bidEndDate = new Date(auctionData.tic * 1000);
+    const earliestEndDate = getEarliestDate(auctionEndDate, bidEndDate);
```
valiafetisov, line 51: Also, `bidEndDate` [is not set](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/flap.sol#L117-L132) during kick, therefore, `bidEndDate` will be 0. Instead of creating `new Date(0)` which will be `Thu Jan 01 1970 01:00:00`, you should set `bidEndDate` to undefined.

That's why I proposed to have `earliestEndDate` in the first place: to figure out this complex edge case once

valiafetisov (dismissed):

LukSteib:
Based on review procedure [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/317#issuecomment-1164173388) step 8 provided me sometimes the following error (but re-trying also yielded the desired outcome):
`Uncaught Error: No contract address found for "MCD_FLAP"`


Other than that looking good to me
* was able fetch 5 surplus auctions via console command
* was able to see details of the auctions

Some observations / assumptions:
* all auctions have the same `receiverAddress` specified -> This address points to a smart contract (could be possible that this was an active keeper back in the days)
* 4 of the auctions are in state `ready-for-collection` -> since the specified block datetime is after the fetched `bidEndTime` this makes sense and is possible in my eyes
* I have a feeling that for execution of `tend()` function there is no need to deposit MKR in the `VAT` beforehand (this needs to be validated)

LukSteib (approved):
Based on review I'd approve this PR - still some unknowns that we need to figure out.

PR #318, opened 2022-06-23
Added CollectSurplusAuction Panel

### LINKED ISSUE #316: Create `CollectSurplusAuctionPanel`
### Goal
The user is can trigger the collection of an auction via a dedicated panel

### Context

This issue covers the creation of the `CollectSurplusAuctionPanel`. Later on it will be used on the surplus transaction page and facilitate the collection of DAI. 
Context: Once an auction is finished and no more bid can be placed, it needs to be collected. In the process of collection the amount of DAI that was auctioned off is transferred to the highest bidding wallet and the MKR is burned. Collection can be triggered by every wallet and is not restricted to be done by the wallet that made the highest bid

### Tasks
- [x] Create `CollectSurplusAuctionPanel` (see Mock below) 
- [x] Account for the different states of the panel
   - [ ] apply existing rule of collapsing and expanding based on panel state
- [ ] Create stories and cover the different possible panel states

zoey-kaiser:
Closes #316.

Screenshot 2022-06-23 at 07 51 50

Checklist:


frontend/components/surplus/CollectSurplusAuctionPanel.vue
```diff
@@ -0,0 +1,109 @@
...
+            {{ currentStateAndTitle.title }}
+            <time-till v-if="auction.state === 'have-bids'" :date="auction.earliestEndDate" />
+        </template>
+        <TextBlock v-if="auction.state === 'ready-for-collection'">
+            Since the end of the auction at {{ auction.earliestEndDate.toUTCString() }}, the winner can now collect the
+            auction amount.
+        </TextBlock>
+        <TextBlock v-else>
+            If no one else bids on the auction, the winner be able to collect the auction amount when the auction is
+            ended after {{ auction.earliestEndDate.toUTCString() }}
```
LukSteib, line 13: Here I would be more general in the auction text and write: 
"If no one else bids on the auction, one will be able to collect the auction amount when the auction is ended after {{ auction.earliestEndDate.toUTCString()"
(background: basically everybody will be able to execute `deal()` function. Not only the winner

frontend/components/surplus/CollectSurplusAuctionPanel.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>
+            {{ currentStateAndTitle.title }}
+            <time-till v-if="auction.state === 'have-bids'" :date="auction.earliestEndDate" />
+        </template>
+        <TextBlock v-if="auction.state === 'ready-for-collection'">
+            Since the end of the auction at {{ auction.earliestEndDate.toUTCString() }}, the winner can now collect the
+            auction amount.
```
LukSteib, line 9: Here I would adjust the text and also make a distinction between the case where the user == winner and user !== winner. 

For the case user == winner I propose: 
"Since the auction ended at {{ auction.earliestEndDate.toUTCString() }}, you can now collect the auction amount."

For the case user !== winner I propose:
"Since the auction ended at {{ auction.earliestEndDate.toUTCString() }}, you can now collect the auction amount. Note however that the auction amount will be transferred to the highest bidder's wallet."

LukSteib (changes requested):

frontend/components/surplus/CollectSurplusAuctionPanel.vue
```diff
@@ -0,0 +1,109 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>
+            {{ currentStateAndTitle.title }}
+            <time-till v-if="auction.state === 'have-bids'" :date="auction.earliestEndDate" />
+        </template>
+        <TextBlock v-if="auction.state === 'ready-for-collection'">
```
LukSteib, line 7: I would show this text for the collected auction state as well instead of the other text

fegbert:
Please update the branch from `main`, since the surplus types here are outdated so make sure to use the new ones. 
This will sadly result in some more errors that you have to fix since the outdated types are used all over the place 🙁

fegbert:
Yep, that's correct. So only take what's coming from the main branch and remove any changes made by this one please 🙂

LukSteib (changes requested):
- [x] In the `ready-for-collection` case the button needs to be clickable (currently disabled)

fegbert:
Also, another nice thing to do:
If you have addressed a comment, please add a reaction to it so its easier to see what is addressed already and what isn't 🙂

frontend/components/surplus/CollectSurplusAuctionPanel.vue
zoey-kaiser: I am not sure what you mean with this. Which text should I display/change? 😞

LukSteib (approved):

fegbert (approved):

KirillDogadin-std (approved):

PR #322, opened 2022-06-27
Extended AuctionRestartPanel for use on SurplusAuctions

### LINKED ISSUE #319: Create new `SurplusAuctionRestartPanels`
### Goal
The user is can trigger the restart of an auction 

### Context

This issue covers the creation of the `SurplusAuctionRestartPanels`. This is similar to the restart panel that we already have in place for the collateral auctions. Only texts need to be adjusted

### Tasks
- [x] Create `SurplusAuctionRestartPanels` (see Mock below for the restart panel that already exist)
- [x] Change text of existing restart panels (see comment 1 below) 
- [x] Account for the different states of the panel
   - [x] apply existing rule of collapsing and expanding based on panel state
- [x] Create stories and cover the different possible panel states

 
### Assets
image

aomafarag:
Closes #319 

sa-screenshot

Checklist:


aomafarag (commented):
I get a weird error with virtually all Vue components, all JSX-related. For example, here's the one with BasePanel:

'BasePanel' cannot be used as a JSX component.
  Its instance type 'CombinedVueInstance<{ isExpanded: boolean; } & { titleClass: string | undefined; iconClass: string | undefined; } & { currentState: string; } & Vue, object, object, object, Record<never, any>>' is not a valid JSX element.ts(2786)

frontend/components/panels/SurplusAuctionRestartPanels.stories.js
```diff
@@ -0,0 +1,88 @@
+import { storiesOf } from '@storybook/vue';
+import faker from 'faker';
+import SurplusAuctionRestartPanel from '~/components/panels/SurplusAuctionRestartPanel';
```
fegbert, line 3: You currently get an error for this line:
```
Could not locate module ~/components/panels/SurplusAuctionRestartPanel
```

That's because your files are actually called `SurplusAuctionRestartPanels`, so plural. Please rename both the component and the story file to just `SurplusAuctionRestartPanel` 🙂

LukSteib:
Please add top level description for this PR
See https://github.com/sidestream-tech/unified-auctions-ui/pull/306 as an example how to do it

LukSteib (approved):

frontend/components/panels/SurplusAuctionRestartPanel.vue
```diff
@@ -0,0 +1,78 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name" class="SurplusAuctionRestartPanel">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+
+        <TextBlock v-if="isExplanationsShown">
+            Auction that was started but hasn't received a single bid in the allotted time. When restarted bids can be
+            placed again.
+        </TextBlock>
```
fegbert, line 8: Since this is the only actual difference from the original `AuctionRestartPanel` I think it might be the better idea to merge these into one component and select the respective `TextBlock` with a new prop `auctionType` like we did in #334

LukSteib: would pass this question to @valiafetisov since it is code base related...

fegbert (approved):

LukSteib (approved):

fegbert (approved):

PR #323, opened 2022-06-28
Interact surplus

KirillDogadin-std:
Closes # .

Checklist:


KirillDogadin-std:
How to test:

1. set hardhat block number to 14078340
2. run the frontend, hardhat
3. navigate to the frontend in the browser.
4. use the metamask to get some mkr:
  - swap testeth into weth (around 1000)
  - swap weth into mkr (around 20).
  - if any error like `Swap failed` appears:
    - in the step named _review swap_ click "edit" next to "max fee" section.
    - go to advanced settings, adjust the max fee to ridiculously high value.
    - make sure to avoid 'nonce to high' error (you can see if it is the case in the hardhat log). Fix this according to the documentation (README)

5. use the console in the browser to utilize the functionality:
```
- `await this.$nuxt.$store.dispatch('surplus/fetchSurplusAuctions')` to see the auctions available.
- `await this.$nuxt.$store.dispatch('surplus/betOnAuction', {auctionIndex: <YOURINDEX>, bet: <YOURBET>})` to bet on the auction
- `await this.$nuxt.$store.dispatch('surplus/giveAllowanceMKR')` to allow operations with your MKR
- `await this.$nuxt.$store.dispatch('surplus/collectAuction', <YOURINDEX>)` to collect / end the auction
- `await this.$nuxt.$store.dispatch('surplus/restartAuction', <YOURINDEX>)` to restart the auction
- `await this.$nuxt.$store.dispatch('surplus/authorizeSurplusAuctions')` to authorize surplus auctions
- `await this.$nuxt.$store.dispatch('surplus/deauthorizeSurplusAuctions')` to unauthorize surplus auctions

```

LukSteib:
Currently when trying to connect with MetaMask I receive the following error:

```
Uncaught (in promise) TypeError: window.ethereum.isConnected is not a function
    at Function.get (eea33d7.js:1:1389812)
    at eea33d7.js:1:1379266
    at j (1d4298d.js:2:72248)
    at Generator._invoke (1d4298d.js:2:72036)
    at forEach.t.<computed> [as next] (1d4298d.js:2:72428)
    at r (1d4298d.js:2:133)
    at l (1d4298d.js:2:336)
    at 1d4298d.js:2:395
    at new Promise (<anonymous>)
    at 1d4298d.js:2:276
```

I've double checked and sadly also get this error in production. 
To add even more confusion: Trying to connect wallet on production with brave browser works

KirillDogadin-std:
> Currently when trying to connect with MetaMask I receive the following error:

extracted it into #326

KirillDogadin-std:
- writing a test that bets on the auction requires having MKR
  - this means we need to swap the currecncy into the mkr
  - https://github.com/sidestream-tech/unified-auctions-ui/pull/323/commits/7098ef74ec4ad5af5f7d09a2304de93666dbaa97 has the debugging state where one can use store to call the swapping functionality.

- swapping via uniswap fails (see commit linked above)
  - there's no error-message provided
  - the contract exists on the block i'm running it on
  - the reason to why it fails is riddling.
    - I assume that the permissions are not an issue because 'approve' is called for the uniswap so that it could access weth.
    - amount of weth that the my testing setup (via frontend + vuex store function invocation) is also apparently sufficient (1000).

- the testing of the programmatic swapping was done on frontend/vuex because:
  - debugging in the `npm run test` results into an informative "failed to estimate gas" error that can (according to the docs) also occur for non-gas related reasons.
  - one can control the balance of the wallet & see transaction logs easier with metamask.

KirillDogadin-std:
> I've heard that you were able to bid and collect surplus auction already

that is true. That works during the manual test. Now i'm introducing the integration test that is automated. For that the automated integration has to receive currency. That has to be done via swapping. Since integration test won't have metamask, i'm forced to implement alternative helper function for that.

> Swapping via what worked?

Metamask

> What means fails then in this case? Nothing is thrown

no, the `message` of the error is `Transaction reverted without a reason string`

valiafetisov:
> Transaction reverted without a reason string

Most probably you're calling non-existent method/using wrong ABI/providing invalid input types.

>> Swapping via what worked?
> 
> Metamask

You can also use contracts/methods that worked by inspecting the successful exchange transaction using etherscan or [tenderly](https://dashboard.tenderly.co/explorer) and figuring out what methods was called with what parameters (and comparing with the ones you're sending)

KirillDogadin-std:
> Most probably you're calling non-existent method/using wrong ABI/providing invalid input types.

There's `path` argument that specifies the addresses of the swapped tokens. One of those has to be WETH. If i provide different address from WETH in the `path` parameter, the error with corresponding text is returned. I think this experiment invalidates this hypothesis.

KirillDogadin-std:
> You can also use contracts/methods that worked by inspecting the successful exchange transaction

making transaction myself apparenly won't work because the metamask extension does not allow it on test nets like goerli and kovan

valiafetisov:
> making transaction myself apparenly won't work because the metamask extension does not allow it on test nets like goerli and kovan

- You can enable that in [advanced settings](https://user-images.githubusercontent.com/3393626/177141656-f929d7b8-4ce9-43ca-9fd9-9fef33f12dd8.png)
- Or you can execute it on the fork and fetch transaction receipt using ethers.js, ie `getTransactionReceipt(transactionHash)` (more involved than option above)

KirillDogadin-std:
> You can enable that

LukSteib:
> @LukSteib , @valiafetisov there was a discussion about having the testing setup for restarting the auctions which never happened before. Was there any progress?

So when it comes to simulations we already have the basic setup to work with the simulations locally see [here](https://github.com/sidestream-tech/unified-auctions-ui/tree/main/chaos-proxy)

Now dedicated simulations for surplus auctions have been added. 

I've updated the secrets for simulations accordingly. You can find the in `auction-ui/localhost/choas-proxy` chamber. New simulation is called `Single Flapper` 
Wanted to test locally on my side but received a `522` error when starting the simulation. I have a feeling that the access token we are currently using might be expired. 
@valiafetisov can you check and confirm?

core/test/helpers/swap.ts
```diff
@@ -0,0 +1,75 @@
+import getSigner from '../../src/signer';
```
valiafetisov, line 1: I actually think it's fine to include this functionality to the core itself (as we might even provide this feature later on to the user). Just please be more specific with naming ABIs, eg `UNISWAP.json` -> `UNISWAP_V2_ROUTER_02.json` (using the contract name)

valiafetisov:
> @valiafetisov , the tests are working through relative paths - not a big deal as for me, should i address it nevertheless?

Maybe adding this to the `core/tsconfig.json` this can help?
https://github.com/sidestream-tech/unified-auctions-ui/blob/948f56f8af844403aadd0f50f26b8ecf957ea783/bot/tsconfig.json#L20-L27

> @LukSteib , @valiafetisov there was a discussion about having the testing setup for restarting the auctions which never happened before. Was there any progress?

I was referencing the possibility to execute [`hardhat_mine`](https://hardhat.org/hardhat-network/reference#hardhat_mine) on the hardhat network to mine enough blocks for the auction to become expired

frontend/store/surplus.ts
```diff
@@ -35,4 +49,66 @@ export const actions = {
         auctions.forEach(auction => commit('addAuctionToStorage', auction));
         return auctions;
     },
+    async giveAllowanceMKR({ rootGetters }: ActionContext<State, State>) {
```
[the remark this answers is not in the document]
KirillDogadin-std: yeah ok, when mom takes a look it suddenly starts working without any additional adjustments.

KirillDogadin-std:
> Maybe adding this to the core/tsconfig.json this can help?

sadly does not work out

the added diff to the tsconfig was

```
        "baseUrl": ".",
        "paths": {
            "~/*": [
                "./*"
            ],
            "@/*": [
                "./*"
            ]
        }
```

valiafetisov (commented):
- [ ] Please provide instructions + store commands for functional tests in the top PR comment

core/test/surplus-test.ts
```diff
@@ -0,0 +1,100 @@
...
+            params: [
+                {
+                    forking: {
+                        jsonRpcUrl: REMOTE_RPC_URL,
+                        blockNumber: 14078338,
+                    },
+                },
+            ],
+        });
+        await hre.network.provider.send('hardhat_mine', ['0x5010', '0x10c']);
```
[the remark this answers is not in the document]
KirillDogadin-std: > Should we make a helper from that thing?

i don't think it's necessary as long as the thing is only used once.

the args are adjusted

core/src/network.ts
```diff
@@ -31,6 +31,22 @@ const SUPPORTED_NETWORKS: NetworkConfig[] = [
     },
 ];
 
+export const getDevelopmentNetworkConig = function (alsoSetState = false): NetworkConfig {
```
valiafetisov, line 34: Since you've used `setupRpcUrlAndGetNetworks` and now only refer to `custom` network, I understand that this function can be deleted, right?

core/src/helpers/swap.ts
```diff
@@ -0,0 +1,75 @@
...
+    const gasParameters = await getGasParametersForTransaction(network);
+    let transactionPromise = contractWeth['approve'](
+        ...[UNISWAP_ADDRESS, new BigNumber(amountPaidAllowanceETH).shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0)],
+        {
+            ...gasParameters,
+            type: gasParameters.gasPrice ? undefined : 2,
+        }
+    );
+
+    await trackTransaction(transactionPromise, notifier, canTransactionBeConfirmed(network));
```
valiafetisov, line 48: Why do you need to use internal `trackTransaction` here reimplementing gas, canTransactionBeConfirmed and the rest instead of just using `executeTransaction`?

Otherwise, please add `executeTransactionWithValue` to the `execute.ts` or extend `executeTransaction` if it's possible

core/src/surplus.ts
```diff
@@ -86,3 +87,48 @@ export const fetchActiveSurplusAuctions = async function (network: string): Prom
...
+    return restartedAuction;
+};
+
+export const restartSurplusAuction = async function (
+    network: string,
+    auctionIndex: number,
+    notifier?: Notifier
+): Promise<SurplusAuction> {
+    await executeTransaction(network, 'MCD_FLAP', 'tick', [auctionIndex], notifier);
+    return await surplusAuctionRestartedCallback(network, auctionIndex);
```
valiafetisov, line 107: I don't think that refetching should be part of the `restart` function (and you also don't use this functionality in the store) so please remove it here as well as `surplusAuctionRestartedCallback`

core/src/surplus.ts
```diff
@@ -86,3 +87,48 @@ export const fetchActiveSurplusAuctions = async function (network: string): Prom
...
+    const transactionParameters = [
+        auctionIndex,
+        auction.receiveAmountDAI.shiftedBy(RAD_NUMBER_OF_DIGITS).toFixed(0),
+        new BigNumber(bet).shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0),
+    ];
+    await executeTransaction(network, 'MCD_FLAP', 'tend', transactionParameters, notifier);
+};
+
+export const collectSurplusAuction = async function (network: string, auctionIndex: number, notifier?: Notifier) {
+    const auction = await getActiveSurplusAuctionOrUndefined(network, Number(auctionIndex));
```
valiafetisov, line 129: Why do you need to `Number(auctionIndex)` if its type is already `number`?

core/test/surplus-test.ts
```diff
@@ -0,0 +1,105 @@
...
+
+const REMOTE_RPC_URL = process.env.REMOTE_RPC_URL;
+const HARDHAT_PRIVATE_KEY = '0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80'; // deterministic private key from hardhat.
+const HARDHAT_FORK_BLOCK_NUMBER = 14078339;
+
+async function createWalletFromPrivateKey(privateKey: string, network: string) {
+    setSigner(network, createSigner(network, privateKey));
+    const signer = await getSigner(network);
+    return await signer.getAddress();
+}
```
valiafetisov, line 25: This can be moved to `signer.ts` if you find it useful, so a) you will need to import only one thing b) it can be reused later on

core/test/surplus-test.ts
```diff
@@ -0,0 +1,105 @@
...
+        expect(auctionAfterCollection.state).to.equal('collected');
+    });
+    it('restarts the auction', async () => {
+        await hre.network.provider.request({
+            method: 'hardhat_reset',
+            params: [
+                {
+                    forking: {
+                        jsonRpcUrl: REMOTE_RPC_URL,
+                        blockNumber: 14078338,
```
valiafetisov, line 80: Should it be `HARDHAT_FORK_BLOCK_NUMBER`?

KirillDogadin-std: `HARDHAT_FORK_BLOCK_NUMBER - 1`** :)

core/src/surplus.ts
```diff
@@ -86,3 +87,37 @@ export const fetchActiveSurplusAuctions = async function (network: string): Prom
     }
     return surplusAuctions;
 };
+
+export const restartSurplusAuction = async function (
+    network: string,
+    auctionIndex: number,
+    notifier?: Notifier
+): Promise<void> {
+    await executeTransaction(network, 'MCD_FLAP', 'tick', [auctionIndex], {}, notifier);
```
valiafetisov, line 96: Then you can skip it from here:
```suggestion
    await executeTransaction(network, 'MCD_FLAP', 'tick', [auctionIndex], undefined, notifier);
```

Althrough, if we're refactoring this function, maybe it's time to provide all non-required arguments via `options` object? Then you can:
```suggestion
    await executeTransaction(network, 'MCD_FLAP', 'tick', [auctionIndex], { notifier });
```

core/src/surplus.ts
```diff
@@ -86,3 +87,37 @@ export const fetchActiveSurplusAuctions = async function (network: string): Prom
...
+};
+
+export const bidToSurplusAuction = async function (
+    network: string,
+    auctionIndex: number,
+    bet: string,
+    notifier?: Notifier
+) {
+    const auction = await getActiveSurplusAuctionOrUndefined(network, Number(auctionIndex));
```
valiafetisov, line 105: Anther leftover
```suggestion
    const auction = await getActiveSurplusAuctionOrUndefined(network, auctionIndex);
```

valiafetisov (approved):

LukSteib (commented):
Can you give me a quick primer on how to set variables in the different `.env` files. 

So for the `/hardhat` related I am currently using the alchemy `RPC_URL`.
For the frontend I am using `http://127.0.0.1:8545/` as `RPC_URL`.
Is there anything I need to change? Are there any other required env variables that I need to set?

---
Currently running in the following error which hinders me to execute the console commands:
```
Uncaught (in promise) Error: could not detect network (event="noNetwork", code=NETWORK_ERROR, version=providers/5.6.8)
    at StaticJsonRpcProvider.eval (json-rpc-provider.js?9d4a:400:1)
```

The network selector in the UI remains in spinning "Switching..." mode

LukSteib (approved):

KirillDogadin-std:
>  do you want to have a look into why fetching auth status provided me with false? 

As discussed verbally - it's due to cache. For now we can skip it

PR #330, opened 2022-07-01
RPC_URL support

valiafetisov:
Closes https://github.com/sidestream-tech/unified-auctions-ui/issues/222
Replaces https://github.com/sidestream-tech/unified-auctions-ui/pull/284

Checklist:


valiafetisov:
@LukSteib frontend functionality is ready and can be tested already, the only outstanding parts:
- [ ] update bot to support RPC_URL
- [ ] add appropriate secrets for development, staging and production

LukSteib:
Tested frontend functionality:

Provided infura URL for the new `RPC_URL` and was able to 
* switch networks
* restart an auction on `goerli`
* bid on an auction on `goerli`
-> looking good ✅

Provided alchemy URL for the new `RPC_URL`
* In UI prompted to switch to custom network (see S1 below) -> however no action was triggered on buttonclick
* When trying to access a network like `goerli` directly via specifing in URL, network modal showed `Chain 0x1` as only supported network (see S2 below) -> however switching to this network failed when metamask wallet is connected and shows modal again
  * created a dedicated alchemy ethereum mainnet network via metamask (see S3 below) but still was not able to connect to the `custom` identified network

S1:
image

S2:
image

S3
image

valiafetisov:
> In UI prompted to switch to custom network (see S1 below) -> however no action was triggered on buttonclick

We are now matching `chainId` returned from the `RPC_URL` with our default configs (eg if the `RPC_URL` returns `0x1`, we treat this network as `mainnet` in our UI) but since it can be misleading (in case the `RPC_URL` points to a fork or a fake network) and there is no way to validate the opposite, I decided to display those networks as `-like` (eg `Goerli-like`):

Screenshot 2022-07-04 at 12 09 33

In another PR we can add support for custom networks (with unknown chainIds, like simulation networks) and automatic determination if it's a fork (get time of the latest block, compare it with browser time or check if [hardhat's rpc methods](https://hardhat.org/hardhat-network/reference#hardhat-network-methods) exist). For now, unknown networks work only for fetching, participation via UI is not possible.

> created a dedicated alchemy ethereum mainnet network via metamask (see S3 below) but still was not able to connect to the custom identified network

That's very surprising since previously metamask did not allow to add new networks with the same `chainId` as default ones (but maybe it is only blocked to automatic addition via metamask API? Will investigate in the dedicated issue).

valiafetisov:
Bot is also testable now via `RPC_URL="http://127.0.0.1:8545"` (and via others)

valiafetisov:
> and participation does not work?

No,

- `RPC_URL` from infura works and feels as before
- `RPC_URL` from other providers
  - if `RPC_URL` returns supported chainId (main, kovan, goerli), we mark it as `...-like` and support participation
  - if `RPC_URL` returns `0x539` (localhost), we support participation
  - if `RPC_URL` returns unsupported chainId, we mark it as `Chain 0x...` and doesn't support participation for now (always display "not supported" popup)

LukSteib:
Tested bot:
Works fine for both `mainnet` and `goerli` with my Alchemy RPC URL ✔️ 

> RPC_URL from other providers

I like the distinction that you make here already. For me this is fine in the first version. We can still improve later on.
Regarding the `...-like` annotation. Like the idea to somehow indicate that we cannot be 100% sure whether this is not a fork. Can we ensure that if somebody is providing an Infura URL? Just to be consistence, if not then we might move away from the `...-like` connotation completely.

valiafetisov:
> Can we ensure that if somebody is providing an Infura URL?

Technically not, but it is ensured by the infura reputation. It's just a common mistake when someone is doing the transfer of real money being on the fork or wise versa, so I think it's important to make it visible: even if `Mainnet` is displayed we are not able to verify it.

LukSteib:
> It's just a common mistake when someone is doing the transfer of real money being on the fork or wise versa, so I think it's important to make it visible: even if Mainnet is displayed we are not able to verify it.

Agree

LukSteib (approved):

deusavalon:
@valiafetisov the corresponding secret path can be found in the individual `values_override.xxxx` values file like [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/f618ecdceacc7f2c6ab20eb2484e8dcfa2d470b7/kubernetes/helm/values.override_auctions.makerdao.network.yaml#L13) for example

BUT thats only for the bot, i checked the current helm charts and if the `RPC_URL` secret is also needed for the frontend we have to add the entry to the corresponding `template` and `value-files`.

valiafetisov:
> he corresponding secret path can be found in the individual values_override.xxxx values file like [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/f618ecdceacc7f2c6ab20eb2484e8dcfa2d470b7/kubernetes/helm/values.override_auctions.makerdao.network.yaml#L13) for example

Thanks, I've added this variable to two more paths:
```
secret export --format=dotenv auction-ui/unified-auctions.makerdao.com/bot | grep RPC_URL
secret export --format=dotenv auction-ui/auctions.makerdao.network/bot | grep RPC_URL
```

> we have to add the entry to the corresponding `template` and `value-files`.

Previously we needed `INFURA_PROJECT_ID`, but I can't find this name in any of the template files. As far as I understand, those variables are baked into frontend during build so they only need to be provided via drone and be set within dockerfile. Both are done.

Do I miss anything else or do you think it's in the mergeable state?

deusavalon:
Ah yeah makes sense that why i couldnt see any variables in our cluster because they are being backed into the dockerfile. Alright looks good for me then! :) Sorry for the delay but i had an important call coming up.

PR #332, opened 2022-07-05
Rewrote `notification-service` to use RPC URLs

zoey-kaiser:
Contributes to #330

I saw that in #330 we did not rewrite the `notification-service` to use RPC URLs, so I opened this PR to fix this!

Checklist:

### DIFF
--- a/notification-service/README.md
+++ b/notification-service/README.md
@@ -27,9 +27,8 @@ $ npm run start
 
 Note: env variables are accessible via the `secret` command under `auction-ui/${environment}/notification-service`.
 
-- `INFURA_PROJECT_ID`: (required) [infura](https://infura.io/) project id (can be found in: dashboard -> ethereum ->
+- `WS_URL`: (required) Ethereum Websocket url used for fetching data from the blockchain
 - `ETHERSCAN_API_KEY`: (required) [Etherscan API Key](https://docs.etherscan.io/getting-started/viewing-api-usage-statistics), which is used to automatically download ABIs 
-- `ETHEREUM_NETWORK`: (optional, default `kovan`) – internal network name on which the bot poll for auctions. Available
 - `SMTP_*`: (optional, required for email usage) - SMTP data for the service to send emails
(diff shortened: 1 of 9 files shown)


zoey-kaiser:
From my testing it seems to work well, I only ran into one small issue:

When using the `mainnet` from infura, it returns the network name `homestead`. This obviously messes up some of our internal logic, as we always refer to the network as the `mainnet`. What is the best way to address this issue? If it is only this one network, I could imagine just referring to it as homestead in the `notification-service`, as the service runs separately from our core and other projects.

zoey-kaiser:
> When using the `mainnet` from infura, it returns the network name `homestead`. This obviously messes up some of our internal logic, as we always refer to the network as the `mainnet`. What is the best way to address this issue? If it is only this one network, I could imagine just referring to it as homestead in the `notification-service`, as the service runs separately from our core and other projects.

I renamed `mainnet` to ` homestead` in this project to ensure correct functionality.

notification-service/README.md
```diff
@@ -27,9 +27,8 @@ $ npm run start
 
 Note: env variables are accessible via the `secret` command under `auction-ui/${environment}/notification-service`.
 
-- `INFURA_PROJECT_ID`: (required) [infura](https://infura.io/) project id (can be found in: dashboard -> ethereum ->
+- `WS_URL`: (required) Ethereum Websocket url used for fetching data from the blockchain
```
valiafetisov, line 30: Suggested naming would be `RPC_WEBSOCKET_URL`

notification-service/src/constants/NETWORKS.ts
```diff
@@ -1,6 +1,6 @@
 export const NETWORKS: Record<string, { network: string; etherscanURL: string; etherscanAPI: string }> = {
-    mainnet: {
-        network: 'mainnet',
+    homestead: {
+        network: 'homestead',
```
valiafetisov, line 3: > When using the `mainnet` from infura, it returns the network name `homestead`

What do you mean by `it returns`?

PR #334, opened 2022-07-07
Added `SurplusAuctionTransaction`

### LINKED ISSUE #333: Create `SurplusAuctionTransaction` component and page
### Goal
The user is provided with an overview on different surplus auction transaction related parameters

### Context

This issue covers the creation of the `SurplusAuctionTransaction`. This will be the page where the user is finally able to engage in the surplus auction and place a bid

### Tasks
- [ ] Create the `SurplusAuctionTransaction` component (see mock below) and page (`BidTransaction` component can be used as a reference here)
   - [ ] since the page itself will include all of the different panels (which are not yet merged) we can also go with a plane page in the first place here and just account for the top level information
- [ ] ensure to display all keys with respective values as shown in the mock
- [ ] Implement `The Amount to Bid` as an input field
   - [ ] borrow logic from the existing input field of the `BidTransaction` component
   - [ ] default value should be set to `Lowest next bid` on default
   - [ ] throw an error when the input value is set to a value that is lower than `lowest next bid` 
   - [ ] allow for all values as input that are greater than the `lowest next bid`
- [ ] create stories 

### Assets
`SurplusAuctionTransaction`:

aomafarag:
Closes #333.

Screenshot 2022-07-13 at 14 50 46

Checklist:


frontend/components/common/BaseValueInput.vue
```diff
@@ -38,13 +41,17 @@ export default Vue.extend({
...
         validator: {
-            type: Function as Vue.PropType<function>,
+            type: Function as Vue.PropType<Function>,
             default: () => {},
         },
         disabled: {
             type: Boolean,
             default: false,
         },
+        auctionType: {
```
zoey-kaiser, line 51: Can you please add new stories to reflect the changes and available options for this parameter 🤗. Thank you!

frontend/components/transaction/SurplusAuctionBidTransactionTable.vue
```diff
@@ -0,0 +1,138 @@
...
+    },
+    computed: {
+        isActive(): boolean {
+            return this.auction.state === 'just-started' || this.auction.state === 'have-bids';
+        },
+        isBidAmountNaN(): boolean {
+            return !!this.transactionBidAmount?.isNaN();
+        },
+        lowestNextBid(): BigNumber {
+            return this.auction.bidAmountMKR.multipliedBy(1.1);
```
zoey-kaiser, line 99: Is this the correct function to calculate the lowest next bid? As I was gone for two weeks, I am not up-to-date on this topic 😄

frontend/components/transaction/SurplusAuctionTransaction.vue
```diff
@@ -0,0 +1,82 @@
...
+        <div>HighestBidCheckPanel goes here</div>
+        <div>WithdrawDaiFromVatPanel goes here</div>
+    </div>
+</template>
+
+<script lang="ts">
+import type { SurplusAuction } from 'auctions-core/src/types';
+import Vue from 'vue';
+import { Alert } from 'ant-design-vue';
+import SurplusAuctionBidTransactionTable from './SurplusAuctionBidTransactionTable.vue';
```
zoey-kaiser, line 46: I think we prefer importing from the root of the project.

```suggestion
import SurplusAuctionBidTransactionTable from '~/components/transaction/SurplusAuctionBidTransactionTable.vue';
```

frontend/components/transaction/SurplusAuctionTransaction.vue
```diff
@@ -0,0 +1,82 @@
...
+            default: false,
+        },
+        isExplanationsShown: {
+            type: Boolean,
+            default: true,
+        },
+    },
+    data() {
+        return {
+            isWalletConnected: false,
```
zoey-kaiser, line 77: What is this for?

frontend/components/transaction/SurplusAuctionTransaction.vue
```diff
@@ -0,0 +1,82 @@
...
+        };
+    },
+    computed: {},
```
zoey-kaiser, line 80: Please remove this, if you do not have any computed properties.

frontend/components/utils/BidInput.vue
```diff
@@ -43,6 +44,10 @@ export default Vue.extend({
             type: Boolean,
             default: false,
         },
+        auctionType: {
```
zoey-kaiser, line 47: Please also add new stories to reflect this change. 😄

zoey-kaiser (changes requested):
Please also merge changes from main 😄

frontend/components/transaction/SurplusAuctionBidTransactionTable.vue
aomafarag: Actually, it is not; it is a temporary calculation until the actual function is implemented 😄

zoey-kaiser: In that case, please add a TODO comment, to make sure we catch this before merging 😄 

Example:
```js
// TODO: Implement correct calculation logic for the next possible lowest bid
```

frontend/components/transaction/SurplusAuctionTransaction.vue
aomafarag: For the `WalletConnectionCheckPanel` above

frontend/components/common/BaseValueInput.vue
zoey-kaiser: I did not even see, we don't have a story for this component. Will have to add it during a repo cleanup sometime.

frontend/components/panels/HighestBidCheckPanel.vue
```diff
@@ -78,30 +74,46 @@ export default Vue.extend({
...
+                    return {
+                        name: 'inactive',
+                        title: `You are the highest bidder`,
+                    };
+                }
+                return {
+                    name: 'inactive',
+                    title: `You are not the highest bidder`,
+                };
+            }
```
KirillDogadin-std, line 107: can we shorten this and 

```js
return { 
  name: 'inactive',
  title: this.isUserLatestBidder ? optionA : optionB
}
```
also for some reason the string is in ` instead of '

frontend/components/panels/WalletAuthorizationCheckPanel.vue
```diff
@@ -78,6 +78,12 @@ export default Vue.extend({
...
             if (!this.isWalletAuthorized) {
+                if (this.disabled) {
+                    return {
+                        name: 'inactive',
+                        title: 'The wallet is not yet authorized to execute DAI transactions',
+                    };
+                }
                 return {
                     name: 'incorrect',
                     title: 'The wallet is not yet authorized to execute DAI transactions',
```
KirillDogadin-std, line 89: also this condition here seems to only affect name but title is the same

frontend/components/panels/WalletMKRBalanceCheckPanel.vue
```diff
@@ -87,6 +87,12 @@ export default Vue.extend({
...
             if (this.walletMKR.isLessThan(this.requiredMKR)) {
+                if (this.disabled) {
+                    return {
+                        name: 'inactive',
+                        title: `The amount of MKR is not present in the connected wallet`,
+                    };
+                }
                 return {
                     name: 'incorrect',
                     title: `The amount of MKR is not present in the connected wallet`,
```
KirillDogadin-std, line 98: here too.

KirillDogadin-std (changes requested):

KirillDogadin-std (approved):

LukSteib (changes requested):
- [x] if easily implemented: let's make the lowest next bid clickable and used as input (similar to what we do with the set maximum bid for the dai bidding flow on collateral auctions)
- [ ] currently we disable the allowance panel if there is not enough MKR in the wallet. Can we decouple this? Meaning a user is able to set the allowance even though not enough MKR is in the wallet to actually place the bid.
- [ ] For the Requires Restart case: IMO we wouldn't even show this transaction page for that state but only the surplus auction page. Similar to the behavior on collateral auctions. So consequently we wouldn't need this story, correct?

LukSteib (approved):

frontend/components/common/BaseValueInput.vue
```diff
@@ -38,13 +41,17 @@ export default Vue.extend({
             default: undefined,
         },
         validator: {
-            type: Function as Vue.PropType<function>,
+            type: Function as Vue.PropType<Function>,
```
valiafetisov, line 44: Why this change was needed? `Function` is a js constructor, `function` is a ts type

frontend/components/common/BaseValueInput.vue
```diff
@@ -3,14 +3,17 @@
         <Tooltip :visible="!!errorMessage" placement="topLeft" :title="errorMessage">
             <Input
                 v-model="inputText"
-                class="Input"
-                :class="{ Error: !!errorMessage }"
+                class="Input pr-8"
+                :class="{ Error: !!errorMessage, 'pr-10': auctionType === 'surplus' }"
```
valiafetisov, line 7: Why do you need to add that here?

frontend/components/common/BaseValueInput.vue
```diff
@@ -3,14 +3,17 @@
...
                 :disabled="disabled"
                 @focus="hideMaxValue"
                 @blur="showMaxValueIfEmpty"
             />
             <div class="Overlay right-2" :class="{ 'opacity-50': disabled }">
-                <format-currency v-if="!inputValue && maxValue" :value="maxValue" />&nbsp;DAI
+                <div v-if="auctionType === 'surplus'">
+                    <format-currency v-if="!inputValue && minValue" :value="minValue" /> MKR
+                </div>
+                <div v-else><format-currency v-if="!inputValue && maxValue" :value="maxValue" />&nbsp;DAI</div>
```
valiafetisov, line 16: Can't you just provide `currency` prop instead?

frontend/components/panels/WalletAuthorizationCheckPanel.vue
```diff
@@ -79,7 +80,7 @@ export default Vue.extend({
             }
             if (!this.isWalletAuthorized) {
                 return {
-                    name: 'incorrect',
+                    name: this.disabled ? 'inactive' : 'incorrect',
```
valiafetisov, line 83: Just wondering, how does those changes ended up in the PR for `SurplusAuctionTransaction`?

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
```diff
@@ -0,0 +1,145 @@
...
+    },
+    props: {
+        auction: {
+            type: Object as Vue.PropType<SurplusAuction>,
+            required: true,
+        },
+        lowestNextBid: {
+            type: Object as Vue.PropType<BigNumber>,
+            required: true,
+        },
```
valiafetisov, line 94: This should be part of the auction, see here https://github.com/sidestream-tech/unified-auctions-ui/pull/345

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
```diff
@@ -0,0 +1,145 @@
...
+                return this.lowestNextBid.dividedBy(this.auction.receiveAmountDAI || 1);
+            }
+            return this.transactionBidAmount.dividedBy(this.auction.receiveAmountDAI || 1);
+        },
+    },
+    watch: {
+        transactionBidAmount: {
+            immediate: true,
+            handler(transactionBidAmount) {
+                this.$emit('inputBidAmount', transactionBidAmount);
```
valiafetisov, line 119: Can those two things be called the same way? Or what is the difference between them?

frontend/components/common/BaseValueInput.vue
zoey-kaiser: Good catch! This was left over from before and I did not notice it!

frontend/components/common/BaseValueInput.vue
zoey-kaiser: The problem is, that we have the space between the cursor in the input field and the currency title on the right hardcoded.

Screenshot 2022-07-14 at 09 58 59

This means that if we add a new currency and the symbol is slightly longer or shorter, the cursor spacing will not adjust and look wonky. 

I do think we should look into rewriting or adjusting the component in the future, however I thought as we are a little bit pressed for time, this does not need to be done in this PR.

frontend/components/common/BaseValueInput.vue
zoey-kaiser: See https://github.com/sidestream-tech/unified-auctions-ui/pull/334#discussion_r920869996 for explanation

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: Yes, I am aware of this. @KirillDogadin-std was not sure when he would finish the PR, so we had agreed to initially calculate it here.

But seeing your new comments on the PR, I will change my types accordingly.

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: I am sorry, I don't quit understand the question 😓

frontend/components/common/BaseValueInput.vue
valiafetisov: > the space between the cursor in the input field and the currency title on the right hardcoded.

It's seems to be super easy to instead of adding the hardcoded padding to add the `invisible` div with the currency text. That wouldn't make a long time, but is not "hacky"

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
valiafetisov: What is the difference between `transactionBidAmount` and `inputBidAmount` so they are called differently?

frontend/components/common/BaseValueInput.vue
zoey-kaiser: We position the currency symbol with absolute. As far as I know I cannot insert a element into the AntD Input field, so I am not sure what you mean here. Could you give me an example? 😊

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: I took this naming from the original component [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/transaction/BidTransactionTable.vue#L148). I do not know what the logic behind it was when first written, however I wanted to keep the two somewhat similar.

frontend/components/common/BaseValueInput.vue
valiafetisov: There is even a slot called `suffix`
https://2x.antdv.com/components/input#components-input-demo-presuffix

zoey-kaiser: Screenshot 2022-07-14 at 11 36 57

Sadly, their `suffix` does not seem to apply the correct padding. 

If I switch the version on their website to V1 (the one we are using, you linked to the V2 docs), you can see that even they have the same padding issues in their demo.
Screenshot 2022-07-14 at 11 39 34

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: Renamed the transactionBidAmount to inputBidAmount.

valiafetisov: Please note that this name suppose to be the same as used in all other components that refer to the bidding amount. What is the name for it in the other places? `inputBidAmount` sounds local to this component (since this component is using `input` to get this value)

frontend/components/common/BaseValueInput.vue
valiafetisov: Ok, I give up, that's the issue for another time then

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
valiafetisov: I don't see the change yet!

valiafetisov: I don't see the change yet!

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
```diff
@@ -0,0 +1,152 @@
...
+            return this.transactionBidAmount.dividedBy(this.auction.receiveAmountDAI || 1);
+        },
+    },
+    watch: {
+        inputBidAmount: {
+            immediate: true,
+            handler(inputBidAmount) {
+                this.$emit('inputBidAmount', inputBidAmount);
+            },
+        },
```
valiafetisov, line 123: I still see `inputBidAmount` as well as `transactionBidAmount` throughout this PR. My question remains: are they the same thing? If so, should we rename `inputBidAmount` into `transactionBidAmount`?

frontend/components/surplus/SurplusAuctionTransaction.stories.js
```diff
@@ -0,0 +1,157 @@
...
+                this.isAuthorizing = false;
+            }, 1000);
+        },
+        authorizeMKR() {
+            this.isAuthorizing = true;
+            setTimeout(() => {
+                this.isAuthorizedMKR = true;
+                this.isAuthorizing = false;
+            }, 1000);
+        },
```
valiafetisov: Based on the https://github.com/sidestream-tech/unified-auctions-ui/issues/346#issuecomment-1184252435, should we already remove them here?

zoey-kaiser: I just removed them while you were reviewing 😄

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: Yes, they are. I will rename them all to `inputBidAmount`. I assume I missed a few, but ill make sure. 
I believe `inputBidAmount` is better, as it is more general then `transactionBidAmount`.

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: I had changed it in the parent and provided it as a prop. But that makes no sense, so I just take it directly. My bad!

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: True, then I will rename it just to bidAmount.

frontend/components/surplus/SurplusAuctionBidTransactionTable.vue
zoey-kaiser: I actully have to rollback on this. While replacing everything I noticed, that `transactionBidAmount` is used [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/transaction/BidTransaction.vue#L177) to calculate what the amount is based on the users `inputBidAmount` or the min or max values, if no input is provided. 


Therefore I suggest (and changed) all references that can be changed by the user to `inputBidAmount` to keep the terminology separated meaning:

`inputBidAmount` = the value directly changed by the user
`transactionBidAmount` = the value we calculate dependent on the user input and other values.

Sorry for all the confusion beforehand!

valiafetisov (dismissed):

valiafetisov (approved):

PR #335, opened 2022-07-07
Update chaos-proxy API

valiafetisov:
Since chaoslabs updated their API methods and the host, we had to update our proxy script as well.

Please note that `CHAOSLABS_SIMULATIONS` was also changed. Get new values via:
```
secret export --format=dotenv auction-ui/localhost/chaos-proxy
```


LukSteib (commented):
Able to start simulation. When accessing the Proxy URL, I receive the following response: 
```
{
"jsonrpc": "2.0",
"id": null,
"error": {
"code": -32700,
"message": "Parse error: Unexpected end of JSON input"
}
}
```

Is this expected?

PR #313, opened 2022-06-17
Added HighestBidCheckPanel

benediktkaiser:
Closes #312

Waiting on #309 to be merged first, therefore this PR merges into it.

Screenshot 2022-06-17 at 16 23 56

Checklist:
- [X] issue number linked above after pound (`#`)
    - replace "Closes " with "Contributes to" or other if this PR does not close the issue
- [X] issue checkboxes are all addressed
- [X] manually checked my feature / not applicable
- [X] wrote tests / not applicable
- [X] attached screenshots / not applicable

PR #336, opened 2022-07-11
Added generate FakeSurplusAuction helper function

zoey-kaiser:
@valiafetisov, I noticed we had a bunch of different helper functions to generate the fakeSurplusAuction. I decided to combine them for simplicity. Can you have a look at this helper function to see if it is correct?

### DIFF
--- a/frontend/helpers/generateFakeSurplusAuction.ts
+++ b/frontend/helpers/generateFakeSurplusAuction.ts
@@ -1,0 +1,58 @@
+import { SurplusAuction, SurplusAuctionBase, SurplusAuctionStates } from 'auctions-core/src/types';
+import BigNumber from 'bignumber.js';
+import faker from 'faker';
+import { random } from 'lodash';
+
+const SURPLUS_AUCTION_STATES: SurplusAuctionStates[] = [
+    'just-started',
+    'have-bids',
+    'ready-for-collection',
+    'collected',
+    'requires-restart',
+];
+
+const NETWORKS = ['mainnet', 'kovan', 'goerli'];
+
+const generateFakeSurplusAuctionBase = function (): SurplusAuctionBase {
+    return {
+        id: faker.datatype.number(),
+        network: faker.helpers.randomize(NETWORKS),
+    };
+};
+
+export const generateFakeSurplusAuction = function (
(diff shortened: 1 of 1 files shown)


valiafetisov:
> I noticed

Where have you noticed that? Can you please link it for the context?

> I decided to combine them

I don't see any changes in the other files. Where then the refactoring will be done?

frontend/helpers/generateFakeSurplusAuction.ts
```diff
@@ -0,0 +1,58 @@
...
+    return {
+        ...auctionBaseData,
+        network: 'mainnet',
+        receiveAmountDAI,
+        receiverAddress,
+        auctionEndDate,
+        bidEndDate,
+        earliestEndDate,
+        state: generatedState,
+        bidAmountMKR,
```
zoey-kaiser, line 52: One issue I saw in the types was that `bidAmountMKR` cannot be undefined. I see an issue here, as if the Auction State is `just-started` (auction is active, but has no bids) this would need to be undefined. What do you think?

zoey-kaiser:
> Where have you noticed that? Can you please link it for the context?

Here are some examples:

https://github.com/sidestream-tech/unified-auctions-ui/blob/493631991e7d1e9a9e7772a8fb0392e660340c5d/frontend/helpers/generateFakeSurplusAuction.ts

https://github.com/sidestream-tech/unified-auctions-ui/blob/501e7f4d96aad1c1c297ad9ee752b6c39736b7db/frontend/helpers/generateFakeSurplusAuction.ts

https://github.com/sidestream-tech/unified-auctions-ui/blob/d35a7b4a5001e0fbd8e732126233b99f9b26a6f7/frontend/helpers/generateFakeSurplusAuction.ts

> I don't see any changes in the other files. Where then the refactoring will be done?

There is no merged version of this helper function yet. This resulted in multiple devs writing their own version of it, leading to inconsistent functions in the PRs. In addition to this it made refactoring the helper function more difficult, after the types were changed (see PRs #313 #318)

I think that in the future we should try to merge a helper function, when we also merge the type definitions. This would help us avoid this.

valiafetisov (approved):

Issue #327, opened 2022-06-30
Create `SurplusAuction` component with page


LukSteib:
### Goal
The user is informed about surplus auction via a dedicated SurplusAuciton page
### Context
We will include the next auction type to the unified auctins UI: Surplus auctions. 
You can find an implementation plan for the whole feature here: https://github.com/sidestream-tech/unified-auctions-ui/issues/188#issuecomment-1152161469

This issue covers the creation of the `SurplusAuction` component together with its page. 
It is very similar to the already existing `Auction` component (and page) -> please check this to get a good understanding what needs to be build.

### Tasks
- [x] Create `SurplusAuction` component and page (see Mock below)
- [x] Create a table with new values similar to the already existing table we use for `Auction` component already
- [x] Ensure that the text below the table is based on dynamic input where needed
   - [x] for `AuctionAmount`
   - [x] for `HighestBid` 
   - [x] for `AuctionPrice` 
   - [x] for `MarketDifference`
- [x] Account for a `finished` state of the auction
   - [x] This includes an alert (again similar to `Auction` component) 
   - [x] And changing the text below the table (see assets for the finished text)
   - [x] dark overlay of table
- [x] Include button 
- [x] Create stories (also for the finished state) 

 
### Assets
<img width="600" alt="image" src="https://user-images.githubusercontent.com/36229155/176649536-1d6e1ecd-7b72-4e90-ba21-0a7264be845e.png">

Finished text:
This auction was finished at {auctionEndTime} at a closing auction price of {highestbidAmount} MKR.

zoey-kaiser:
Needed values:

- fetchedAt
- marketUnitPriceToUnitPriceRatio
- marketUnitPrice

zoey-kaiser:
> This auction was finished at {auctionEndTime} at a closing auction price of {highestbidAmount} MKR.

As discussed in the daily today, this text will not be possible when an auction was collected. Therefore I will change the text to:

"This auction was collected"

This is also just an edge case, as we wont fetch any auctions that were collected. Therefore it would only be shown to the user who collected the auction and is on the page.

PR #338, opened 2022-07-11
Initial implementation

aomafarag:
Contributes to #327.

image

Checklist:

**TODO**
- Proper fetching of surplus auctions, market difference and price on Uniswap

### DIFF
--- a/frontend/components/SurplusAuction.vue
+++ b/frontend/components/SurplusAuction.vue
@@ -1,0 +1,247 @@
+<template>
+    <TextBlock :title="`Surplus auction #${auctionId}`">
+        <div class="my-3">
+            <Alert v-if="auctionError && auctionError.showBanner" :message="auctionError.error" type="error" />
+        </div>
+        <div v-if="auction">
+            <div v-if="auction.state === 'requires-restart'">
+                <AuctionRestartPanel
+                    :wallet-address="walletAddress"
+                    :is-explanations-shown="isExplanationsShown"
+                    :is-restarting="auction.state === 'requires-restart'"
+                    @restart="$emit('restart', auctionId)"
+                    @connectWallet="$emit('connect')"
+                    @disconnectWallet="$emit('disconnect')"
+                />
+            </div>
+            <div class="relative mt-2">
+                <table class="w-full table-fixed border-collapse border">
+                    <tbody>
+                        <tr>
+                            <td>Auction State</td>
+                            <td>Ends in <time-till :date="auction.auctionEndDate" /></td>
+                        </tr>
+                        <tr>
+                            <td>Auction Amount</td>
+                            <td>
+                                <format-currency :value="auction.receiveAmountDAI" currency="DAI" />
+                            </td>
+                        </tr>
+                        <tr>
+                            <td>Highest Bid</td>
+                            <td v-if="auction.bidAmountMKR">
+                                <format-currency :value="auction.bidAmountMKR" currency="MKR" />
+                            </td>
+                            <td v-else>
+                                <span class="opacity-50">No bids yet</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <td>Auction Price</td>
+                            <td>
+                                <template v-if="auction.bidAmountMKR">
+                                    <format-currency :value="unitPrice" currency="MKR" />
+                                    per <format-currency currency="DAI" />
+                                    <PriceDropAnimation :auction="auction" />
+                                </template>
+                                <span v-else class="opacity-50">No bids yet</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <!--Awaiting another issue: Fetching price on Uniswap-->
+                            <td>Price On Uniswap</td>
+                            <td>
+                                <template v-if="isActive && auction.marketUnitPrice">
+                                    <format-currency :value="auction.marketUnitPrice" currency="MKR" /> per
+                                    <format-currency currency="DAI" />
+                                </template>
+                                <span v-else class="opacity-50">Unknown</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <!--Awaiting another issue: Calculating market to unit price ratio-->
+                            <td>Market Difference</td>
+                            <td>
+                                <template v-if="isActive && auction.marketUnitPriceToUnitPriceRatio">
+                                    <format-market-value :value="auction.marketUnitPriceToUnitPriceRatio" />
+                                </template>
+                                <span v-else class="opacity-50">Unknown</span>
+                            </td>
+                        </tr>
+                        <tr>
+                            <!--Awaiting another issue: Fetching surplus auctions-->
+                            <td>Last updated</td>
+                            <td>
+                                <div v-if="areAuctionsFetching" class="flex items-center space-x-2">
+                                    <LoadingIcon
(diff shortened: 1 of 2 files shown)


frontend/components/SurplusAuction.vue
```diff
@@ -0,0 +1,247 @@
...
+            type: Object as Vue.PropType<SurplusAuction>,
+            default: null,
+        },
+        auctionId: {
+            type: String,
+            required: true,
+        },
+        error: {
+            type: String,
+            default: '',
```
zoey-kaiser, line 170: Shouldn't the default be null?

frontend/components/SurplusAuction.vue
```diff
@@ -0,0 +1,247 @@
...
+            immediate: true,
+            handler(newIsExplanationsShown) {
+                if (!newIsExplanationsShown) {
+                    this.isTableExpanded = true;
+                }
+            },
+        },
+        areAuctionsFetching(areAuctionsFetching) {
+            if (!areAuctionsFetching && !this.auction && !this.error) {
+                this.$emit('fetchTakeEventsFromAuction', this.auctionId);
```
zoey-kaiser, line 232: Let us remove this line for now. For reference, this line is called when the auction could not be found. In the case of swap auctions, we fetch the take events (all the transactions used to take up the auction), however I don't know what we want to fetch for expired surplus auctions.

zoey-kaiser (changes requested):

core/src/types.ts
```diff
@@ -167,6 +167,10 @@ export declare interface SurplusAuctionActive extends SurplusAuctionBase {
     bidEndDate?: Date;
     earliestEndDate: Date;
     state: 'just-started' | 'have-bids' | 'ready-for-collection' | 'requires-restart';
+    fetchedAt: Date;
+    isRestarting: boolean;
```
zoey-kaiser, line 171: @valiafetisov for the swap auctions we had a system of the following types:

- AuctionInitialInfo
- Auction
- AuctionTransaction

How do we plan to handle this for these new auctions? With this PR we need to store more information we would normally try to exclude from the Initial Auction Data (isRestarting, fetchedAt and MarketPrices).

Do you have any kind of idea to better structure the types?

zoey-kaiser: > * `SurplusAuctionTransaction` would contain most other fields like `marketPrice` or `...TransactionFee`-like fields

What I take from this, to simplify it for @aomafarag (correct me if I am wrong):
- for isRestarting look at the logic from the other PR and copy it over (at least via props or whatever else is needed)
- fetchedAt can be moved to the SurplusAuctionBase
- A new type (extending `SurplusAuctionActive`) called `SurplusAuctionTransaction` can be created, with the new properties needed (for now just `marketUnitPrice` and `marketUnitPriceToUnitPriceRatio` to keep it simple, the rest can be added later on)

zoey-kaiser: @aomafarag, please check the following link for the types that @KirillDogadin-std and I have agreed on using for his projects and some other components I am working on:

https://github.com/sidestream-tech/unified-auctions-ui/issues/341#issuecomment-1181436064

zoey-kaiser: Merged types from main into this PR. Everything should be available now!

KirillDogadin-std (changes requested):

frontend/helpers/generateFakeSurplusAuction.ts
```diff
@@ -68,7 +68,7 @@ export const generateFakeSurplusAuctionTransaction = function (
     const surplusAuction = generateFakeSurplusAuction(state);
 
     // if auction does not have any bids yet or is finished return without market data
-    if (!surplusAuction.bidAmountMKR || !surplusAuction.receiveAmountDAI) {
+    if (!surplusAuction.bidAmountMKR || !surplusAuction.receiveAmountDAI || surplusAuction.state === 'just-started') {
```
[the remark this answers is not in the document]
zoey-kaiser: I would generally do that, however this is more a typescript issue. Typescript does not know that every Auction aside from just-started will have these values and therefore throws errors. 

As this is only a generation file for mock data, I did not put too much effort into making sure that typescript would identify it here in this case. Thats why I only have the values.

KirillDogadin-std (approved):

frontend/helpers/generateFakeSurplusAuction.ts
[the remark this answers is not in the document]
valiafetisov: Very good point, since it can actually be (and I would argue _should be_) expressed via types. You can define separate `SurplusAuctionStarted` type and make `SurplusAuction` to be a union between those. See https://www.typescriptlang.org/docs/handbook/unions-and-intersections.html#discriminating-unions for example.

LukSteib (changes requested):
- [x] let's make the button ("Bid using MKR" or "Collect earnings") primary buttons. Currently they are secondary but there is no other option
- [x] since we denominate `unitPrice` and `marketUnitPrice` in `MKR` per `DAI` we can expect these figures to be quite low. Meaning we need to display more decimal places. Expect values like 0.00033. 
- [x] The text part below the table "This equals MKR per DAI" lacks a variable before MKR
- [ ] We only emit button click actions for the "Has Bids" story. Don't know whether this needs to be added strictly to the other stories as well

LukSteib (approved):

frontend/helpers/generateFakeSurplusAuction.ts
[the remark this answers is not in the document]
zoey-kaiser: I ran into issues using this approach when writing the generation files. I sadly cannot find the commit where this error occurred, but I will outline it here:

types:
```ts
interface BaseAuctionData {
   id: number,
   network: string,
}

interface CollectedAuction extends BaseAuctionData {
   state: 'collected'
}

interface ActiveAuction extends BaseAuctionData {
   state: 'just-started', 'have-bids'
}

interface Auction = ActiveAuction || CollectedAuction

interface AuctionTransaction extends Auction {
   unitPrice: number,
   marketUnitPrice: number,
}
```

I then had one function that generated a `Auction`, randomly determining the state. When I then wanted to generate a AuctionTransaction, I would begin by creating a Auction (with a set state of `have-bids`, so ensure no collected Auction was generated).
Then I would take the data from the generated Auction to compute the values needed for the AuctionTransaction.

```ts
function generateFakeAuction(state?: string): Auction {
   ...generate mock data for auction based on provided state and if none is provided a random state.
}

function generateFakeAuctionTrasnaction(): AuctionTransaction {
   auction = generateFakeAuction('have-bids');

  const unitPrice = auction.totalAmountDAI.dividedBy(auction.bidAmountMKR)
  ...etc.
}
```

I would then get a typescript error thrown at the line:
```ts
const unitPrice = auction.totalAmountDAI.dividedBy(auction.bidAmountMKR)
```

due to `totalAmountDAI` and `bidAmountMKR` not existing on a completed Auction. As far as I could see I had two choices.

- Dont create new types for separate Auction States and have it all be handled via just Auction.
- Create seperate generation functions for the Auctions (such as `generateFakeActiveAuction` and `generateFakeCompletedAuction`)

Ultimately, I decided to use option two, as I also feared, I would run into more similar type based issues on a component or store level (I don't know if this fear was justified, but it did lead me to make my decision). 

I hope this text helped clear up the issue I ran into and my thought process 🤗

PR #339, opened 2022-07-11
Market price mkr

KirillDogadin-std:
Closes # .

Checklist:

### DIFF
--- a/core/src/calleeFunctions/helpers/uniswapV2.ts
+++ b/core/src/calleeFunctions/helpers/uniswapV2.ts
@@ -1,4 +1,4 @@
-/* 
+/*
     Most of the uniswap logic based on the `auction-demo-keeper` source code
     https://github.com/makerdao/auction-demo-keeper/blob/main/src/clipper.js#L134-L160
     More info on how UniSwap works and other available methods can be found here:
@@ -14,6 +14,7 @@ import { getDecimalChainIdByNetworkType } from '../../network';
 import getProvider from '../../provider';
 import { getTokenAddressByNetworkAndSymbol, getTokenDecimalsBySymbol } from '../../tokens';
 import { getCollateralConfigBySymbol } from '../../constants/COLLATERALS';
+import { DAI_NUMBER_OF_DIGITS, MKR_NUMBER_OF_DIGITS } from '../../constants/UNITS';
 
 const EXCHANGE_RATE_CACHE = 20 * 1000;
 
@@ -82,7 +83,7 @@ export const getUniswapPairBySymbols = async function (
 };
 
 export const splitArrayIntoPairs = function (array: string[]): string[][] {
-    /* 
+    /*
         Function that takes an array of strings, e.g.: `[ 'one', 'two', 'three' ]`
         and turns it into array or pairs, e.g.: `[ ['one', 'two'], [ 'two', 'three' ] ]`
     */
@@ -101,6 +102,23 @@ export const getLpTokenTotalSupply = async function (network: string, symbol: st
     const collateral = getCollateralConfigBySymbol(symbol);
     return new BigNumber(totalSupply.toString()).shiftedBy(-collateral.decimals);
 };
+
+export const getExchangeRateDaiMkr = async function (
+    network: string,
+): Promise<BigNumber> {
+    const completeExchangePath = ['DAI', 'MCD_GOV'];
+    const uniswapPairs = await getUniswapPairBySymbols(network, completeExchangePath[0], completeExchangePath[1]);
+    const exchangeToken = await getUniswapTokenBySymbol(network, 'DAI');
+    const uniswapRoute = new Route([uniswapPairs], exchangeToken);
+    const uniswapTrade = new Trade(
+        uniswapRoute,
+        new TokenAmount(exchangeToken, (new BigNumber(1)).shiftedBy(MKR_NUMBER_OF_DIGITS).toFixed(0)),
+        TradeType.EXACT_INPUT
+    );
+    const rate = new BigNumber(uniswapTrade.outputAmount.toSignificant(DAI_NUMBER_OF_DIGITS));
+    return new BigNumber(1).div(rate)
+};
+
 
 export const getRegularTokenExchangeRateBySymbol = async function (
     network: string,
--- a/core/src/calleeFunctions/helpers/uniswapV3.ts
+++ b/core/src/calleeFunctions/helpers/uniswapV3.ts
@@ -3,7 +3,7 @@ import { abi as UNISWAP_V3_QUOTER_ABI } from '@uniswap/v3-periphery/artifacts/co
 import BigNumber from '../../bignumber';
 import getProvider from '../../provider';
 import { getContractAddressByName } from '../../contracts';
-import { DAI_NUMBER_OF_DIGITS } from '../../constants/UNITS';
+import { DAI_NUMBER_OF_DIGITS, MKR_NUMBER_OF_DIGITS } from '../../constants/UNITS';
 import { getCollateralConfigBySymbol } from '../../constants/COLLATERALS';
 
 const UNISWAP_V3_QUOTER_ADDRESS = '0xb27308f9F90D607463bb33eA1BeBb41C27CE5AB6';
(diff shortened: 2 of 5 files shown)


LukSteib (dismissed):
- [ ] I have a feeling that we are currently fetching the wrong `marketUnitPrice`. So what happens in the auction is the following: User bids `MKR` to obtain `DAI`. In order to have a valid comparison we need to fetch the market price to swap `MKR` for `DAI` equivalently. Is this done already? I have a feeling that we are getting price to swap `DAI` for `MKR`.
- [ ] On the similar vein: I think we need to adjust the `unitPrice` calculation and flip it the other way around -> `const unitPrice = auction.bidAmountMKR.div(auction.receiveAmountDAI)`

Would like to put a note here and get @valiafetisov's review and confirmation on that as well

KirillDogadin-std:
adjusted the logic pls see if it is what you meant'

frontend/store/surplus.ts
```diff
@@ -92,6 +98,27 @@ export const mutations = {
     setAuctionState(state: State, { auctionId, value }: { auctionId: number; value: AuctionState }) {
         Vue.set(state.auctionStates, auctionId, value);
     },
+    setIsMarketPriceLoading(state: State, isLoading: boolean) {
```
zoey-kaiser, line 101: I would remove all of this logic from the store. I think we should have the following pattern in the core when retrieving auctions: 

- fetch all auctions
- enrich each auction with transaction fees
- enrich each auction with market price
- return all auctions enriched with every value we need

This is how we do it with the swap auctions. 
https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/auctions.ts#L143

This will make it a lot easier when we want to interact with surplus auctions in other parts of the project (such as the bot), as we would not need to rebuild the logic in there.

core/src/calleeFunctions/helpers/uniswapV3.ts
```diff
@@ -66,3 +66,17 @@ export const convertCollateralToDai = async function (
     const daiAmount = new BigNumber(daiIntegerAmount._hex).shiftedBy(-DAI_NUMBER_OF_DIGITS);
     return daiAmount;
 };
+
+export const convertDaiToMkr = async function (network: string, amountDai: BigNumber): Promise<BigNumber> {
```
valiafetisov, line 70: Sorry, I think I misunderstood your question during one of the standups. Please refactor `convertCollateralToDai` into `convertCollateralToDai`+`convertSymbolToDai`. Where you then can provide `MKR`? otherwise there is too much duplicative code

core/src/tokens.ts
```diff
@@ -19,6 +19,9 @@ export const getTokenDecimalsBySymbol = function (symbol: string): number {
     if (tokenName === 'DAI') {
         return DAI_NUMBER_OF_DIGITS;
     }
+    if (tokenName === 'MCD_GOV') {
```
valiafetisov, line 22: I also suggest to use `MKR` here and everywhere, eg by extending `getTokenAddressByNetworkAndSymbol`

core/src/calleeFunctions/helpers/uniswapV3.ts
```diff
@@ -66,3 +66,17 @@ export const convertCollateralToDai = async function (
...
+    const auctionIntegerAmount = amountDai.shiftedBy(DAI_NUMBER_OF_DIGITS).toFixed(0);
+    const uniswapV3quoterContract = await getUniswapV3quoterContract(network);
+    const daiIntegerAmount = await uniswapV3quoterContract.callStatic.quoteExactInputSingle(
+        await getContractAddressByName(network, 'MCD_DAI'),
```
valiafetisov, line 74: I think `getTokenAddressByNetworkAndSymbol` is the better function for that purpose

frontend/store/surplus.ts
```diff
@@ -103,7 +122,9 @@ export const actions = {
         try {
             commit('setAuctionsFetching', true);
             const auctions = await fetchActiveSurplusAuctions(network);
-            auctions.forEach(auction => commit('addAuctionToStorage', auction));
+            const enrichedAuctionPromises = auctions.map(auc => enrichSurplusAuctionWithMarketPrice(network, auc));
+            const auctionTransactions = await Promise.all(enrichedAuctionPromises);
+            auctionTransactions.forEach(auctionTransaction => commit('addAuctionToStorage', auctionTransaction));
```
valiafetisov, line 127: As suggested by @zoey-kaiser, please enrich surplus auction in one step via a single `enrichSurplusAuction` function that does all kind of enrichments

core/src/calleeFunctions/helpers/uniswapV2.ts
```diff
@@ -102,6 +103,20 @@ export const getLpTokenTotalSupply = async function (network: string, symbol: st
     return new BigNumber(totalSupply.toString()).shiftedBy(-collateral.decimals);
 };
 
+export const getExchangeRateDaiMkr = async function (network: string, amount: BigNumber): Promise<BigNumber> {
```
valiafetisov, line 106: Let's remove it, since we don't use it

valiafetisov (commented):
The main question is: When the user bid MKR on the auction, it effectively exchanges MKR into DAI. Let's say 1 MKR will be exchanged into 1000 DAI if the user bid 1 MKR with exchange rate being 1/1000. Therefore, we need to fetch the same thing: how much MKR can be exchanged for 1000 DAI, right? Or is it not possible/actually the same number as 1000 DAI -> MKR?

frontend/store/surplus.ts
```diff
@@ -112,7 +131,19 @@ export const actions = {
...
+                if (auc.state === "collected") {
+                    collectedAuctions.push(auc)
+                } else {
+                    activeAuctions.push(auc)
+                }
+            })
+            const enrichedAuctionPromises = activeAuctions.map(auc => enrichSurplusAuction(network, auc));
+            const auctionTransactions = await Promise.all(enrichedAuctionPromises);
+            auctionTransactions.forEach(auctionTransaction => commit('addAuctionToStorage', auctionTransaction));
+            collectedAuctions.forEach(auction => commit('addAuctionToStorage', auction));
```
valiafetisov, line 146: Why do you need to work with them separately here? The main rule is: frontend doesn't need to know anything about auctions internal logic, it should just call fetch, enrich, big, etc. So I would rather move this part into enrich function?

KirillDogadin-std:
> The main question is

here i don't see the line you're trying to conduct.

From my perspective: essentially it does not matter what ratio is going to be displayed, because the purpose there is to indicate how profitable the deal is. and however we flip it (display how much __mkr per dai__ or __dai per mkr__). For now i did it the way when you see how many day you get per paid mkr.

valiafetisov:
> For now

Alright, let's roll out with the current rates 🙂

valiafetisov (approved):

core/src/surplus.ts
```diff
@@ -152,28 +153,29 @@ export const collectSurplusAuction = async function (network: string, auctionInd
 export const enrichSurplusAuction = async (
     network: string,
     auction: SurplusAuctionActive
-): Promise<SurplusAuctionWithMinimumBid> => {
+): Promise<SurplusAuctionEnriched> => {
+    const unitPrice = auction.receiveAmountDAI.div(auction.bidAmountMKR);
+    const marketUnitPrice = (await convertMkrToDai(network, auction.bidAmountMKR)).div(auction.bidAmountMKR);
```
valiafetisov, line 158: While `unitPrice` = `receiveAmountDAI` / `bidAmountMKR` – is effectively MKR price
The other number is something else: `convertMkrToDai` returns amount of DAI will be paid for `bidAmountMKR`, so in order to get MKR unit price, you need to divide it by the amount of MKR, not DAI.

But in general, we want the opposite number (MKR/DAI):
```suggestion
    const unitPrice = auction.bidAmountMKR.div(auction.receiveAmountDAI);
    const mkrMarketPrice = (await convertMkrToDai(network, auction.bidAmountMKR)).div(auction.bidAmountMKR);
    const marketUnitPrice = new BigNumber(1).div(mkrMarketPrice);
```

valiafetisov (approved):

PR #345, opened 2022-07-13
Fetch the minimal bid coefficient increase rate

KirillDogadin-std:
Closes # .

Checklist:

### DIFF
--- a/core/test/surplus-test.ts
+++ b/core/test/surplus-test.ts
@@ -5,6 +5,8 @@ import {
     collectSurplusAuction,
     fetchSurplusAuctionByIndex,
     restartSurplusAuction,
+    getSurplusAuctionBidIncreaseCoefficient,
+    getNextMinimumBet,
 } from '../src/surplus';
 import { setAllowanceAmountMKR } from '../src/authorizations';
 import { setupRpcUrlAndGetNetworks } from '../src/rpc';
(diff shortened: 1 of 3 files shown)


valiafetisov:
- [ ] Can we please add this value under `nextMinimumBid` to the `SurplusAuction` or `SurplusAuctionTransaction` or other intermediate type
- [ ] Can we please enrich auction after fetching base data (but please create a separate function for enriching in case we ever need to only fetch auction data without extra enrichment)
- [ ] Can please please always use word `bid` instead of `bet` since they are not the same thing

valiafetisov (approved):

core/test/surplus-test.ts
```diff
@@ -96,4 +111,15 @@ describe('Surplus Auction', () => {
...
     });
+    it('calculates the minimum bid increase', async () => {
+        const auctions = await fetchActiveSurplusAuctions('custom');
+        expect(auctions[0].id).to.equal(2328);
+
+        const bid = await getNextMinimumBid('custom', auctions[0]);
+        if (!bid || !auctions[0].bidAmountMKR) {
+            throw new Error('Expected values were not defined.');
+        }
+        expect(bid.toString()).to.equal('16.94241213279722952');
```
valiafetisov, line 122: There are actually can be a problem that this value is rounded on a wrong way. And in order to properly test it we need to try to bid on the auction with this value (and don't receive an error)

valiafetisov:
> Can please please always use word `bid` instead of `bet` since they are not the same thing

Can we please change it everywhere?

I'm also surprised that I've missed it, but `bidToSurplusAuction` as well as any other function that deal with crypto in our codebase, should always receive numbers as a `BigNumber` with decimals, not as a string. That's a standard we were aiming in our codebase since the beginning with only a few exceptions since conversion `toString` and back is a bad practice (we actually loosing precision here since it's can end up using scientific notations to numbers, in comparison to `toFixed` for example which doesn't). Do you think it's also feasible to change those kind of issues throughout the code in this PR?

Issue #131, opened 2022-03-09
Explain `transactionNetProfit` to user in UI


LukSteib:
**Goal:**
The user understands the calculation behind our indicator for the Transaction Outcome

**Context:**
In PR #95 we introduced changes to the calculation for `transactionProfitMinusFees`. It is `transactionProfit` minus `wallet auth fee` (if not done) minus `collateral auth fee` (if not done) minus `biddingFee`.
Before it was only `transactionProfit` minus `biddingFee`. So we need to take into conisderation now whether the user already did the `wallet auth fee` step and `collateral auth fee` step. 
This needs to be explained to the user properly.

Alternative 1: 
* Add explanation to the text below the transaction table 

Alternative 2:
* List fees as additional rows of the transaction table (take into consideration that these can be 0 if they were executed for the wallet already)
  * added complexity: We can only determine the state of these auth steps once a wallet is connected 

**Tasks:**
- [ ] Decide on the approach to explain the calculation to the user in the UI
- [ ] Implement changes

PR #352, opened 2022-07-14
Explain transaction fees

aomafarag:
Contributes to #131.

This is an initial implementation of a breakdown of the transaction fees in the form of a table that pops up in an `Explain` block. For that, I passed the different fees as props all the way to `TransactionMessage`, and added a new prop, `setWidth`, to `Explain` in order to be able to increase the block's size to accommodate the table and make it readable.

image

Checklist:

TODO:
- [x] We should be able to modify the generation of fake auction transactions, such that we can vary the values of `isWalletAuthed` and `isCollateralAuthed` to account for the different scenarios. Currently, the displayed sum in the newly implemented `TransactionFeesTable` is wrong.
- [x] Generalize `TransactionFeesTable` to other transactions
- [x] Add stories for `TransactionFeesTable`


LukSteib (changes requested):
- [ ] Can you at a story for the case where both wallet and collateral have been authorised and only swapping fees remain
- [ ] what is displayed in the case no wallet is connected? Ie. when we don't know yet the auth status?

frontend/components/transaction/TransactionFeesTable.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+    <div v-if="isWalletConnected" class="flex flex-col space-y-1 text-gray-700 dark:text-gray-100">
+        <div v-if="swapTransactionFeeETH" class="flex justify-between">
+            <div>Swap Transaction Fee</div>
+            <div>
+                <FormatCurrency :value="swapTransactionFeeETH" :decimals="5" :currency="ETH" />
+            </div>
+        </div>
+        <div v-else-if="bidTransactionFeeETH" class="flex justify-between">
```
zoey-kaiser, line 9: You currently account for `bid` and `swap` transactions. Do we also want to add support for `surplus` transactions? You can find the types for these [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/types.ts#L191).

You may need to rename `bidTransactionFee` as it would then overlap. We would also need support for the collection fees (See code to determine total fees [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/surplus.ts#L164))

frontend/components/transaction/TransactionFeesTable.vue
```diff
@@ -0,0 +1,97 @@
+<template>
+    <div v-if="isWalletConnected" class="flex flex-col space-y-1 text-gray-700 dark:text-gray-100">
+        <div v-if="swapTransactionFeeETH" class="flex justify-between">
+            <div>Swap Transaction Fee</div>
+            <div>
+                <FormatCurrency :value="swapTransactionFeeETH" :decimals="5" :currency="ETH" />
```
zoey-kaiser, line 6: For some reason, neither on the screenshots nor in my testing can I see the currency symbol behind the number. Please ensure it says `ETH` behind the numbers so the users know what currency it is in. 

@LukSteib maybe we also want to display the DAI amount for them? As 1 DAI = 1 USD, it may give the user better insight then the ETH amount.

frontend/components/transaction/TransactionFeesTable.vue
```diff
@@ -0,0 +1,97 @@
...
+                <FormatCurrency
+                    v-else-if="combinedBidFeesETH"
+                    :value="combinedBidFeesETH"
+                    :decimals="5"
+                    :currency="ETH"
+                />
+            </div>
+        </div>
+    </div>
+    <div v-else>Connect a wallet for an approximate fee breakdown</div>
```
zoey-kaiser, line 48: I think we can already display some approximate values here. We could break down the different fees, but remove the combined transaction fee at the bottom. 

This way the user can get more insight into how transaction fees look at the current moment.

frontend/components/transaction/TransactionFeesTable.vue
```diff
@@ -0,0 +1,97 @@
...
+import BigNumber from 'bignumber.js';
+import FormatCurrency from '~/components/utils/FormatCurrency.vue';
+
+export default Vue.extend({
+    components: {
+        FormatCurrency,
+        Icon,
+    },
+    props: {
+        swapTransactionFeeETH: {
```
zoey-kaiser, line 63: I have an alternative idea, to make this component easier to extend.

Why don't you pass an object of fees: 

```ts
fees = {
   'Swap Fee': new Bignumber(5),
   ... etc etc
}
``` 

Afterwards you compute the total amount as a computed property. This allows us more flexibility when adding/removing fees. (You create the fee object in the parent component where we receive an auction/surplus auction object)

More insight:
This allows us the remove a lot of these props. The parent component then checks if the wallet is auth for example and only hands down the needed fees.

frontend/components/utils/Explain.vue
```diff
@@ -46,6 +46,10 @@ export default Vue.extend({
             type: String,
             default: 'topRight',
         },
+        setWidth: {
```
zoey-kaiser, line 49: Do we want to call this setWidth? Can we just use the term `width` or is there any kind of overlap?

zoey-kaiser (changes requested):

frontend/components/transaction/TransactionFeesTable.vue
zoey-kaiser: Also see lower comment for implementation suggestion.

frontend/components/transaction/TransactionFeesTable.vue
zoey-kaiser: The auth fees are always the same. The only variable is will they apply. If the user has their wallet authenticated we just don't add them to the list. But if say both of us did not do the authorisation and we use the UI, we will have the same values. 

You can find all Gas Prices on our dashboard here:
https://unified-auctions.makerdao.com/dashboard

So users can use that link to see the prices in advance (this also works without a wallet connected). You can see your component as a "micro" version of the dashboard integrated into the flow.

zoey-kaiser:
NIT: In some instances there is a small space between the explain component and the comma. See picture:

Screenshot 2022-07-18 at 13 30 06

zoey-kaiser:
@LukSteib or @valiafetisov, do we want to integrate the `TransactionMessage` into the Surplus Flow as well? The new Transaction Table was written in a such a way, that it can be expanded for any kind of auction, regardless of the types of Fees. 

If we do want to add it to the SurplusFlow, where do we want to put it?

LukSteib:
image

LukSteib (approved):

valiafetisov (approved):

Issue #329, opened 2022-07-01
keeper: unexpected error: cannot estimate gas; 


dharmendrakariya:
Getting error like below

```
ℹ transaction 2648774c-080a-4cd5-9342-9011f599d6fd: Transaction is preparing to be mined...                                                                                                                19:11:31
ℹ transaction 2648774c-080a-4cd5-9342-9011f599d6fd: Transaction error: "cannot estimate gas; transaction may fail or may require manual gas limit"                                                         19:11:36

[...]

PR #364, opened 2022-07-18
Investigation profit calculation test

### LINKED ISSUE #232: Investigate profit calculation

KirillDogadin-std:
Closes #232

As https://github.com/sidestream-tech/unified-auctions-ui/issues/232#issuecomment-1187019367 requested, adding the test that fails/

Fixes that were added: 

 1. Adjust the logic to already predict the `approximateUnitPrice` based on the next step (price drop) when the current step is close to the end.
 2. Adjust/Reintroduce the `calculateTransactionCollateralOutcome` so that it does not directly compare the values but instead subtracts the values and compares them to close-to-zero value. In other words changed the line that duplicates the logic`(owe > tab)`  from `potentialOutcomeTotalPrice > auction.debtDAI` to `debtDai - PotentialOutcomeTotalPrice < 0.00000000000001`

Checklist:


KirillDogadin-std:
run tests for different configurations

here's the summary

| Block | works with approx | works without approx | works with maximum | :index: |
| :- | :--: | :-: | :-: | :-: |
| 14068414 | n | y | did not check | 7 |
| 14068414 | y | y | did not check | 6 |
| 14068414 | y | n | did not check | 4 |
| 14068414 | n | n | works without | 3 |
| 14068414 | y | y | did not check | 2 |
| 14068414 | y | y | did not check | 1 |
| 14068414 | n | y | works with , without | 0 |
| 14760072 | n | y | with approx, without | 1 |
| 14760072 | y | y | did not check | 2 |
| 14760072 | y | y | did not check | 3 |
| 14760072 | y | y | did not check | 4 |
| 14760072 | y | y | did not check | 9 |
| 14760072 | y | n | did not check | 10 |
| 14760072 | y | n | did not check | 11 |
| 14052140 | y | n | did not check | 0 |
| 14052140 | y | y | did not check | 2 |
| 14052140 | n | n | work without, with | 3 |

KirillDogadin-std:
The observations from https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1190102525 : 

1. Using `calculateAuctionPrice` is justified: it does adjust the price so that the prediction is more accurate at least for some cases which is proven with experiments number 2, 12, 13, 14
2. Sometimes it is harmful that is proven for the case number 0, 6, 7
3. Anomaly: predicting the price with the assumption that the whole `collateralAmount` is going to be purchased seems like valid approach sometimes - according to the records 3, 6, 7, 16

KirillDogadin-std:
for case 0 the following change makes the test pass "with approx"
```diff
export const fetchDateByBlockNumber = async function (network: string, blockNumber: number): Promise<Date> {
    const provider = await getProvider(network);
-    const block = await provider.getBlock(blockNumber);
+    const block = await provider.getBlock(blockNumber + 1);
    return new Date(block.timestamp * 1000);
};

```

KirillDogadin-std:
So apparently there're 2 problems:
 
 1. Precision during comparison.
 2. Speed of execution of the code - if we compute the approximate unit price value during moment of "transition" to the next step (when the price decreases), we end up with an incorrect value.

valiafetisov:
As discussed, please look out for much bigger issue than a few percent difference. In the bug report in #329 relevant line is:
```
keeper: auction "ETH-A:15" clear profit is 1643364586864524 DAI after transaction fees, moving on to the execution 
```
> Note `clear profit` in this context is the net profit

KirillDogadin-std:
Local tests pass, on the drone no :/ 

will address this later, moving on to https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1191231583

KirillDogadin-std:
regarding https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1191499814

this is the case because the hardhat is essentially slow and needs a lot of time. Some tests even time out.
Generally as i understand this, it's due to prediction being made for one moment of time and the execution happening way later in time where the previously made prediction is not relevant anymore.

For that i can't see a good solution that can be implemented from the code side perspective. I would vouch for adjusting the infrastructure here via https://github.com/meltwater/drone-cache . This might even improve the overall company's ci/cd experience.

For fast, cheap and ugly solution we can just cut the tests from the drone for now if we really really have to have this merged.

To reproduce the drone state (where tests fail), one should delete the local hardhat cache.

KirillDogadin-std:
regarding https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1191231583

sadly i just miserably fail to reproduce this via either of the following means:

1. Attach the bot to the goerli _today_ and restart several auctions manually. Even waiting up until almost-deadline state and periodically checking did not result into me seeing the bug. (yes it was done one main and not on this PR's branch)
2. Go to 7151068 block on goerli via hardhat fork, attach bot to the fork. Then i see how the auction that had this bugged value reported actually does not contain a mistake. Moving back one block (7151067) just shows that this auction is missing.

valiafetisov:
> prediction being made for one moment of time and the execution happening way later in time where the previously made prediction is not relevant anymore

Do you actually fetch/enrich prediction right before the execution? Will cleaning all `memoizee` cache help here (not the best strategy ofc, since it may create a situation where cache breaks something)? Will increasing prediction threshold help here?

> To reproduce the drone state (where tests fail), one should delete the local hardhat cache.

Why is that?

> sadly i just miserably fail to reproduce this via either of the following means:

Were you able to reproduce the high number of profit by "failing" fetch function that determines block time?

KirillDogadin-std:
> Do you actually fetch/enrich prediction right before the execution?
```js
                const enrichedAuction = await enrichAuction(NETWORK, auction);
                const expectedProfit = enrichedAuction.transactionGrossProfit as BigNumber;
                await bidWithCallee(NETWORK, auction, wallet);
```

> Will increasing prediction threshold help here?

Yes, setting it to more forgiving value would "fix" it.

> Why is that?

Cache is deleted, hardhat has to actually go and compute the result instead getting it from the fetch. This takes time, hardhat is slow.

> Were you able to reproduce the high number of profit by "failing" fetch function that determines block time?

Could you reiterate about this option? Failing the function implies what action? Intentional faking of returning NaN/null/undefined/0 to try to trigger the behaviour?

> Will cleaning all memoizee cache help here ?

Actually cleaning the cache that is related to fetching date (`fetchLatestBlockDateAndCacheDate`) breaks the test completely.

KirillDogadin-std:
> Were you able to reproduce the high number of profit by "failing" fetch function that determines block time?

changing `fetchDateByBlockNumber` logic so that it returns `null` does not make the prediction fail in the bot on the same block number. Same goes if i throw a fake error in the middle of execution of this funciton so that it does not return.

KirillDogadin-std:
>  Or even simpler, you can mock memoize function to return input function, so there is no cache at all inside tests.

Bare-result-wise. The cache is already being clear according to the suggestion. 

Regarding the other topics: what would be your take on the suggestions [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1192668243)?

core/test/helpers/cache.ts
```diff
@@ -0,0 +1,7 @@
+import { Memoized } from 'memoizee';
+
+export default (cachedFunctions: Memoized<any>[]) => {
```
KirillDogadin-std, line 3: as suggested regarding the clearing cache

valiafetisov:
> Regarding the other topics: what would be your take on the suggestions [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1192668243)?

Long-term, I think we should enable drone cache not only because it will decrease text execution speed, but also because we might reach alchemy quota or will need paid account otherwise. So please create an issue for devops.

For the short-term, I need to understand what exactly is failing: `Some tests even time out` (a) is quite different from `it's due to prediction being made for one moment of time and the execution happening way later` (b). `a` can be fixed via increased provider timeout (you can introduce env/runtime/mocked variable specific for tests). `b` should be fixed by cache invalidation, otherwise needs to be investigated further to come to a definitive description of the problem. For me, current theory doesn't sound convincing: actually, for the execution itself, we [don't use predicted values](https://github.com/sidestream-tech/unified-auctions-ui/blob/7d6c44fed2bb1d370679a5195a10ceeadd0ecad3/core/src/auctions.ts#L243). The predicted `approximateUnitPrice` going down earlier can only trigger failing execution "earlier": while in hardhat there will be no price drop due to no blocks being mined. 

Therefore, I would suggest to refactor this whole logic of forked time based on the actual restrictions we have: eg do not predict `approximateUnitPrice` on a fork and instead return `unitPrice` under that variable.

KirillDogadin-std:
> should be fixed by cache invalidation

so i've went ahead and observed how to invalidate cache. Sadly i did not find a function that just wipes every single cached item in `memozee` package.

alternatively, i could go stub the `memozee` function there but encountered the problem with the fact that one has to invest time - research how to stub the default module export via some library. Sadly, i did not find fast & ez, but instead spotted https://github.com/sinonjs/sinon/issues/1623

so for experimental purposes i just manually wrote the functionality that diables caching for tests, see the diff

https://github.com/sidestream-tech/unified-auctions-ui/pull/364/commits/f7b08cfdd07d17529277e0b9ddf2aa8c0938374a

Assuming that there is not a killer-mistake and the introduced changes actually disable caching successfully, the result is that cache is not the reason for differing results in the runs _with/without_ cache.


so i guess i am going with the following step as [mentioned](https://github.com/sidestream-tech/unified-auctions-ui/pull/364#issuecomment-1193897517):
> needs to be investigated further

___

> eg do not predict approximateUnitPrice on a fork

also tried this one out and this either did not yield any successful change in the results.

valiafetisov:
> research how to stub the default module export via some library

Does this https://stackoverflow.com/a/47058957 not do the trick?

KirillDogadin-std:
During debugging i noticed that the hardhat logs differ depending on whether there's a cache prepared. 

The difference is at least the presense of the error on the chain that originates from the dss-chainlog:

```
dss-chain-log/invalid-key
```

___

@valiafetisov 

Any imminent idea about what it might hint towards? (would save me time researching)

valiafetisov:
`dss-chain-log/invalid-key` error comes from the requesting [contract address](https://github.com/makerdao/dss-chain-log/blob/2eaea03e7ebd8db563eea003272094a5b5bc9ce5/src/ChainLog.sol#L133) from the chainlog when there is no such contract. So it's probably some block number under which some of the collateral types defined in our config doesn't exist yet on the chain (which is pretty normal and expected, that's why it fails silently)

KirillDogadin-std:
Ok. The non-cached hardhat setup now is capable of actually successfully running tests.

The thought process:

1. Notice that when the first test from the profit calculation times out, the logs from it still are output into the console. 
2. Increase the timeout time,
3. have the majority of the tests suddenly pass

Conclusion / hypothesis of what could be the reason:

The blockchain fork actually was super slow, the tests did not wait for its response and failed, the next test was started and the request to reset the chain was sent, thus the test runner and the hardhat fork were not "syncronized" and the fork did not behave as was expected because it was not yet done with the previously sent requests.

___

Additional sidenotes:

- the current main still has some bugs that we tried to fix with #380 
  - this is proven by the tests failing
  - the main problem there:
    - while computing the collateral needed to cover debt, the `if` with comarison was still not done properly. One has to subtract value, then the result has to be rounded `toPrecision` and compared against 0
- to not mess around with the real time that is going, the test forces the each consequent block to have the same timestamp as the previous one. This is ok in my opinion because the goal of the test is to assert the correct prediction for "now", so transactions that give allowances, ... are not the concern, however they do influence the timestamp of blocks and time on for in general.

LukSteib:
Closing this for the moment as we are not actively working on it

PR #365, opened 2022-07-18
Add `WithdrawDAIPanel` component and stories

### LINKED ISSUE #359: Add Dai withdrawal panel to the transaction page
### Goal
A user is able to withdraw DAI from VAT from within the SurplusTransaction page

### Context
After the test made we know now that once the `deal()` method on a collactable auction is called, the obtained DAI is transferred to the user's VAT balance. From there it can be withdrawn. 
Hence we want to include the WithdrawDaiPanel to the TransactionPage

### Tasks

- [ ] Create WithdrawDAIPanel (for the base you might re-use the existing `WithdrawCollateralPanel`) - see Mock below
   - [ ] adjust to DAI
   - [ ] adjust the text
- [ ] Include the existing `WalletAuthorizationCheckPanel` inside the withdrawal panel
- [ ] Create stories for the different possible states
   - [ ] no DAI available in VAT
   - [ ] DAI available but authorization not given
   - [ ] DAI available and auth given 
   - [ ] missing wallet
   - [ ] Loading
   - [ ] expert mode 

### Assets
WithdrawDAIPanel including the `WalletAuthorizationCheckPanel`

aomafarag:
Closes #359.

image

Checklist:

**This is a storybook implementation!**

TODO:
- Emitters (fetching and withdrawing DAI) need to be properly implemented in wallet and store


zoey-kaiser (dismissed):
I looks really good already. Do you want to take it one step further and attempt to integrate it into the SurplusFlow?

Pointers:

- You would need to add the panel in the file `/components/surplus/SurplusAuctionTransactionFlow.vue`
- Any missing props and functions will need to be emited all the way to the `SurplusAuctionsContainer`, where you can connect them to the store. 
- The functions to interact with the VAT are in the [Wallet Store](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/store/wallet.ts#L223)

If you need any help feel free to give me a call and I can give you a few more pointers 😄

LukSteib (changes requested):
- [x] Dai in VAT is not properly fetched (assume this is the same problem as already addressed in #367 point 3)). Meaning that the panel shows that there is no DAI in VAT although the wallet popup clearly indicates that there is. -> No need to address here if addressed via other PR
- [x] Another problem: The withdraw DAI panel only becomes relevant when the auction is collected (ie. finished). However once auction is finished we do not display any panel anymore on the SurplusTransaction page. I see two options
a) Show the WithdrawDAI panel on the finished tx page as the only panel 
b) Extending the message displayed finished alert in the way to make the user aware that DAI might need to be withdrawn from VAT and how to do it in the UI 
c) Keep the as-is status allowing a user to withdraw dai from vat only via wallet popup (without providing further inside) 
@valiafetisov what's your take from a UX perspective here?

valiafetisov:
I would just not hide transaction page in case the auction is finished, as we're already doing for the collateral options. So:
- [ ] Do not hide the whole transaction page in case the auction is finished, but keep it as is
- [ ] Add new panel to the bottom of other panels
- [ ] Make sure to `disable` all other panels in case auction is finished
- [ ] Keep this panel expanded in case there is DAI
- [ ] Unrelated: please add some space between the text and the inner panel (same as below the inner panel)

LukSteib (changes requested):
- [x] Tested the flow but once the auction is collected the finished state is displayed again as before (see screenshot below). So this is not as proposed [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/365#issuecomment-1191658745). Would expect in this case to still display the transaction page with the withdrawDai panel expanded


image

zoey-kaiser:
I implemented the new flow. I had one question about the new Panel. Currently we open the panel whenever there is DAI to extract from the VAT. I believe that during the first steps of the bidding flow, this makes the entire UI a bit more messy and harder to use (see screenshot). 

Do we want to only open the panel when the auction state is also `collected`? This would mean that otherwise the panel still says "There is DAI to collect from VAT", but the panel is just closed. 

Screenshot 2022-07-26 at 10 54 36

LukSteib:
> Do we want to only open the panel when the auction state is also collected? This would mean that otherwise the panel still says "There is DAI to collect from VAT", but the panel is just closed.

Yes, IMO this would be the desired approach. So expand the WithdrawDAIPanel if 
* Auction is collected
* There is DAI in the VAT for the connected wallet

---
On the review side of things: 
- [ ] The DAI amount in VAT seems not to be fetched properly. While testing the panel indicates that there is no DAI in the VAT for the wallet even after refreshing. However our Manage Wallet popup correctly shows that there in fact is DAI in the VAT. Screenshots below show the difference (panel indicating no DAI, Wallet showing effective DAI in VAT) 

image

Screenshot 2022-07-28 at 13 58 49

LukSteib:
>  The DAI amount in VAT seems not to be fetched properly. While testing the panel indicates that there is no DAI in the VAT for the wallet even after refreshing. However our Manage Wallet popup correctly shows that there in fact is DAI in the VAT. Screenshots below show the difference (panel indicating no DAI, Wallet showing effective DAI in VAT)

problem still persists. 
console provides the following error message (don't know whether related or not) 
`Collateral 1849 authorization status error: No contract address found for "MCD_CLIP_1849"`

LukSteib:
To summarise the current state of this PR after the last review based on most recent commits. Still facing the following problems: 
- [ ] `WithdrawDaiPanel` doesn't properly fetch and display the actual amount of DAI that is in VAT (see comment [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/365#issuecomment-1198047451)
- [ ] In the console the following error is thrown constantly: `Collateral 1849 authorization status error: No contract address found for "MCD_CLIP_1849"`

I was not able to test the following desired behaviour: 
> expand the WithdrawDAIPanel if
>> Auction is collected
>> There is DAI in the VAT for the connected wallet

KirillDogadin-std:
https://github.com/sidestream-tech/unified-auctions-ui/blob/12b4914fa2693ea480ddd06a8305e1090a98db9a/frontend/containers/SurplusContainer.vue#L78

this line is the one that is supposed to be the source of the dai balance, but sadly i dont see how "DAI" is the "Collateral", i would expect DAI_VAT value to be contained in the usual `walletBalances` field of the store.

frontend/containers/SurplusContainer.vue
```diff
@@ -78,6 +95,12 @@ export default Vue.extend({
                 }
             },
         },
+        selectedAuction(): SurplusAuction | undefined {
+            return this.auctions.find((auction: SurplusAuction) => auction.id === this.selectedAuctionId);
+        },
```
KirillDogadin-std, line 100: isn't auctions a dict? it's if it's mapped from getters, then i assume it's the same as in store which is `auctionStorage: Record<string, AuctionTransaction>;`

frontend/containers/SurplusContainer.vue
```diff
@@ -102,13 +136,27 @@ export default Vue.extend({
             refreshWallet: 'fetchWalletBalances',
             disconnectWallet: 'disconnect',
         }),
+        authorizeWallet() {
+            this.$store.dispatch('authorizations/authorizeWallet');
+        },
+        withdrawAllDaiFromVat() {
+            this.$store.dispatch('wallet/withdrawAllCollateralFromVat', 'DAI');
+        },
```
KirillDogadin-std, line 144: here you want to withdraw DAI,  which i don't think is a collateral. Please verify this call. I would expect that `withdrawFromVAT` would be called.

KirillDogadin-std (changes requested):

frontend/containers/SurplusContainer.vue
```diff
@@ -75,8 +74,7 @@ export default Vue.extend({
             isWalletAuthorized: 'isWalletAuthorizationDone',
         }),
         daiVatBalance(): BigNumber | undefined {
-            const vatBalances = this.$store.getters['wallet/collateralVatBalanceStore'];
-            return vatBalances.DAI || undefined;
+            return this.walletBalances?.walletVatDAI || undefined;
```
KirillDogadin-std, line 77: i don't see the purpose of the `|| undefined`

```
> a = {'a':  1}
{a: 1}
> a.b?.ekek
undefined
```

KirillDogadin-std (changes requested):

LukSteib (changes requested):
* DAI in VAT is now fetched (after hitting refresh button) ✔️
* Able to withdraw DAI from VAT to wallet via the expandable ✔️ 

The part that is not yet properly done:
- [ ] Only expand WithdrawDAI panel if auction is collected AND DAI is present in VAT

LukSteib (approved):
Approved from my side

frontend/components/auction/surplus/SurplusAuctionBidTransactionTable.vue
```diff
@@ -106,6 +114,9 @@ export default Vue.extend({
             return !!this.inputBidAmount?.isNaN();
         },
         unitPriceAfterBid(): BigNumber | undefined {
+            if (!this.isActive) {
+                return undefined;
```
KirillDogadin-std, line 118: `return` would work the same way

KirillDogadin-std (approved):

PR #369, opened 2022-07-21
Improve site switching

### LINKED ISSUE #357: Discuss switching between auction pages
### Goal
We recently noticed, that there is no dedicated link to go back to either the unified page or the other auction page from either the dashboard, collateral auction or surplus auction page. This discussion hopes to find a solution to this issue.

### Context
I purpose two solutions:

- Add a back button to the top right of each page that lets one navigate back to the unified page, from which the user can select a new option. (See image below)
  - Pro: The user is much more aware of where to click to get back to the unified page
  - Con: It moves the logo a bit to the right, which may interrupt the flow of our page
- Make the logo on the top right navigate the user back to the unified auctions page.
  - Pro: It does not interrupt the flow of our page
  - Con: Some users may not make the connection, that the logo brings them back to the other page

Taking both options into consideration, I definitely prefer option 2.

Screenshot 2022-07-18 at 10 30 04

zoey-kaiser:
Closes #357

Screenshot 2022-07-21 at 09 59 42

This new logic should take network parameters into account. 

Checklist:


frontend/components/layout/HeaderLogo.vue
```diff
@@ -0,0 +1,59 @@
...
+                { label: 'Collateral auctions', value: 'collateral' },
+                { label: 'Surplus auctions', value: 'surplus' },
+            ];
+        },
+    },
+    methods: {
+        switchPage(page: string) {
+            if (this.$router) {
+                const link = generateLink(this.network, page);
+                this.$router.push(link);
```
valiafetisov, line 54: I personally always use control-click and dislike all non-native links on websites. For example, how would the user be able to open another auction type in a separate tab? I know it's much more work, but do you see a way of making it work without reimplementing the whole select? Providing `href` key to the options?

zoey-kaiser: I was debating this as well, as I generally agree (also helps with SEO optimisation). 

Would passing a `href` be enough? I would have gone with using `nuxt-link`. Are there any big differences between the two? I just know that the NextJS Link component does a lot more then just redirecting.

valiafetisov: No, I mean, maybe you extend Select component which is quite complicated component (with dropdown vs modal logic) by adding new `{ label: 'Maker Auction Services', value: '', href: '/' }` so when it's present it turns into `nuxt-link` instead of a button?

zoey-kaiser: Ahhh, that makes more sense now 🤗 

I think it is worth it, Ill get on it.

valiafetisov (commented):
There is also a `modal`, which suppose to contain the same list of links/buttons. As this point I would maybe move some repetitive parts to the subcomponents, otherwise it's easy to miss fixing some parts and not others

zoey-kaiser:
> As this point I would maybe move some repetitive parts to the subcomponents, otherwise it's easy to miss fixing some parts and not others

How would you recommend doing this? I had thought of adding another component that handles the repetition. Can I do this in the same file (like with React) or do I need to create a separate component?

valiafetisov:
Yes, I think you need to create a separate component for it. You can just use the naming scheme to indicate that it's subcomponent, eg `SelectOption` or create a generic `LinkOrButton` component or both 🙂

valiafetisov (approved):

LukSteib (approved):

Issue #367, opened 2022-07-20
Fix known wallet related problems


LukSteib:
### Goal
Improve overall wallet connection experience by fixing known issues

### Context
Recently we've encountered the following three problems that need to be addressed with this issue:
~1) Switching between accounts from within metamask modal is not reflected by the indicated wallet address in the header of our UI~
2) Disconnect connected wallet via button in header doesn't work
3) The auth states for the connected wallet are not properly fetched (While auth to withdraw DAI from VAT via wallet popup was given and correctly indicated when on `/collateral` page, the same step is indicated as requiring authorisation when on `/surplus` page.)

### Tasks
- [x] Fix problem 1)
- [x] Fix problem 2)
- [ ] Fix problem 3)

### Assets:
2)
https://www.loom.com/share/32fa7d678fa246ec9edebeda67dfa399

3)
unauth:
<img width="900" alt="image" src="https://user-images.githubusercontent.com/36229155/179964509-add5eb3d-ae87-4ef7-8825-8542a3ec12f4.png">

auth (same wallet)
<img width="900" alt="image" src="https://user-images.githubusercontent.com/36229155/179964661-847b8ff2-7670-4eb3-83ff-b2812feac58b.png">

LukSteib:
> Switching between accounts from within metamask modal is not reflected by the indicated wallet address in the header of our UI

Not able to reproduce this behaviour. Checked on production and it looks good. I am able to switch between different connect accounts from within metamask and the changes are properly reflected in the UI (e.g. by showing the correct address in header). Did you experience this problem on another environment (@zoey-kaiser , @valiafetisov )

valiafetisov:
> Not able to reproduce this behaviour

Not able to reproduce it either. Also on staging

LukSteib:
> Not able to reproduce it either. Also on staging

Updated issue accordingly

PR #371, opened 2022-07-21
Wallet-related fixes

aomafarag:
Contributes to #367.

Checklist:


LukSteib:
> Could you elaborate the 3rd problem more, @LukSteib?

As indicated in the two screenshots of the issue, we face the following problem:
When opening the wallet modal (ie. hovering over the wallet address in the header and clicking on manage wallet) a user is able to deposit or withdraw DAI from VAT. 
This should be possible no matter which auction type we are referring two. And it should also not be dependent on any of the auction types. 
However I am experiencing the problem that the authorisation to withdraw DAI from VAT is given when opening the modal for one of the auction types (`.../collateral`). But when opening the exact same modal while on the other auction type (`.../surplus`), the authorisation is not indicated as missing. 

Background: As of my understanding what we are authorising at this step is the `MCD_JOIN_DAI` contract

LukSteib (changes requested):
- [ ] The auth state related problem doesn't seem to be fixed. See recording below where for the same wallet the same auth step is one time authorised and the other time un-authorised

https://www.loom.com/share/b7bf97f88a76475b8f93078be8180cde

zoey-kaiser:
Closing this, as I took it over in #379!

PR #372, opened 2022-07-21
Hotfix: Updated `ChangeWalletNetworkModal` to use new network logic

### LINKED ISSUE #370: Fix network switch problem
### Goal
Allow switching network from within the wallet again

### Context
The way to switch network from within the wallet is currently broken. When switching from within metmask to another supported network (e.g. `mainnet`) the popup indicates that `mainnet` is not supported anymore. 
However switching to `mainnet` from within our UI is working fine as expected. Let's fix it and establish the flow that we had before. 

Note: This was probably introduced via the `RPC_URL` support. So let's be careful to not break this one while fixing the issue at hand.

### Tasks
- [x] Fix problem described above (also indicated in screenplay below)

### Assets
recording of problem: 
https://www.loom.com/share/32fa7d678fa246ec9edebeda67dfa399

zoey-kaiser:
Closes #370 

Screenshot 2022-07-21 at 16 28 42

Context:
The force network switch modal was still attempting to import an outdated function from the core folder to determine if the network exists. This made it appear as if even the mainnet was not supported by the UI. How did our tests not find this mistake @valiafetisov?

Checklist:

### DIFF
--- a/frontend/components/modals/ChangeWalletNetworkModal.vue
+++ b/frontend/components/modals/ChangeWalletNetworkModal.vue
@@ -33,7 +33,6 @@
 <script lang="ts">
 import { Modal, Alert } from 'ant-design-vue';
 import Vue, { PropType } from 'vue';
-import { getNetworks } from 'auctions-core/src/network';
 import BaseButton from '~/components/common/inputs/BaseButton';
 
 export default Vue.extend({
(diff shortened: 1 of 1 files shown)


valiafetisov (approved):

valiafetisov:
> How did our tests not find this mistake?

Good question. Is there a story for this component?

zoey-kaiser:
> Good question. Is there a story for this component?

Yes there is, which made it even more confusing for me.

LukSteib (approved):

PR #379, opened 2022-07-26
Wallet related fixes

### LINKED ISSUE #367: Fix known wallet related problems

zoey-kaiser:
Closes #367

Checklist:

### DIFF
--- a/frontend/store/wallet.ts
+++ b/frontend/store/wallet.ts
@@ -14,8 +14,6 @@ import getWallet, { WALLETS } from '~/lib/wallet';
 import notifier from '~/lib/notifier';
 import { getContractAddressByName } from '~/../core/src/contracts';
 
-let WAS_WALLET_DISCONNECTED = false;
-
 interface State {
     walletType?: string;
     address?: string;
@@ -171,7 +169,6 @@ export const actions = {
         }
     },
     async disconnect({ getters, commit, dispatch }: ActionContext<State, State>): Promise<void> {
-        WAS_WALLET_DISCONNECTED = true;
         commit('setIsConnecting', true);
         try {
             const wallet = getWallet(getters.getWallet);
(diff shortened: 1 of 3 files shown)


LukSteib (approved):

frontend/store/wallet.ts
```diff
@@ -295,9 +298,12 @@ export const actions = {
...
         commit('reset');
         if (!getters.isConnected) {
-            await dispatch('autoConnect');
+            if (!WAS_WALLET_DISCONNECTED) {
+                await dispatch('autoConnect');
+            }
         }
         dispatch('refetch');
         dispatch('surplus/setup', undefined, { root: true });
+        WAS_WALLET_DISCONNECTED = false;
```
valiafetisov, line 307: Using a local variable is a very hacky approach, since calling `setup` twice will no longer work and will autoconnect the wallet again. It just doesn't fix the problem, but introduces more side effects.

Instead, I would remove `await dispatch('autoConnect')` from the setup action (since it's only need to be called on the page load and not with every `setup`) and move it to the place where we call `setup` at the page load. I guess right before this line: https://github.com/sidestream-tech/unified-auctions-ui/blob/19a386750245af34e194f8717cc0d9495aa9db8d/frontend/plugins/vuex-persist.client.js#L17

LukSteib (approved):

valiafetisov (approved):

PR #380, opened 2022-07-28
Minor price calculation adjustment

KirillDogadin-std:
contributes #232 

separates the fixes from the tests so that the problems that arise from harhat chache could be addressed later on.

Fixes that were added: 

 1. Adjust the logic to already predict the `approximateUnitPrice` based on the next step (price drop) when the current step is close to the end.
 2. Adjust/Reintroduce the `calculateTransactionCollateralOutcome` so that it does not directly compare the values but instead subtracts the values and compares them to close-to-zero value. In other words changed the line that duplicates the logic`(owe > tab)`  from `potentialOutcomeTotalPrice > auction.debtDAI` to `debtDai - PotentialOutcomeTotalPrice < 0.00000000000001`

It is possible to manually test it via the ui in the frontend on the blocks:

14052140
14068414
14760072


LukSteib (dismissed):
Tested 4 auctions on block 14052147 using hardhat. 
Result: Comparison between indicated net profit and effective DAI in the wallet after swap was for all 4 auctions in the range of +-2 DAI precision. Since this most probably can be traced back to a slight difference in tx fees calculated vs. paid  -> Approved from my side

core/src/price.ts
```diff
@@ -19,7 +19,11 @@ const calculateElapsedSteps = function (auction: Auction, currentDate: Date): nu
     checkAuctionStartDate(auction.startDate, currentDate);
     const auctionStartTimestamp = auction.startDate.getTime();
     const elapsedTime = (currentDate.getTime() - auctionStartTimestamp) / 1000;
-    return Math.floor(elapsedTime / auction.secondsBetweenPriceDrops);
+    const distance = elapsedTime / auction.secondsBetweenPriceDrops;
+    const steps = Math.floor(distance);
+    // already predict for the next step when the next step is about to reached
+    const compensation = distance - steps > 0.99 ? 1 : 0;
+    return steps + compensation;
```
valiafetisov, line 26: Am I reading this correctly: this change drops the price of the auction 1% of the `now - auctionStart` time? So the longer the auction lives the earlier one step compensation will apply? Technically, in the end, it can be bigger than the step itself, eg 2 hours after the start the step will be applied 72 seconds earlier? I don't think it's correct, sounds super random adjustment

Moreover, I think that dropping price earlier shouldn't be correct. If this fix only helps reduce the margin, then we should not apply it at all. If this fix helps making more auctions in the "executable state" on our side, you can instead warp time before execution to make sure that all auctions are profitable. Better be more careful (later apply cut) than earlier

core/src/price.ts
```diff
@@ -66,7 +69,8 @@ export const calculateTransactionCollateralOutcome = function (
     const potentialOutcomeTotalPrice = potentialOutcomeCollateralAmount.multipliedBy(unitPrice); // owe
     if (
         // if owe > tab
-        potentialOutcomeTotalPrice.isGreaterThan(auction.debtDAI)
+        // soft compensation because of precision problems.
+        auction.debtDAI.minus(potentialOutcomeTotalPrice).isLessThan(new BigNumber('0.0000000000000001'))
```
valiafetisov, line 73: Why is this needed? If it is indeed needed, please use `toPrecision` instead https://mikemcl.github.io/bignumber.js/#toP

core/src/price.ts
```diff
@@ -41,23 +45,22 @@ export const calculateAuctionDropTime = function (auction: Auction, currentDate:
     return auction.secondsBetweenPriceDrops - (elapsedTime % auction.secondsBetweenPriceDrops);
 };
 
-export const calculateTransactionGrossProfit = function (auction: Auction): BigNumber {
+export const calculateTransactionGrossProfit = function (
+    auction: Auction,
+    collateralToCoverDebt: BigNumber
```
valiafetisov, line 50: Isn't `collateralToCoverDebt` already part of the auction?

valiafetisov:
Also, in the other PR we discussed:
- [ ] removing unused `time.ts` file
- [ ] improving `date.ts` `fetchDateByBlockNumber` to
  - [ ] try / catch `await provider.getBlock(blockNumber)`
  - [ ] check for `block.timestamp` being 0 or undefined
  - [ ] I guess `fetchDateByBlockNumber` then should return something like `undefined`? And this will later needs to be checked and for example don't calculate approximate price at all?

core/src/price.ts
valiafetisov: I think you forgot to use second parameter to the `toPrecision` which can be `ROUND_DOWN` or `ROUND_UP`. That was the suggestion in the first place. Then you might not need to convert both numbers (otherwise the code you've did doesn't guarantee that they would be rounded to the same number and `isLessThan` wouldn't work)

core/src/date.ts
```diff
@@ -4,22 +4,37 @@ import { getNetworkConfigByType } from './network';
...
 };
 
 const _fetchLatestBlockDateAndCacheDate = async function (
     network: string
-): Promise<{ blockDate: Date; cacheDate: Date }> {
+): Promise<{ blockDate: Date; cacheDate: Date } | undefined> {
     const blockDate = await fetchLatestBlockDate(network);
+    if (!blockDate) {
+        return undefined;
+    }
```
valiafetisov, line 37: If you return undefined here if will be cached for `CURRENT_BLOCK_DATE_CACHE_EXPIRY_MS` if instead you will throw, it will not get cached and will be recalculated during the next call. So you better throw directly at `fetchDateByBlockNumber` than caching it and propagating up just to throw up there

KirillDogadin-std: this would than mean that we dont trycatch at this function though. So we only throw on 0 value, correct?

see the adjustment.

valiafetisov: You can try/catch and then throw inside catch more elaborate message

valiafetisov: But it's also fine I guess

core/src/date.ts
```diff
@@ -7,6 +7,9 @@ const CURRENT_BLOCK_DATE_CACHE_EXPIRY_MS = 60 * 1000;
 export const fetchDateByBlockNumber = async function (network: string, blockNumber: number): Promise<Date> {
     const provider = await getProvider(network);
     const block = await provider.getBlock(blockNumber);
+    if (!block?.timestamp) {
+        throw new Error('Received incorrect timestamp value');
+    }
```
valiafetisov, line 12: Last question: what happens one the frontend if this function throws an error? Where is it catched or does it crash the whole frontend? 🙂

KirillDogadin-std: - in the auction store the call that fetches all aucitons (and has this in the call tree) is in try-except block.
- in the surplus store the call that fetches all surplus auctions also is in trycatch block
- no other calls that impact frontend were found by me.

LukSteib (dismissed):

core/src/date.ts
LukSteib: Given the previous answer it seems like we are handling the error properly. 
@valiafetisov if fine for you as well I would proceed and merge this one.

core/src/price.ts
KirillDogadin-std: Sorry, i don't understand what precisely you mean here with:
  - not needing to convert both numbers and
  - what's the reason not to do it

core/src/price.ts
```diff
@@ -45,28 +46,27 @@ export const calculateTransactionGrossProfit = function (auction: Auction): BigN
...
-    auction: AuctionTransaction
+    auction: Auction
 ): BigNumber {
     // Based on the clipper contract logic
     // https://github.com/makerdao/dss/blob/60690042965500992490f695cf259256cc94c140/src/clip.sol#L357-L380
     const collateralToBuyForTheBid = bidAmountDai.dividedBy(unitPrice);
     const potentialOutcomeCollateralAmount = BigNumber.minimum(collateralToBuyForTheBid, auction.collateralAmount); // slice
     const potentialOutcomeTotalPrice = potentialOutcomeCollateralAmount.multipliedBy(unitPrice); // owe
+    const approximateDebt = new BigNumber(auction.debtDAI.toPrecision(16, 0));
+    const approximatePotentialOutcomeTotalPrice = new BigNumber(potentialOutcomeTotalPrice.toPrecision(16, 0));
```
valiafetisov, line 65: Please next time use `ROUND_DOWN` variable provided by the bignumber instead of the cryptic number ([which is not round down by the way](https://user-images.githubusercontent.com/3393626/181745611-ed378ec9-3fda-41d2-943b-136a08a37a26.png))

```suggestion
    const approximateDebt = new BigNumber(auction.debtDAI.toPrecision(16, BigNumber.ROUND_DOWN));
    const approximatePotentialOutcomeTotalPrice = new BigNumber(potentialOutcomeTotalPrice.toPrecision(16, BigNumber.ROUND_DOWN));
```

valiafetisov (approved):

LukSteib (approved):

PR #386, opened 2022-08-01
Support dynamic decimals in tweets

### LINKED ISSUE #355: Update Twitter message to show more decimals
### Goal
I was using the Twitter bot, when I noticed that some auctions don't display a value, if the collateral amount is under 0.00 (i1).

Do we want to copy the logic from out [`AnimatedNumber`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/utils/AnimatedNumber.vue#L44) component and add it to the [Twitter bot](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/bot/src/notify.ts#L4)? 

### Assets

[i1]
image

### Tasks

- [x] Discuss how we want to display values
- [x] Implement the solution

aomafarag:
Closes #355.

Checklist:

### DIFF
--- a/bot/src/notify.ts
+++ b/bot/src/notify.ts
@@ -1,12 +1,16 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
-import { formatToAutomaticDecimalPoints } from 'auctions-core/src/helpers/formatToAutomaticDecimalPoints';
+import {
+    isValueSmallButNotZero,
+    formatToAutomaticDecimalPoints,
+} from 'auctions-core/src/helpers/formatToAutomaticDecimalPoints';
 import { sendNotification } from './twitter';
 
 const generateNotificationText = function (auction: AuctionInitialInfo): string {
     const url = `${process.env.FRONTEND_ORIGIN}/collateral/?network=${auction.network}&auction=${auction.id}`;
-    return `Collateral auction with ${formatToAutomaticDecimalPoints(auction.collateralAmount)} ${
-        auction.collateralSymbol
-    } just started. Follow the link to participate: ${url}`;
+    const formattedValue = formatToAutomaticDecimalPoints(auction.collateralAmount);
+    const displayValue = isValueSmallButNotZero(auction.collateralAmount) ? `under ${formattedValue}` : formattedValue;
+
+    return `Collateral auction with ${displayValue} ${auction.collateralSymbol} just started. Follow the link to participate: ${url}`;
 };
 
 const notify = async function (auction: AuctionInitialInfo) {
--- a/bot/src/notify.ts
+++ b/bot/src/notify.ts
@@ -5,7 +5,7 @@ import { sendNotification } from './twitter';
 const DECIMAL_PLACES_DEFAULT = 2;
 const DECIMAL_PLACES_MAX = 5;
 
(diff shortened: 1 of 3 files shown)


bot/src/notify.ts
```diff
@@ -1,9 +1,69 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
+import BigNumber from 'auctions-core/node_modules/bignumber.js/bignumber';
 import { sendNotification } from './twitter';
 
+const DECIMAL_PLACES_DEFAULT = 2;
+const DECIMAL_PLACES_MAX = 5;
+
+function isValidNumber(value: BigNumber): boolean {
```
valiafetisov, line 8: Since you (hopefully) copied this from the frontend code: Can it be shared between the two? For example it can be a `formatToAutomaticDecimalPoints` function that lives inside `core/src/helpers` and is used both here and in the frontend component?

valiafetisov:
> I still don't have Goerli up and running 😅 ; so, I couldn't manually check whether this is working properly.

1. What is the problem?
2. If taken the suggestion from above, using a frontend story would be a better way to test it

valiafetisov:
> as it allows for decimal points exceeding the default maximum

This is desired functionality. Why do you think it should be removed?

valiafetisov:
Well, this doesn't answer my question. Why do you think the functionality of the existing component should be downgraded?  Due to its insignificance? Isn't it additional work to make it downgraded in the first place? 

As I can't see a good reason for doing this, I suggest that we keep it as it was: providing the possibility to overwrite number of decimal places from being automatic to being whatever value is specified.

valiafetisov (dismissed):

zoey-kaiser:
> It is only used to format the currency in surplus auctions and their bid transaction tables to a precision of 6 decimal places, e.g. [format-currency](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/auction/surplus/SurplusAuction.vue#L59); an insignificant increase over the default max of 5 decimal places.

Just a quick explanation for why we use 6 decimal places here: 

@LukSteib and I were looking into the normal prices inside SurplusAuctions and concluded that 5 decimal places might not always be enough. Therefore we settled on adding an additional decimal place as a compromise between displaying the needed information and ensuring clean UI design (there are also extra stories for the surplus auctions that show the usage of these decimal places).

LukSteib (approved):
Checked the stories `AnimatedNumber` and `FormatCurrency`  -> looking good and hence approved from my side

frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -1,17 +1,21 @@
...
 import Vue from 'vue';
 import AnimatedNumber from 'animated-number-vue';
 import BigNumber from 'bignumber.js';
+import {
+    isValidNumber,
+    isValueSmallButNotZero,
+    formatToAutomaticDecimalPoints,
+} from 'auctions-core/src/helpers/formatToAutomaticDecimalPoints';
 
 const DECIMAL_PLACES_DEFAULT = 2;
```
zoey-kaiser, line 18: This can be removed right?

zoey-kaiser (changes requested):

valiafetisov (approved):

Issue #350, opened 2022-07-14
Add txFee calculation


LukSteib:
### Goal
Ensure that we properly calculate tx fees in order to display it in the UI

### Context
Given the "last mile" issue for the surplus auction feature one missing part is proper determination of txFees

Refernce to the following types:
```
export declare interface SurplusTransactionFees {
    bidTransactionFee?: BigNumber;
    authTransactionFeeDAI?: BigNumber;
    authTransactionFeeMKR?: BigNumber;
    restartTransactionFee?: BigNumber;
    collectTransactionFee?: BigNumber;
}
```

Given recent discussion on required auth steps it is enough to focus on the following for now: 
* bidTransactionFee
* restartTransactionFee
* collectTransactionFee

We might think about adding the allowanceTransactionFee

### Tasks
- [x] Extend store with needed fee calculations

### Assets
Determined the following gaslimits via metamask UI
* bidTransactionFee = 85181
* allowanceTransactionFee = 48373
* collectTransactionFee = 94114
* _was not able to determine the restartTransactionFee_

PR #387, opened 2022-08-01
Introduce the debt auctions functionality

### LINKED ISSUE #388: Introduce core functionality for debt auctions
### Goal

Have core functions that enable interactions with debt (flop) auctions.

### Context

todo

KirillDogadin-std:
Closes #388 

Checklist:

### Testing the functionality via vuex store 

1. Reset the accont in the metamask to avoid nonce error
2. Adjust `baseFeePerGas` in the `hardhat/hardhat.config.js` so that metamask gas estimation (that ususally is incorrect) does not block you from swapping to eth to mkr later

```
        hardhat: {
            ...
            forking: {
               ...
            },
            /// THIS LINE IS IMPORTANT
            initialBaseFeePerGas: 1,
        },
```

4. swap TEST-ETH to WETH via metamask
5. swap WETH to DAI via metamask
6. Interact with auctions
```sh
# Get the auctions that are active
> await this.$nuxt.$store.dispatch('debt/fetchDebtAuctions')
> this.$nuxt.$store.getters['surplus/auctionStorage']
# Take part in the auctions via  making a bet:
# - give allowance
# - authorize auctions
# - ensure that you're authorized
# - place a bet
> await this.$nuxt.$store.dispatch('authorizations/authorizeWallet')
> await this.$nuxt.$store.dispatch('authorizations/setAllowanceAmount')
> await this.$nuxt.$store.dispatch('wallet/depositToVAT', '10000')

> await this.$nuxt.$store.dispatch('debt/bidToDebtAuction', {auctionIndex: <NUMBER>, lot: <STRING>})
> this.$nuxt.$store.getters['debt/auctionStorage'][<NUMBER>].receiveAmountMKR.toString()
```
7. Testing the collection of the auction is possible on block `14078339`
```
> await this.$nuxt.$store.dispatch('debt/collectAuction', <NUMBER>)
```


KirillDogadin-std:
Tested the present surplus logic manually since it was altered / refactored: 

 - can see auctions 
 - can collect auctions 
 - can bid on the auctions
 - can provide necessary authorizations  & allowances.

core/src/compensationAuction.ts
```diff
@@ -0,0 +1,185 @@
...
+    network: string
+): Promise<CompensationAuctionTransactionFees> {
+    const gasPrice = await getGasPriceForUI(network);
+    const exchangeRate = await getMarketPrice(network, 'ETH');
+
+    const restartTransactionFeeEth = gasPrice.multipliedBy(80563);
+    const allowanceTransactionFeeEth = gasPrice.multipliedBy(48373);
+    const bidTransactionFeeEth = gasPrice.multipliedBy(85181);
+    const collectTransactionFeeEth = gasPrice.multipliedBy(94114);
+    const authTransactionFeeEth = gasPrice.multipliedBy(48356);
```
valiafetisov, line 168: Since those are two different contracts that does different things underneath, I wouldn't expect that restarting, bidding and collecting will cost the same amount of gas

valiafetisov (commented):
I actually don't see the value in mixing logic of two auction types which have two quite different contracts (that can also be changed independently) and only coincidentally share some similarities. From my perspective it would've been better to copy existing parts of the surplus logic instead of introducing more complexity by intertwining those two and even introducing new type and terminology of `compensation`, while still passing low-level details like `contractName`. But I guess at this point we're too far into this change

KirillDogadin-std:
I split the modules, the module of `compensation` is gone.

core/src/debt.ts
```diff
@@ -0,0 +1,98 @@
...
+    restartCompensationAuction,
+} from './compensationAuction';
+
+const CONTRACT = 'MCD_FLOP';
+export const getNextMaximumMkrReceived = async (
+    network: string,
+    debtAuction: DebtAuctionActive
+): Promise<BigNumber> => {
+    const increaseCoefficient = await getCompensationAuctionBidIncreaseCoefficient(network, CONTRACT);
+    return debtAuction.receiveAmountMKR.dividedBy(increaseCoefficient);
```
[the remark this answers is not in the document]
KirillDogadin-std: Someone of us is confusing `assert` and `require`. Are you sure it is me?

frontend/store/debt.ts
```diff
@@ -0,0 +1,196 @@
...
+            Vue.set(state.auctionStorage, auction.id, auction);
+            auctionIdsToBeModified = auctionIdsToBeModified.filter(id => id !== auction.id);
+        }
+        // set others as collected
+        for (const collectedAuctionId of auctionIdsToBeModified) {
+            Vue.set(state.auctionStorage, collectedAuctionId, {
+                id: collectedAuctionId,
+                state: 'collected',
+                fetchedAt: new Date(),
+            } as DebtAuctionCollected);
```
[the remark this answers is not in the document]
KirillDogadin-std: maybe was relevant before . Now with types done in different PR and decoupling of functionality became irrelevant

core/src/compensationAuction.ts
KirillDogadin-std: I beleive i might be lacking info on how to extract the correct info then.

Two factors:

  - if the debt auction never happened - can we still estimate the fees?
  - i'm not familiar with the flow on how to extract them (never done before) and in https://github.com/sidestream-tech/unified-auctions-ui/issues/350 the fees are already listed.

valiafetisov: It's simple, you need to try executing this transaction on the frontend, and then check it's `gas limit` or calling https://docs.ethers.io/v5/api/contract/contract/#contract-estimateGas with the same parameters. You're right, since there were never auctions of this type estimation will only be possible within a simulation/test

core/src/debt.ts
```diff
@@ -0,0 +1,239 @@
...
+        return 'have-bids';
+    }
+    if (isBidExpired) {
+        return 'requires-restart';
+    }
+    return 'just-started';
+};
+
+const _getContractAuctionDuration = async (network: string, contractName: 'MCD_FLAP' | 'MCD_FLOP') => {
+    const contract = await getContract(network, contractName);
```
valiafetisov, line 143: Do you want to clean up unrelated logic and remove contract name parameter in all those functions?
```suggestion
    const contract = await getContract(network, CONTRACT);
```

core/src/debt.ts
```diff
@@ -0,0 +1,239 @@
...
+import { convertMkrToDai } from './calleeFunctions/helpers/uniswapV3';
+import { getGasPriceForUI } from './gas';
+import { getMarketPrice } from './calleeFunctions';
+
+const CONTRACT = 'MCD_FLOP';
+export const getNextMaximumMkrReceived = async (
+    network: string,
+    debtAuction: DebtAuctionActive
+): Promise<BigNumber> => {
+    const increaseCoefficient = await getDebtAuctionBidIncreaseCoefficient(network, CONTRACT);
```
valiafetisov, line 27:
```suggestion
    const decreaseCoefficient = await getDebtAuctionBidDecreaseCoefficient(network, CONTRACT);
```

valiafetisov (commented):
Can you also provide console commands for testing (we can do that as soon as we have a simulation)?

core/src/compensationAuction.ts
KirillDogadin-std: So i assume we're blocked here until the simulation is solved.

KirillDogadin-std:
Will provide the commands in the header comment to test

core/src/debt.ts
```diff
@@ -0,0 +1,236 @@
...
+    return await Promise.all(auctionsWithNextMinimumBidsPromises as Promise<DebtAuctionTransaction>[]);
+};
+
+export const getDebtAuctionLastIndex = async (contract: Contract): Promise<number> => {
+    const auctionsQuantityBinary = await contract.kicks();
+    return new BigNumber(auctionsQuantityBinary._hex).toNumber();
+};
+
+const _getDebtAuctionBidIncreaseCoefficient = async (network: string): Promise<BigNumber> => {
+    const contract = await getContract(network, 'MCD_FLOP');
```
valiafetisov, line 115: Do you want to replace all `MCD_FLAP` with the CONTRACT defined above for consistency? 💅

core/src/debt.ts
```diff
@@ -0,0 +1,237 @@
...
+        auctionStartDate,
+    };
+};
+
+export const getActiveDebtAuctionOrUndefined = async (network: string, auctionIndex: number) => {
+    const auction = await fetchDebtAuctionByIndex(network, auctionIndex);
+    if (auction.state === 'collected') {
+        return;
+    }
+    return auction;
```
valiafetisov, line 198: Do you want to directly typecast the output to be to be active auction, instead of typecasting every time you're using this function?

valiafetisov (approved):
Code approved with minor comments

@LukSteib do you want to wait for the simulation or pursue our own tests in order to merged this? Since currently it can't be test functionally and I can't fully validate the logic behind

LukSteib:
> @LukSteib do you want to wait for the simulation or pursue our own tests in order to merged this? Since currently it can't be test functionally and I can't fully validate the logic behind

How involved would the testing on our side be? Basically simulation the whole flow of forking, increasing the protocol debt, kicking a debt auction?

valiafetisov (approved):

core/src/debt.ts
```diff
@@ -0,0 +1,231 @@
...
+import getContract from './contracts';
+import { RAD_NUMBER_OF_DIGITS, WAD_NUMBER_OF_DIGITS } from './constants/UNITS';
+import { Contract } from 'ethers';
+import { MKR_NUMBER_OF_DIGITS, RAD, WAD } from './constants/UNITS';
+import memoizee from 'memoizee';
+import getNetworkDate from './date';
+import { CompensationAuctionBase, DebtAuction, Notifier, CompensationAuctionTransactionFees } from './types';
+import { getEarliestDate } from './helpers/getEarliestDate';
+import executeTransaction from './execute';
+import { convertMkrToDai } from './calleeFunctions/helpers/uniswapV3';
```
valiafetisov, line 12: Why was this function used instead of the newly created `convertDaiToMkr`?

KirillDogadin-std: That's the incautious use of autocomplete from my side. Essentially needs to be fixed. Should i open a quickfix PR?

PR #389, opened 2022-08-02
Extend twitter bot to tweet about surplus auctions

### LINKED ISSUE #385: Extend twitter bot to tweet about surplus auctions
### Goal
The twitter bot informs about the start of a surplus auction

### Context
Our bot currently tweets about each new started collateral auction. 
With the support for surplus auctions that was recently introduced we also want to make sure that the bot tweets about a started surplus auction

### Tasks
- [x] make sure that collateral and surplus auctions are running in independent loops. Meaning that
  - [x] if fetching ones takes longer or throws an error, that shouldn't affect the others
  - [x] the logic for different auction types should not live in the same files
- [x] get live surplus auctions every `DEFAULT_REFETCH_INTERVAL`
- [x] determine which ones are new via `THRESHOLD_FOR_NEW_AUCTIONS`
- [x] send notification if there is a new surplus auction using the same notification logic, using the same `FRONTEND_ORIGIN` logic

### Assets
Example tweet text:
```
Surplus auction with 30000 DAI just started. Follow the link to participate:
https://auctions.makerdao.network/surplus/?network=mainnet&auction=1849
```

aomafarag:
Closes #385.

Screenshot from 2022-08-02 14-40-35

Checklist:


valiafetisov:
No, `fetchedAt` is the time when the auction was fetched, but it's a good question nevertheless. 

@KirillDogadin-std can you please help here and add new `auctionStartDate` property to the surplus auction using [`tau`](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/flap.sol#L127) in another PR? Another solution would be [to subscribe to new `Kick` events](https://docs.ethers.io/v5/api/providers/provider/#Provider--events), instead of refetching them periodically – that would be more precise, but would require even more work and would be harder to test, so I vouch for the first proposal.

KirillDogadin-std:
I agree with the choice of the first option and will open the corresponding issue + PR

bot/src/notify.ts
```diff
@@ -1,17 +1,28 @@
...
-    const text = generateNotificationText(auction);
+const generateNotificationTextSurplus = function (auction: SurplusAuctionActive): string {
+    const url = `${process.env.FRONTEND_ORIGIN}/surplus/?network=${auction.network}&auction=${auction.id}`;
+    return `Surplus auction with ${auction.receiveAmountDAI.toFixed(
+        0
+    )} DAI just started. Follow the link to participate: ${url}`;
+};
+
+export const notifyCollateral = async function (auction: AuctionInitialInfo) {
+    const text = generateNotificationTextCollateral(auction as AuctionInitialInfo);
```
valiafetisov, line 19: Why do you need to apply typecasting here?

bot/src/notify.ts
```diff
@@ -1,17 +1,28 @@
...
     console.info(`notify: content "${text}"`);
     await sendNotification(text);
 };
 
-export default notify;
+export const notifySurplus = async function (auction: SurplusAuction) {
+    const text = generateNotificationTextSurplus(auction as SurplusAuctionActive);
```
valiafetisov, line 25: And here?

bot/src/notify.ts
```diff
@@ -1,17 +1,28 @@
...
+    console.info(`notify: content "${text}"`);
```
valiafetisov, line 26: Please make sure that logging messages are different for different functions

core/test/surplus-test.ts
```diff
@@ -77,9 +77,6 @@ describe('Surplus Auction', () => {
         const auctions = await fetchActiveSurplusAuctions('custom');
         const currentAuction = auctions[0];
         expect(currentAuction.id).to.equal(2328);
-        if (currentAuction.state === 'collected') {
-            throw new Error('auction state can not be "collected"');
-        }
```
valiafetisov, line 82: Why exactly this was removed? What introduced changes caused this auction to become `collected`?

bot/src/index.ts
```diff
@@ -12,20 +13,33 @@ const DEFAULT_REFETCH_INTERVAL = 60 * 1000;
...
         const newAuctions = getNewAuctionsFromActiveAuctions(activeAuctions);
-        newAuctions.map(notify);
+        newAuctions.map(notifyCollateral);
         participate(network, activeAuctions);
     } catch (error) {
         console.error('loop error:', error);
     }
 };
 
+const loopSurplus = async function (network: string): Promise<void> {
```
valiafetisov, line 30: Can you please move those functions into the appropriate files (`surplus.ts` vs `collateral.ts`)?

bot/src/surplus.ts
```diff
@@ -0,0 +1,35 @@
...
+        return isNew && checkIfAuctionIsAlreadyKnown(activeAction);
+    });
+    console.info(`surplus auctions: "${newAuctions.length}" of "${activeActions.length}" auctions are new`);
+
+    newAuctions.map(markAuctionAsKnown);
+    return newAuctions;
+};
+
+export const getAllActiveSurplusAuctions = async function (network: string): Promise<SurplusAuctionActive[]> {
+    const auctions = await fetchActiveSurplusAuctions(network);
```
valiafetisov, line 29: Does this function provides already enriched auctions? Then I would argue to use one underlying function, so that we do less requests every 60 seconds

bot/src/auctions.ts
```diff
@@ -1,6 +1,8 @@
 import type { AuctionInitialInfo } from 'auctions-core/src/types';
```
valiafetisov, line 1: Since we introduce another type of the auctions, we should do a bit or refactoring. As proposed, can you please rename this file to `collateral` or even to `auctions/collateral`, moving surplus to the same folder?

valiafetisov (approved):

valiafetisov (approved):

PR #391, opened 2022-08-02
feat: Add the Surplus start date in core

### LINKED ISSUE #390: Add Surplus auction start property
### Goal

Have surplus auction object contain starting date

### Context

in https://github.com/sidestream-tech/unified-auctions-ui/pull/389 it was requested to introduce the value that represents the satarting time of the auction.

### Tasks

- [ ] Add field to active surplus auctions that contains the date when it was started / kicked.

KirillDogadin-std:
Closes #390

Checklist:

### DIFF
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -62,7 +62,7 @@ const _getContractAuctionDuration = async (network: string) => {
 const getContractAuctionDuration = memoizee(_getContractAuctionDuration, {
     maxAge: 24 * 60 * 60 * 1000,
     promise: true,
-    length: 2,
+    length: 1,
 });
 
 export const fetchSurplusAuctionByIndex = async function (
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -54,7 +54,8 @@ const getAuctionState = async (network: string, earliestEndDate: Date, greatestB
     return 'just-started';
 };
 
-const _getContractAuctionDuration = async (contract: Contract) => {
+const _getContractAuctionDuration = async (network: string) => {
+    const contract = await getContract(network, 'MCD_FLAP');
     return await contract.tau();
 };
 
@@ -69,7 +70,7 @@ export const fetchSurplusAuctionByIndex = async function (
     auctionIndex: number
 ): Promise<SurplusAuction> {
     const contract = await getContract(network, 'MCD_FLAP');
-    const auctionDuration = await getContractAuctionDuration(contract);
+    const auctionDuration = await getContractAuctionDuration(network);
     const auctionData = await contract.bids(auctionIndex);
     const isAuctionCollected = new BigNumber(auctionData.end).eq(0);
     const fetchedAt = new Date();
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -54,15 +54,24 @@ const getAuctionState = async (network: string, earliestEndDate: Date, greatestB
     return 'just-started';
 };
 
+const _getContractAuctionDuration = async (contract: Contract) => {
+    return await contract.tau();
+};
+
+const getContractAuctionDuration = memoizee(_getContractAuctionDuration, {
+    maxAge: 24 * 60 * 60 * 1000,
+    promise: true,
+    length: 2,
+});
+
 export const fetchSurplusAuctionByIndex = async function (
     network: string,
     auctionIndex: number
 ): Promise<SurplusAuction> {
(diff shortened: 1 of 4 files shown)


core/src/surplus.ts
```diff
@@ -59,8 +59,10 @@ export const fetchSurplusAuctionByIndex = async function (
     auctionIndex: number
 ): Promise<SurplusAuction> {
     const contract = await getContract(network, 'MCD_FLAP');
+    const auctionDuration = await contract.tau();
     const auctionData = await contract.bids(auctionIndex);
-    const isAuctionCollected = new BigNumber(auctionData.end).eq(0);
+    const createdAt = await auctionData.end;
+    const isAuctionCollected = new BigNumber(createdAt).eq(0);
```
valiafetisov, line 65: `auctionData.end` is not `createdAt`. Moreover, there is already a property called `auctionEndDate` that contains date that you need to calculate `auctionStartDate`

KirillDogadin-std: done

core/src/surplus.ts
```diff
@@ -54,11 +54,22 @@ const getAuctionState = async (network: string, earliestEndDate: Date, greatestB
     return 'just-started';
 };
 
+const _getContractAuctionDuration = async (contract: Contract) => {
```
valiafetisov, line 57: The contract is an object with functions so it doesn't cache well. Likely, this function will still be executed every time or wouldn't expire after switching the network. Instead, I suggest to use `network` as a parameter
```suggestion
const _getSurplusAuctionsDurationInSeconds = async (network: string) => {
```

core/src/surplus.ts
```diff
@@ -54,11 +54,22 @@ const getAuctionState = async (network: string, earliestEndDate: Date, greatestB
...
+    return await contract.tau();
+};
+
+const getContractAuctionDuration = memoizee(_getContractAuctionDuration, {
+    maxAge: 24 * 60 * 60 * 1000,
+    promise: true,
+    length: 2,
```
valiafetisov, line 64: This indicates number of arguments it needs to cache
```suggestion
    length: 1,
```

core/test/surplus-test.ts
```diff
@@ -42,8 +42,10 @@ describe('Surplus Auction', () => {
     it('fetches active auctions', async () => {
         const auctions = await fetchActiveSurplusAuctions('custom');
         expect(auctions.length).to.equal(5);
-        expect(auctions[0].id).to.equal(2328);
-        expect(auctions[0].state).to.equal('have-bids');
+        const auction = auctions[0] as SurplusAuctionActive;
+        expect(auction.id).to.equal(2328);
+        expect(auction.state).to.equal('have-bids');
+        expect(auction.auctionStartDate.toISOString()).to.equal('2022-01-26T01:31:58.000Z');
```
valiafetisov, line 48: Do you want to at least compare that end date is later than the start date? (cause this number doesn't make any meaning to the reader)

core/src/surplus.ts
KirillDogadin-std: > The contract is an object with functions so it doesn't cache well.

do you have this claim by experience or is there a handy-dandy article i could read to understand why it is so?

valiafetisov: You probably know that in js there is a difference between primitive values (string, number, boolean, undefined, symbol, null, etc) and objects (objects, arrays, functions, classes, etc) in which `=` or `===` operators [work differently](https://flaviocopes.com/difference-primitive-types-objects/) and basically compare references instead of values.

I can't find what `memoizee` using for serialisation of values, therefore I assume that they compare references to `contract` until you set `primitive` mode to true and write a serialiser. If contract is created with the new keyword every time, then it will have different reference and the cache wouldn't work. If the `contract` modifies itself every time it's called (eg updates count inside itself) then even serialisation wouldn't help. Therefore, the golden rule is to avoid caching functions that accept non-primitive values.

valiafetisov (approved):

core/src/surplus.ts
```diff
@@ -54,11 +54,23 @@ const getAuctionState = async (network: string, earliestEndDate: Date, greatestB
     return 'just-started';
 };
 
+const _getContractAuctionDuration = async (network: string) => {
```
valiafetisov, line 57: As suggested, it would be better to call this using `surplus` in its name
```suggestion
const _fetchSurplusAuctionsDurationInSeconds = async (network: string) => {
```

PR #398, opened 2022-08-08
feat: Introduce Debt Auction Types

KirillDogadin-std:
Related to #387 

As requested by @zoey-kaiser the types are extracted into separate PR so that we can discuss and alter it more efficiently.


zoey-kaiser:
@KirillDogadin-std, how did you come to call the types `CompensationAuction`?

KirillDogadin-std:
If the system goes out of balance into either (surplus, debt) direction, we have to compensate this disbalance. Hence 'compensation' is the name. 

If you have the better option that is better related to the economical terminology, i would appreciate if you could propose the alternative.

core/src/types.ts
```diff
@@ -205,4 +214,70 @@ export declare interface SurplusTransactionFees {
...
+export declare interface DebtAuctionEnriched extends DebtAuctionActive {
+    nextMaximumLotReceived: BigNumber;
+    marketUnitPrice: BigNumber | string;
+    marketUnitPriceToUnitPriceRatio: BigNumber;
+    unitPrice: BigNumber;
+}
+
+export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, CompensationAuctionTransactionFees {}
+
+export const getCompensationAuctionTransactionFees = async function (
```
zoey-kaiser, line 256: I assume this is in here for testing right? We should remove it tho, as we only want to sort out the typing in this PR.

zoey-kaiser (changes requested):

core/src/types.ts
```diff
@@ -156,53 +156,89 @@ export declare interface CollateralStatus {
...
+    combinedBidFeesEth: BigNumber;
+    allowanceTransactionFeeDai: BigNumber;
+    restartTransactionFeeDai: BigNumber;
+    bidTransactionFeeDai: BigNumber;
+    collectTransactionFeeDai: BigNumber;
+    authTransactionFeeDai: BigNumber;
+    combinedBidFeesDai: BigNumber;
+}
+
+export declare interface SurplusAuctionActive extends CompensationAuctionBase {
```
zoey-kaiser, line 189: The data inside this and `DebtAuctionActive` are nearly identical. Do we want to try and move this into a combined `CompensationAuctionActive` type or do we want to keep them seperated?

KirillDogadin-std: I would prefer to keep them split, because even though there're similarities, I don't think we should mix up as much fields as possible:

 - Debt auctions and surplus auctions are similar, but are (acc. to my perseption) different + unrelated entities and hence the data structures should be more-or-less decoupled.
 - Extracting the fields into the common ancestor usually would have to be a solid design decision. Here i do not see enough incentive: We use different contracts to extract the data about the contracts, the varialbes and methods from the contracts are named differently, ... (maybe more depth of difference that i don't remember now).
 Parsing the data that we receive (e.g. to determine state). So there should be two different functions that extract the info from the contracts. It would be odd to have these 2 different functions initialize the objects that are closely related (e.g. have 8/9 similar fields). In other words, unless initialization is very similar, i don't see the great benefit of putting all the common fields we've come up with into the same interface.

zoey-kaiser: I agree!

zoey-kaiser (approved):

LukSteib (approved):

Issue #392, opened 2022-08-03
Create debt auction mock


LukSteib:
### Goal
Create mock for UI flow of debt auctions

### Context
debt auctions is the third auction type to be integrated in our UI. As always mock should be used to get an understanding and agreement on the planned UX flow.
In this case we also want to use the mock to check what to components to copy over from surplus auctions

### Tasks
- [x] Create Mock
- [x] Discuss 
- [x] Create outline of what needs to be done

LukSteib:
See Mock: https://whimsical.com/makerdao-debt-TLjy6A3d3hV7XphhayiXq

<details><summary>MOCK v1</summary>
<p>

![MakerDAO debt@2x](https://user-images.githubusercontent.com/36229155/182543058-dfb27320-3df8-4ede-97e3-715579fd71e7.png)


</p>
</details>

LukSteib:
- Create basic debt auction page
    - adjust header, extend new navigation modal, adjust landingblock, adjust texts
- Create `DebtAuctionsTable`
    - Copy `SurplusAuctionsTable`
    - to be adjusted columns
        - "Auction amount" -> "Bid amount"
        - "Highest bid" -> "Lowest auction amount"
    - to be adjusted auction states
        - "Collect DAI" -> "Collect MKR"
- Create `DebtAuction` page
    - Copy `SurplusAuction`
    - to be changed table rows
        - "Highest Bid" -> "Bid"
        - "Auction Amount" -> "Auction amount to receive”
        - Change text
        - change button copy
- Create `DebtAuctionRestartPanel`
    - Use `AuctionRestartPanel` and use the `Surplus Auction` version -> no need to change anything
- Create `DebtAuctionBidTransactionTable`
    - Copy `SurplusAuctionBidTransactionTable`
    - to be adjusted rows
        - "Auction Amount" -> "Bid"
        - "Current Highest Bid" -> "Current amount to receive"
        - "Lowest Next Bid" -> "Highest next amount to receive"
        - "The Amount to bid" -> "The amount to receive" (ie. input field)
        - in general: price denotation should be DAI per MKR wherever we currently have MKR per DAI
- Use existing `WalletConnectionCheckPanel`
- Create `WalletdDaiDepositCheckPanel`
    - Take over https://github.com/sidestream-tech/unified-auctions-ui/pull/314 and adapt it in a way to change MKR for DAI
- Use existing `WalletAuthorizationCheckPanel`
[...]

PR #403, opened 2022-08-09
Added `DebtAuctionsTable`

zoey-kaiser:
Screenshot 2022-08-09 at 12 04 48

Checklist:

### DIFF
--- a/frontend/components/auction/debt/DebtAuctionsTable.stories.js
+++ b/frontend/components/auction/debt/DebtAuctionsTable.stories.js
@@ -1,0 +1,44 @@
+import { storiesOf } from '@storybook/vue';
+import faker from 'faker';
+import DebtAuctionsTable from './DebtAuctionsTable';
+import { generateFakeDebtAuctionTransactions } from '~/helpers/generateFakeDebtAuction';
+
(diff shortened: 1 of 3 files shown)


LukSteib (changes requested):
- [x] change denotation of market value column to DAI per MKR 
- [x] not sure whether the case (see S1) will be possible. I expect to have both the fixed bid amount and compensation right away. 

S1:
image

zoey-kaiser:
> * [ ]  not sure whether the case (see S1) will be possible. I expect to have both the fixed bid amount and compensation right away.

Okay! This has less to do with the component and more with the helper generation function. I will change it in this PR and then all the other components will also reflect it once this is merged!

LukSteib:
> Okay! This has less to do with the component and more with the helper generation function. I will change it in this PR and then all the other components will also reflect it once this is merged!

If we cannot determine a value for the compensation let's display `unknown` 
See the discussion here: https://github.com/sidestream-tech/unified-auctions-ui/issues/392#issuecomment-1210265953

valiafetisov (approved):

LukSteib (approved):

Issue #12, opened 2021-10-25
Display currency relationships with FIAT better


zoey-kaiser:
**Goal**
Have the currency relation to FIAT displayed better

**Context**
While using the website, I noticed that there is often a disconnect to Fiat currencies. While experienced users might know, that 1 DAI is around 1 USD, I think many new users may find themselves googling how much a currency is in USD. 

**Tasks**
Extend the `FormatCurrency` component to display the currencies value in USD as a tooltip.

**Mock**
![Screenshot 2021-10-25 at 11 11 41](https://user-images.githubusercontent.com/30908158/138668551-389232da-49e9-48a5-a25d-c48eac856038.png)
![Screenshot 2021-10-25 at 11 12 02](https://user-images.githubusercontent.com/30908158/138668561-8d7108c5-73f7-48d8-8bcc-c29749f598fd.png)
![Screenshot 2021-10-25 at 11 12 16](https://user-images.githubusercontent.com/30908158/138668565-03cff46e-1df7-4239-a137-3e67f3c87136.png)

PR #404, opened 2022-08-09
Collateral onboarding docker container.

KirillDogadin-std:
References https://github.com/sidestream-tech/auction-ui/issues/430

### DIFF
--- a/collateral-onboarding/collateral-onboarding/Dockerfile
+++ b/collateral-onboarding/collateral-onboarding/Dockerfile
@@ -1,0 +1,50 @@
+FROM ubuntu:latest as base-installs
+
+RUN apt-get update && \
+    apt-get -y install curl xz-utils git make jq python3 pip
+RUN pip install requests
+
+FROM base-installs as tool-installs
+
+RUN useradd -s /bin/bash -d /home/blockchain/ -m -G sudo blockchain
+RUN mkdir -m 0755 /nix && chown blockchain /nix
+USER blockchain
+ENV USER="blockchain"
+ENV HOME="/home/${USER}"
+ENV PATH="${HOME}/.nix-profile/bin:${PATH}"
+ENV NIX_PATH="${HOME}/.nix-defexpr/channels"
+RUN curl -L https://nixos.org/nix/install | sh -s -- --no-daemon
+RUN echo kek
+RUN curl -L https://dapp.tools/install | sh
+
+FROM tool-installs as base-clones
+
+USER blockchain
+RUN mkdir /home/blockchain/spells
+WORKDIR /home/blockchain/spells
+COPY ./onboard.sh .
+RUN git clone https://github.com/makerdao/spells-mainnet.git
+RUN git clone https://github.com/makerdao/exchange-callees.git
+
+FROM base-clones as cast-spell-base
+
+ARG ETH_RPC_URL
+ENV ETH_RPC_URL=${ETH_RPC_URL}
+ARG ETH_GAS
+ENV ETH_GAS=${ETH_GAS}
+ARG ETH_FROM
+ENV ETH_FROM=${ETH_FROM}
+
+
+ENV USER="blockchain"
+ENV HOME="/home/${USER}"
+ENV PATH="${HOME}/.nix-profile/bin:${PATH}"
+
+USER blockchain
+RUN export ETH_GAS_PRICE=$(seth --to-wei 100 "gwei")
+WORKDIR /home/blockchain/spells/spells-mainnet
+RUN echo "export ETH_RPC_ACCOUNTS=yes" >> /home/blockchain/.sethrc
+
+FROM cast-spell-base as cast-spell-execute
+
+CMD ["../onboard.sh"]
--- a/collateral-onboarding/collateral-onboarding/onboard.sh
+++ b/collateral-onboarding/collateral-onboarding/onboard.sh
@@ -7,35 +7,4 @@ dapp create DssSpell
 cd ~/spells/exchange-callees
 dapp update
 make
-
-# Deploy mainnet calless
-UniswapV2CalleeDai=$(dapp create UniswapV2CalleeDai \
-  0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D \
-  0x9759A6Ac90977b93B58547b4A71c78317f391A28)
-
-UniswapV2LpTokenCalleeDai=$(dapp create UniswapV2LpTokenCalleeDai \
-  0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D \
-  0x9759A6Ac90977b93B58547b4A71c78317f391A28)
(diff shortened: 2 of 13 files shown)


collateral-onboarding/collateral-onboarding/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+RUN useradd -s /bin/bash -d /home/blockchain/ -m -G sudo blockchain
+RUN mkdir -m 0755 /nix && chown blockchain /nix
+USER blockchain
+ENV USER="blockchain"
+ENV HOME="/home/${USER}"
+ENV PATH="${HOME}/.nix-profile/bin:${PATH}"
+ENV NIX_PATH="${HOME}/.nix-defexpr/channels"
+RUN curl -L https://nixos.org/nix/install | sh -s -- --no-daemon
+RUN echo kek
+RUN curl -L https://dapp.tools/install | sh
```
valiafetisov, line 18: Does this work on the M1 now?

collateral-onboarding/collateral-onboarding/onboard.sh
```diff
@@ -0,0 +1,41 @@
...
+dapp create DssSpell
+
+cd ~/spells/exchange-callees
+dapp update
+make
+
+# Deploy mainnet calless
+UniswapV2CalleeDai=$(dapp create UniswapV2CalleeDai \
+  0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D \
+  0x9759A6Ac90977b93B58547b4A71c78317f391A28)
```
valiafetisov, line 14: Where is this code coming from? Why is it copied over to our repo and not cloned/executed directly? Can't you run `https://github.com/makerdao/exchange-callees/blob/3c4055571c0ab17efbd9e7ed809862bc47d8c2fd/Makefile#L7` this instead?

collateral-onboarding/docker-compose.yml
```diff
@@ -0,0 +1,19 @@
...
+    environment:
+      - ETH_RPC_URL=http://hardhat:8545
+      - ETH_GAS=5000000
+      - ETH_FROM=0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
+  hardhat:
+    platform: linux/amd64
+    build:
+      context: ./hardhat
+    expose:
+      - 8545
```
valiafetisov, line 17: Can you please also expose it to the outside, since then we can test it via UI/bot?

collateral-onboarding/collateral-onboarding/onboard.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/bash
+cd ~/spells/spells-mainnet
+dapp update
+make
+dapp create DssSpell
```
valiafetisov, line 5: This `make all` only builds them, but doesn't execute
What `dapp create DssSpell` actually do? Can 1+ particular spell be executed instead?

collateral-onboarding/hardhat/Dockerfile
```diff
@@ -0,0 +1,27 @@
...
+COPY ./package-lock.json .
+
+COPY ./ /hardhat
+
+RUN npm ci
+
+FROM $NODE_VERSION AS development
+
+ARG ALCHEMY_URL
+ENV ALCHEMY_URL=${ALCHEMY_URL}
```
valiafetisov, line 20: Please provide rood readme with instructions on how to start, the list of required env vars, and any other magic required (eg M1-related)

collateral-onboarding/hardhat/hardhat.config.js
```diff
@@ -0,0 +1,27 @@
...
+module.exports = {
+    solidity: '0.8.4',
+    networks: {
+        hardhat: {
+            chainId: 1337,
+            forking: {
+                url: ALCHEMY_URL,
+                // Few blocks before WSTETH-A is taken at 14052147,
+                // https://etherscan.io/address/0x49a33a28c4c7d9576ab28898f4c9ac7e52ea457a
+                blockNumber: 14052140,
```
valiafetisov, line 22: Why this specific block number? I imagine that we should actually fork without it to be on the latest state of the maker network

Otherwise, for now, you can take the block number where `MCD_CLIP_WSTETH_A` didn't exist yet (ie [`13426393 - 1`](https://etherscan.io/address/0x49a33a28c4c7d9576ab28898f4c9ac7e52ea457a#internaltx)) and try to execute the spell which introduced it, then check if `MCD_CLIP_WSTETH_A` exists in the `chainlog` contract (which should at this point)

collateral-onboarding/collateral-onboarding/Dockerfile
KirillDogadin-std: yes

collateral-onboarding/hardhat/hardhat.config.js
KirillDogadin-std: well, there's no reason for a specific block number: 

- in order to test that e.g. the MCD_CLIP_WSTETH_A , i would also have to checkout different commit there
- because currently there is no collateral being introduced by the latest version of the spell.

 also why i claimed that this runs without failure and not that it works.

valiafetisov: > there's no reason for a specific block number
> i would also have to checkout different commit there

Well, I still think that it makes sense to fork on a specific block number and maybe (as you suggest above) run a spell on a specific commit, in order to validate that contract that was not available is available after the execution

valiafetisov: So the takeaway here is the additional test setup that might need to be created

collateral-onboarding/collateral-onboarding/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
```
valiafetisov, line 19: Would this also run without errors?

```suggestion
RUN curl -L https://dapp.tools/install | sh
RUN seth --version
```

KirillDogadin-std:
```
 => ERROR [tool-installs 6/6] RUN seth --version                                                                                                                                                                                                                                                                         0.4s
------
 > [tool-installs 6/6] RUN seth --version:
#12 0.215 /bin/sh: 1: seth: not found
------
executor failed running [/bin/sh -c seth --version]: exit code: 127

valiafetisov: Thanks! From this I assume that `dapp` tools did not correctly install despite finishing without error code. Correct me if I am missing something here

valiafetisov:
This issue is closed via adding a complete instruction in #506 

https://github.com/sidestream-tech/unified-auctions-ui/blob/cbf572753d9dde45d03ff63be8a62cbe03f70548/core/README.md?plain=1#L49-L83

Issue #416, opened 2022-08-15
Investigate `chaos-proxy`'s failure to run


aomafarag:
### Goal

Get `chaos-proxy` up and running again.

### Context

Recently, I tried running `chaos-proxy` and ran into the following error (screenshot under Assets): `Chaoslabs API request failed with error: Request failed with status code 400`.

### Assets

```bash
aomafarag@HP-Pavilion-Notebook ~/sidestream/unified-auctions-ui/chaos-proxy (keeper-surplus)$ npm run dev

> dev
> npx ts-node src/index.ts

✔ Please select the simulation › Single Flapper
Starting simulation "[secret]"...
/home/aomafarag/sidestream/unified-auctions-ui/chaos-proxy/src/api.ts:20
        throw new Error(`Chaoslabs API request failed with error: ${error instanceof Error ? error.message : error}`);
              ^
Error: Chaoslabs API request failed with error: Request failed with status code 400
    at executeAPIrequest (/home/aomafarag/sidestream/unified-auctions-ui/chaos-proxy/src/api.ts:20:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async startSimulationInstance (/home/aomafarag/sidestream/unified-auctions-ui/chaos-proxy/src/api.ts:25:22)
    at async getSimulationUrl (/home/aomafarag/sidestream/unified-auctions-ui/chaos-proxy/src/api.ts:51:24)
    at async /home/aomafarag/sidestream/unified-auctions-ui/chaos-proxy/src/index.ts:22:23
```

### Tasks

- [ ] Investigate

valiafetisov:
Thanks for submitting this, @aomafarag! Please next time prefer posting actual logs over the screenshot of the code. The reasons are:
- code is copy-pastable and searchable
- you can remove sensitive information from the code that is normally stored in secrets (in this case, simulation id is exposed)

valiafetisov:
Subject-wise:
- [ ] Please check that you're using the latest values from the secrets
- [ ] Please check whether they've updated API/simulation ids https://github.com/sidestream-tech/auction-ui/issues/421#issuecomment-1168618091
  - [ ] Update simulation ids in the secrets
  - [ ] Or update the code via a separate PR

valiafetisov:
Does the updated link works?

valiafetisov:
Outdated, closed via https://github.com/sidestream-tech/unified-auctions-ui/pull/543

PR #417, opened 2022-08-16
Extend keeper to bid on surplus auctions

### LINKED ISSUE #415: Extend the Keeper to bid on Surplus Auctions
### Goal
Our current implementation of the keeper only works with Collateral Auctions. This PR aims to incorporate the Surplus Auctions into the flow as well. 

### Context
The current keeper implementation has two main ingress points. The first is a function called [`setupKeeper`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/bot/src/keeper.ts#L10), which runs once when the bot is started. Inside the function we do the following steps:
- Check if the user set a `WALLET_PRIVATE_KEY`
- Setup the Signer (wallet with perms to execute transactions)
- Verify the wallet works
- Check the `KEEPER_MINIMUM_NET_PROFIT_DAI`

You might have to either modify the `KEEPER_MINIMUM_NET_PROFIT_DAI` or add a second variable for Surplus Auctions, as we do want to give the user the option to define how much profit they want to make. 
This initial Setup Keeper function can stay as it is. When you try to execute Surplus Auctions, you will be able to use the same signer, defined here.

The second ingress point is the [`participate`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/bot/src/keeper.ts#L131) function. I will be going into less detail here, as every auction type has a different bidding flow, meaning that you will not be able to copy the logic from here. However I would recommend moving this function into a new file (called `collateralKeeper` for example) and renaming the function to `participateInCollateralAuctions`. 

In order to introduce the bidding on SurplusAuctions, you will need to use the function [`bidToSurplusAuction`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/surplus.ts#L143), however there are some caveats here we need to examine! Before I get into them exactly, I will provide a rough outline of what the participate function would need to look like:

```js
export const participateInSurplusAuction (surplusAuction: SurplusAuction) {
  // check if setupKeeper hasn't run
  if (!isSetupCompleted) {
    return;
  }

  const signer = await getSigner(network);

  // Check the SurplusAuction State, to ensure you can actually bid on it

  // enrich the SurplusAuction into a SurplusAuctionTransaction

  // check the end auction profit (combine end profit and fees to see if it is profitable)

  // ensure the auction clear profit (without transaction fees) meets the set requirements

  // ensure the wallet has enough MKR allowance to bid on the auction

  // if the allowance is not enough, set it to unlimited or to the amount needed (up to you)

  // bid on the auction
}
```
Whenever you complete one of the checks (say allowance check) and set new values, be sure to re run the participate script!

Now some more explanations:

Firstly, you will need to enrich the `SurplusAuction` to a `SurplusAuctionTransaction`. You can use the [enrich function](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/surplus.ts#L204) for this. The information we gain from this enrichment is important as we need to...

- Know the next minimum bid
- Add any Transaction fees
- Check if the net profit is still high enough to cover the `KEEPER_MINIMUM_NET_PROFIT_DAI`

Extra: If you want to, you can also write a second function that collects the `ready-for-collection` auctions, that were won by the keepers wallet. When you do the auction state check, you can see if the auction is ready for collection. If it is, return the participate function and call your new function. You can retrieve the keeper wallet address by running: 

```js
const signer = await getSigner(network);
const address = await signer.getAddress();
```

aomafarag:
Closes #415.

Checklist:

hardhat block number used for testing: 14052140


bot/src/keeper/setup.ts
```diff
@@ -0,0 +1,28 @@
+import getSigner, { createSigner, setSigner } from 'auctions-core/src/signer';
+import { KEEPER_MINIMUM_NET_PROFIT_DAI, KEEPER_WALLET_PRIVATE_KEY } from '../variables';
```
zoey-kaiser, line 2: Do we maybe want to add a separate variable for Surplus Auctions? That way the user can define minimums for each?

Currently the keeper does not run if `KEEPER_MINIMUM_NET_PROFIT_DAI` is not set. Maybe we want to introduce the following behavior:

`KEEPER_MINIMUM_NET_PROFIT_DAI` Is set = Keeper runs for collaterals. 

`KEEPER_MINIMUM_NET_PROFIT_DAI` is not set = Keeper does not run for collaterals.

And the same for the Surplus Auctions as well. This will allow the user to also determine which auctions the keeper should run on.

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,109 @@
...
+        return;
+    }
+
+    const signer = await getSigner(network);
+
+    // enrich the auction with more numbers
+    const auctionTransaction = await enrichSurplusAuction(network, auction);
+
+    // check if auction became inactive or finished
+    if (auctionTransaction.state === 'ready-for-collection') {
```
zoey-kaiser, line 23: Do we want to check if the user is the winner and then collect the auction in this case?

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,109 @@
...
+        console.info(
+            `keeper: surplus auction "${auction.id}" clear profit is ${clearProfit.toFixed(
+                0
+            )} DAI after transaction fees, checking wallet MKR allowance`
+        );
+    }
+
+    const walletAddress = await signer.getAddress();
+
+    // check the wallet's MKR allowance
```
zoey-kaiser, line 62: Don't we want to check how much MKR in VAT the wallet holds before we set the allowance?

If not enough is present in the VAT we can also deposit any MKR into the VAT, the wallet holds.

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,109 @@
...
+        await setAllowanceAmountMKR(network, walletAddress, auctionTransaction.bidAmountMKR);
+        await checkAndParticipateIfPossible(network, auction);
+        return;
+    } else {
+        console.info('keeper: wallet MKR allowance is within limits, moving on to execution');
+    }
+
+    // bid on the Auction
+    console.info(`keeper: surplus auction "${auctionTransaction.id}": attempting bid execution`);
+    await bidToSurplusAuction(network, auctionTransaction.id, auctionTransaction.bidAmountMKR);
```
zoey-kaiser, line 81:
```suggestion
    await bidToSurplusAuction(network, auctionTransaction.id, auctionTransaction.nextMinimumBid);
```

You used `bidAmountMKR` for a lot of the checks, however this is the current highest bid. Please change it to `nextMinimumBid`, as we already calculate what the user would need to bid.

zoey-kaiser (changes requested):

LukSteib:
> I haven't been able to test the bot due to the chaos-proxy issue (see https://github.com/sidestream-tech/unified-auctions-ui/issues/416)

You should still be able to test the bot via a hardhat fork, right?

bot/src/keeper/setup.ts
zoey-kaiser: Question: What would the new variable look like?

LukSteib:
@zoey-kaiser in the [above comment](https://github.com/sidestream-tech/unified-auctions-ui/pull/417#issuecomment-1216625683) you are referencing the debt auction logic. Since this issue is focused on the keeper extension for surplus auction this might be missleading. 
So to make that clear: In the surplus auction we are bidding with increasing amounts of MKR on a fixed amount of DAI that is auctioned off.

Now, from a keeper operator (aka user) perspective I would argue for the following features: (Please judge these from both a logical point of view and a technical feasibility point of view) 

General:
* Being able to determine in which auction type the keeper should participate in. I can see two approaches to that:
  * 1) having one env variable that allows to pass a list of auction types
  * 2) having one env variable per auction type as boolean that allows to set participation

Surplus auction related variable (equivalent to the `KEEPER_MINIMUM_NET_PROFIT_DAI`): 
* Being able to specify the lowest accepted market difference threshold 
  * read: I want the keeper to participate in the surplus auction as long as the auction price is X % below the market price (ie. the price on uniswap)
* advanced feature to safe tx cost (for a later stage): Being able to specify a range for the accepted market difference
  * read: I want the keeper to participate in the surplus auction as long as the auction price is between Y % and X% below the market price (ie. the price on uniswap)
* Rationale: Since the transaction outcome will always be the same (fixed auction amount of DAI), the denominating factor wether participation in the surplus auction is profitable or not is the market difference between the auction price and the market price. Hence I would propose to offer the possibility to determine this market difference via a dedicated env variable
  * Downside: I assume that the above would not factor in transaction costs. But for me it would be fine to neglect them for now

valiafetisov:
Since there can be more variables related to different auction types, I suggest to name them like
```
KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI
```
Instead of adding auction type to the end.

In the end we can also have a number of vars to enable specific functionality
```
KEEPER_COLLATERALS = true
KEEPER_SURPLUS = false
KEEPER_VAULTS = false
```
and prevent from fetching extra information

valiafetisov:
> We would calculate the gross `profit` as follows: `(1 / (marketUnitPrice - unitPrice)) * bidAmountMKR`

I suppose that instead of `bidAmountMKR` you should use "next possible bid"

> we proceed only when `profit` is positive

Keeper should bid on the auction when `net profit` is higher than the threshold `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI` set by the user. Checking `gross profit` to be positive is not required (compared to the collateral auction, where it is enforced by the contract itself, so we have to check it in order to avoid execution errors)

hardhat/hardhat.config.js
```diff
@@ -19,7 +19,7 @@ module.exports = {
                 url: ALCHEMY_URL,
                 // Few blocks before WSTETH-A is taken at 14052147,
                 // https://etherscan.io/address/0x49a33a28c4c7d9576ab28898f4c9ac7e52ea457a
-                blockNumber: 14078340,
```
zoey-kaiser, line 22: I would not hardcode this change! Maybe document the new BlockNumber in this PR for later, but keep the original one. 

@LukSteib what do you think?

zoey-kaiser (changes requested):

hardhat/hardhat.config.js
LukSteib: Yes, fine for me

LukSteib (changes requested):
When trying to run the keeper with `RPC_URL` specified to `http://127.0.0.1:8545`, I am experiencing the following problems on startup:
- [ ] get a warning that boolean flags and profit variables are not set properly although I've specified `KEEPER_SURPLUS` to `true` and set a value for `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI` -> what is expected here? 
- [ ] Getting the error that signer cannot be created -> can you point me to the cause of this problem? 


 WARN  keeper: boolean flags and profit variables are not set properly or are combined incorrectly, keeper will not run   08:15:01

ℹ keeper: "KEEPER_PREAUTHORIZE" is true. Attempting to authorize wallet and collaterals: "ETH-A, ETH-C, ETH-B, WSTETH-A"  08:15:01

 ERROR  No signer has been created for the "custom" network                                                               08:15:01

  at getSigner (/Users/luksteib/Github/unified-auctions-ui/core/src/signer.ts:20:11)
  at executePreAuthorizationsIfRequested (src/authorisation.ts:62:44)
  at runMicrotasks (<anonymous>)
  at async start (src/index.ts:30:3)

bot/src/keeper/setup.ts
```diff
@@ -0,0 +1,47 @@
...
+    if (!KEEPER_WALLET_PRIVATE_KEY) {
+        console.warn('keeper: KEEPER_WALLET_PRIVATE_KEY variable is not set, keeper will not run');
+        return;
+    }
+    if (
+        !(KEEPER_COLLATERAL && !Number.isNaN(KEEPER_COLLATERAL_MINIMUM_NET_PROFIT_DAI)) ||
+        !(KEEPER_SURPLUS && !Number.isNaN(KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI))
+    ) {
+        console.warn(
+            'keeper: boolean flags and profit variables are not set properly or are combined incorrectly, keeper will not run'
```
zoey-kaiser, line 24: Can we please reference the environment variables by name here?

This way a user does not need to go searching in the README what the boolean flags are, but can see what they are named just by reading your error message. (Might make sense to split the error message or make a longer one, up to you)

zoey-kaiser:
> * [ ]  Getting the error that signer cannot be created -> can you point me to the cause of this problem?

This is not an issue with you, but with the code. The following issue occurred: 

- The setupKeeper function could not find the needed environment variables, therefore it printed the warning message and returned (not creating the signer)
- For some reason, altho the setup function was canceled, another function attempted to call `getSigner`, which returned an error, as the setupKeeper was returned before a signer could be created. 

Please investigate further @aomafarag. I would start with looking at the Pre-Auth steps, I feel like we are not waiting properly to execute those functions until setupKeeper has completed (may also just be a old error, that we never caught and noticed for the first time now)

zoey-kaiser (dismissed):

zoey-kaiser:
I implemented all the changes I saw fit. Please ensure to set a minimum of the following Environment variables when testing the keeper:

```
RPC_URL="http://127.0.0.1:8545" // for localhost testing

KEEPER_WALLET_PRIVATE_KEY="PRIVATE_KEY"

KEEPER_COLLATERAL=true
KEEPER_SURPLUS=true
KEEPER_PREAUTHORIZE=false

KEEPER_COLLATERAL_MINIMUM_NET_PROFIT_DAI=1
KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI=1
```

Also make sure to change the blocknumber in `/hardhat/hardhat.config.js` to `14052140`

The only issue I have run into while testing is that the accounts on the hardhat do not hold any MKR, resulting them in not being able to bid on any surplus auctions. Can we give the accounts MKR or is there another way we can test the keeper?

LukSteib (approved):
> Can we give the accounts MKR or is there another way we can test the keeper?

One can do that manually via the UI running in parallel and there facilitating the required swaps in order to obtain MKR

---
Testing documentation: 

* keeper is properly fetching surplus auctions (check done via comparison with UI)
* keeper not able to bid if there are not enough MKR funds in wallet for placing the `NextPossibleBid`

```
ℹ surplus auctions: "0" of "5" auctions are new                                                                           13:43:25
ℹ keeper: surplus auction "2328" clear profit is 19864.60 DAI after transaction fees, checking wallet MKR balance         13:43:26
ℹ keeper: wallet MKR balance (0.00) is less than min bid amount (16.94) for surplus auction "2328"                        13:43:26
ℹ keeper: surplus auction "2327" has already finished                                                                     13:43:26
ℹ keeper: surplus auction "2326" has already finished                                                                     13:43:27
ℹ keeper: surplus auction "2325" has already finished                                                                     13:43:27
ℹ keeper: surplus auction "2324" has already finished
```

* Once keeper wallet is sufficiently funded with MKR auction execution works including setting the allowance

```
ℹ surplus auctions: found "5" auctions ("2328", "2327", "2326", "2325", "2324") on "custom" network                       13:46:25
ℹ surplus auctions: "0" of "5" auctions are new                                                                           13:46:25
ℹ keeper: surplus auction "2328" clear profit is 19864.60 DAI after transaction fees, checking wallet MKR balance         13:46:26
ℹ keeper: wallet MKR balance is within limits, checking wallet MKR allowance                                              13:46:26
ℹ keeper: wallet MKR allowance (0.00) is less than desired bid amount (16.94), increasing wallet MKR allowance            13:46:26
ℹ transaction b75ccc0d-1d00-4818-987f-e386a216f687: Transaction is preparing to be mined...                               13:46:26
ℹ transaction b75ccc0d-1d00-4818-987f-e386a216f687: Transaction is pending to be mined...                                 13:46:26
ℹ transaction b75ccc0d-1d00-4818-987f-e386a216f687: Transaction was mined into block "14078350"                           13:46:26
ℹ keeper: surplus auction "2328" clear profit is 19864.60 DAI after transaction fees, checking wallet MKR balance         13:46:26
ℹ keeper: wallet MKR balance is within limits, checking wallet MKR allowance                                              13:46:26
ℹ keeper: wallet MKR allowance is within limits, moving on to execution                                                   13:46:26
ℹ keeper: surplus auction "2328": attempting bid execution                                                                13:46:26
ℹ transaction 066a3421-59ef-4e00-9604-aaad25a2058d: Transaction is preparing to be mined...                               13:46:26
ℹ transaction 066a3421-59ef-4e00-9604-aaad25a2058d: Transaction is pending to be mined...                                 13:46:27
ℹ transaction 066a3421-59ef-4e00-9604-aaad25a2058d: Transaction was mined into block "14078351"                           13:46:27
ℹ keeper: surplus auction "2328" was succesfully executed
```

-> Approved as functionally works IMO 

--- 
What we should take with us and adjust (can be done in seperate issue)
- [ ] Provide a better explanation for the `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI` as the underlying calculation is non trivial and it's the dependent variable for the keeper to make a bid or not

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,131 @@
...
+import { KEEPER_SURPLUS, KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI } from '../variables';
+import { isSetupCompleted } from './setup';
+
+const currentlyExecutedAuctions = new Set();
+
+const checkAndParticipateIfPossible = async function (network: string, auction: SurplusAuctionActive) {
+    // check if setupKeeper hasn't run
+    if (!isSetupCompleted) {
+        return;
+    }
```
valiafetisov, line 17: This is not what it does, it only checks if `isSetupCompleted` (which is function) is not falsy, which is never the case

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,131 @@
...
+
+    if (!KEEPER_SURPLUS || Number.isNaN(KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI)) {
+        return;
+    }
```
valiafetisov, line 21: You should not need to put those checks into every function. There should be a single place where it is checked – previously this place was setup function

bot/README.md
```diff
@@ -27,22 +27,25 @@ $ npm run start
...
+-   `KEEPER_*`: (optional) set of env variables to enable keeper bot:
+    -   `KEEPER_WALLET_PRIVATE_KEY`: (required) The wallet private key (https://metamask.zendesk.com/hc/en-us/articles/360015289632-How-to-Export-an-Account-Private-Key)
+    -   `KEEPER_COLLATERAL`: (required to bid on collateral auctions) A boolean flag that determines whether the keeper should bid on collateral auctions
+    -   `KEEPER_SURPLUS`: (required to bid on surplus auctions) A boolean flag that determines whether the keeper should bid on surplus auctions
+    -   `KEEPER_COLLATERAL_MINIMUM_NET_PROFIT_DAI`: (required if `KEEPER_COLLATERAL` is set to `true`) The minimum net profit a collateral auction must yield before the keeper automatically bids on it. Can be negative if one is willing to spend ETH on transaction fees to keep DAI stable
+    -   `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI`: (required if `KEEPER_SURPLUS` is set to `true`) The minimum net profit a surplus auction must yield before the keeper automatically bids on it. Can be negative if one is willing to spend ETH on transaction fees to keep DAI stable
```
valiafetisov, line 39: You marked it as required, but actually never require it

bot/src/keeper/surplus.ts
```diff
@@ -0,0 +1,131 @@
...
+    const walletAddress = await signer.getAddress();
+
+    // enrich the auction with more numbers
+    const auctionTransaction = await enrichSurplusAuction(network, auction);
+
+    // check if auction became inactive or finished
+    if (auctionTransaction.state === 'ready-for-collection') {
+        console.info(`keeper: surplus auction "${auction.id}" has already finished`);
+        // check if the current user is the winner
+        if (auctionTransaction.receiverAddress === walletAddress) {
```
valiafetisov, line 33: This should be checked against the same case since it can (and it is in practice) have different cases stored on the chain and derived by `ethers` from the private key

valiafetisov:
I would be happy to receive another functional review on twitter, collateral keeper and surplus keeper functionality, since this PR affects all of it in some way.

Regarding surplus keeper: as far as I remember, initially surplus auction doesn't have bids, meaning that `nextMinimumBid` is `0` MKR? This edge case was not covered by the keeper yet. I suppose that in that case
- `nextMinimumBid` should be undefined instead of being zero
- the keeper should come up with another `nextBid` strategy that is sufficient for to reach `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI` and bid that instead
- the frontend should also be checked against this edge case

valiafetisov:
> meaning that `nextMinimumBid` is 0 MKR

Looking at the calculation of `netProfit` for the dept auctions I can see that in case `nextMinimumBid` is 0, `netProfit` will become 0 itself, so it's technically safe in case `KEEPER_SURPLUS_MINIMUM_NET_PROFIT_DAI` is set to a positive number, but misleading in case it is negative. So the issue is less critical

LukSteib (approved):
Review after latest changes: 
* Surplus participation ✔️
* collateral participation ✔️
* haven't checked the twitter bot (tbd); but would still proceed. 

---
Great edge case! Let's take that with us and tackle in a dedicated issue.

PR #423, opened 2022-08-22
fix: Update Meta Tags to properly display links on Twitter

zoey-kaiser:
Checklist:


LukSteib (dismissed):
- [ ] Besides the not properly rendered image I would also like to see the actual link back in our tweet that point to the respective auction. Missing right now and the only way to access our UI is via clicking the preview (see example below)

image

zoey-kaiser:
I updated the card type. I believe that if we set the card type to "App", it will still display the link. See example here:
https://twitter.com/jaysonst/status/480024621454856192

frontend/nuxt.config.js
```diff
@@ -2,6 +2,7 @@ import pkg from './package.json';
 
 const SITE_TITLE = 'Unified Auctions';
 const SITE_DESCRIPTION = `The "${SITE_TITLE}" is the portal to all Maker Auctions related services. Easily interact with the Maker Protocol through streamlined interfaces, inform yourself about how the protocol works and receive updates on current auctions.`;
+const TWITTER_HANDEL = '@MakerDAO_SAS';
```
valiafetisov, line 5: typo
```suggestion
const TWITTER_HANDLE = '@MakerDAO_SAS';
```

frontend/nuxt.config.js
```diff
@@ -38,8 +39,12 @@ export default {
             { property: 'og:type', content: 'website' },
             { property: 'og:description', content: SITE_DESCRIPTION },
             { property: 'og:image', content: '/preview.jpeg' },
-            { name: 'twitter:card', content: 'summary_large_image' },
             { name: 'theme-color', content: '#1aab9b' },
+            { name: 'twitter:card', content: 'app' },
+            { name: 'twitter:site', content: TWITTER_HANDEL },
+            { name: 'twitter:title', content: SITE_TITLE },
+            { name: 'twitter:description', content: SITE_DESCRIPTION },
+            { name: 'twitter:image', content: '/preview.jpeg' },
```
valiafetisov, line 47: Meta-tags usually do not support relative paths (same for the other `og:image` tag)
https://stackoverflow.com/a/45941439

zoey-kaiser: This is not a relative path in this case. We have the image `preview.jpeg` in our static folder. When the frontend is started all the images and files in the static folder are served. Meaning that the image is available in this way (it also works on discord and slack). 

I assume that Twitter was just missing the `twitter:image` tag. Does this explanation make sense?

valiafetisov: I mean that twitter expects `twitter:image` or `og:image` tag to include domain name, eg `https://domain.com/path/to/the/image.jpg` at least that's how it was some time ago: https://stackoverflow.com/a/9858694

zoey-kaiser: Hmm, okay that makes sense. 

Can we use the environment variable `PRODUCTION_DOMAIN` as the base URL? Or is there an even easier way I can retrieve the current domain? I am not sure how this works due to our Kubernetes Cluster (if the WebApp runs in containers it might not know its own domain right?)

valiafetisov: `PRODUCTION_DOMAIN` is a domain without schema and is used for different purpose as stated in the readme. I guess you would need to add env variable something like `FRONTEND_ORIGIN` for this. Please don't forget to modify:
- dockerfile
- drone
- nuxt.config
- readme
- secrets (or provide commands if you don't have enough rights)

zoey-kaiser:
`secret write auction-ui/production/frontend FRONTEND_ORIGIN "https://unified-auctions.makerdao.com"`
`secret write auction-ui/staging/frontend FRONTEND_ORIGIN "https://main.auction-ui.k8s.sidestream.tech"`

Please double check that the name are correct! Localhost does not need this variable, as it will never be exposed to say Twitter!

valiafetisov:
> - dockerfile
> - drone
> - readme

Has not been changed so far

> `auction-ui/staging/frontend`

There is no such path. Correct paths can be found in the drone config for the frontend build pipelines

zoey-kaiser:
New commands:

`secret write auction-ui/production.auction-ui.k8s.sidestream.tech/frontend FRONTEND_ORIGIN "https://unified-auctions.makerdao.com"`
`secret write auction-ui/main.auction-ui.k8s.sidestream.tech/frontend FRONTEND_ORIGIN "https://main.auction-ui.k8s.sidestream.tech"`

valiafetisov:
> > * readme
> 
> Has not been changed so far

frontend/README.md
```diff
@@ -42,3 +42,4 @@ Help on both things is given in the linked resources above.
 - `STAGING_BANNER_URL`: (optional) When set a banner will be displayed, warning the user that they are using a staging version. The text will use `STAGING_BANNER_URL` as a link to production UI.
 - `MAX_PRIORITY_FEE_PER_GAS_WEI`: (optional, default can be found in core/src/gas.ts) – [EIP-1559](https://eips.ethereum.org/EIPS/eip-1559) `max_priority_fee_per_gas` value
 - `HEAPIO_ID`: (optional) [HeapIO analytics](https://heapanalytics.com/) Project's Environment ID. Required for tracking analytics.
+- `FRONTEND_ORIGIN`: (optional) Used to correctly register meta tags for social media sites
```
valiafetisov, line 45: You should say not (only) where it's used, but what is it
```suggestion
- `FRONTEND_ORIGIN`: (optional, default empty) – public url of the frontend, eg `https://unified-auctions.makerdao.com`. Used to correctly specify open graph meta tags
```

valiafetisov (approved):

PR #426, opened 2022-08-26
Simple simulations

### LINKED ISSUE #420: Improve hardhat setup
### Goal

Make running local testing environment more developer-friendly 

### Context

After merging https://github.com/sidestream-tech/unified-auctions-ui/pull/408, we would have at least 2 main scenarios for testing, but actually more (if counting surplus auctions on the previous contract). Therefore, we would like to have a setup where:

1. From one side it's very easy to run local testing environment via interactive cli. The developer should be able to navigate to the `hardhat` folder (or maybe we abandon it and use `core` folder for that) and execute single command which will:
    - check env vars
    - show list of available simulations (eg: `3 WSTETH-A auctions`, `Single debt auction`, etc)
2. From the other side make adding new simulations an easy task via extensible config (stored directly in our codebase) where developer can specify
    - Name of the simulation
    - Optional block number
    - Function that needs to be executed
    - etc

For inspiration, please have a look at the [`choose.ts`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/chaos-proxy/src/choose.ts) and the use of `prompts` in that folder.

### Tasks

- [ ] Propose config structure, refactoring scope
- [ ] Implement

KirillDogadin-std:
Closes #420 

Checklist:


KirillDogadin-std:
For manual testing:

```sh
> cd core
> npm run hardhat
# new terminal
> cd core
> npm run simulate
```

Then there're simulations and if you start the frontend up you could also test out the effects simulations take on the app

valiafetisov:
As suggested, can you please use something like [concurrently](https://www.npmjs.com/package/concurrently) to make `npm run simulate` a single command?

valiafetisov (approved):

LukSteib (changes requested):
Some early feedback: 
- [ ] can we find a more speaking title for `Reset blockchain fork` -> what is done here? forking the blockchain at a certain block? 
- [ ] Can we execute the first step of each simulation - currently: first selecting via return key and then again confirming via return key
- [ ] Can we include some kind of indication in console that the step is successfully executed
- [ ] If one has to reset the current fork in order to start the next simulation I would like to see an indication on what simulation is currently running

core/simulations/scripts/warpTime.ts
```diff
@@ -0,0 +1,5 @@
+import { warpTime } from '../../helpers/hardhat';
+
+export default async () => {
+    await warpTime('custom');
+};
```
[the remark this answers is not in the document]
KirillDogadin-std: the initial idea is that all the scripts here do not require the arguments. so this just wraps the function so that it provides the necessary args and the config is abstracted from them - it only has to know the name of the script and nothing more.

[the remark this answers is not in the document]
KirillDogadin-std: removed the older warptime function from the store , now we can add it in the simulation instead.

core/simulations/index.ts
```diff
@@ -0,0 +1,39 @@
...
+        const answer = await prompts(promtConfig);
+        const simulationConfig = SIMULATIONS.find(item => item.title === answer.value);
+        if (!simulationConfig) {
+            throw new Error(`Simulation config not found: ${answer.value}`);
+        }
+        for (const step of simulationConfig.steps) {
+            console.info(`Next step: ${step.title}`);
+            await keypress();
+            step.entry();
+        }
```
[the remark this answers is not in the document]
KirillDogadin-std: >  Can you please add logging to match suggested output?

this is not entirely clear

LukSteib (changes requested):
Would suggest the following re-naming of simulations: 
- [x] Fork while WSTETH-A is running -> Fork block with active WSTETH-A auction
- [x] Fork while system is in surplus -> Fork block with active surplus auction

Other than that:
- [x] What is the wallet that is used for the simulations (e.g. when adding DAI and MKR) -> guess it is the first hardhat address so maybe this can be shown as an output (incl private key) so that one can do the UI based checks in parallel with this wallet
- [ ] NIT: Can we replace the `-` indicator for the balances in the terminal output with a `:` or something else; otherwise one could be confused of it being negative balances
   - [x] the balance indicator in the terminal seems to be off. Indicating `mkr - 100000000000000000000` while actually `100` is added to wallet
   - [x] can the amount of `100` be increased -> e.g. for bidding on debt auction more is needed
- [ ] After starting the `WSTETH-A` simulation auction data is not fetched. See console error below (s1) 
   - [ ] note: browser hard refresh resolved the problem 




s1:
```
auctions.ts?b334:198 
fetch auction error Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="list()", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.4)
    at Interface.decodeFunctionResult (interface.js?8abb:345:1)
    at Contract.eval (index.js?f596:293:1)
    at Generator.next (<anonymous>)
    at fulfilled (index.js?f596:5:1)
```

KirillDogadin-std:
> After starting the WSTETH-A simulation auction data is not fetched. See console error below (s1)

failed to reproduce. Try again, see if something changes. if the bug still occurs, we can have a call to see what's going on there.

LukSteib (changes requested):
- [x] An additional simulation step that might be added -> warp time; this would allow to a) either test collecting functionality or b) test restart functionality (in case no bid was placed)

core/simulations/config.ts
```diff
@@ -0,0 +1,31 @@
+import causeDebt from './scripts/causeDebt';
+import resetFork from './scripts/resetFork';
+import warpTime from './scripts/warpTime';
+export const SIMULATIONS = [
+    {
+        title: 'create debt auction',
+        steps: [
+            {
+                title: 'Reset blockchain fork',
+                entry: resetFork,
```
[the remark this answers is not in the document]
valiafetisov: > But i can't get why the latter means that we branch off the latest block.

Cause the block number is not provided and as per suggestion `latest` should be the default one

core/hardhat.config.ts
```diff
@@ -5,11 +5,23 @@ import ethers from '@nomiclabs/hardhat-ethers';
...
+const REMOTE_RPC_URL = process.env.REMOTE_RPC_URL;
+
+if (!REMOTE_RPC_URL) {
+    throw new Error('REMOTE_RPC_URL env variable not set, please check your `hardhat/.env` file');
+}
 
 const customNetworkConfig = {
     testnetwork: {
         url: LOCAL_RPC_URL,
     },
```
valiafetisov, line 17: Why do you need to create this `testnetwork` and provide url to the hardhat itself within the hardhat config? (also, this env var is not declared in the dockerfile)

core/helpers/hardhat.ts
```diff
@@ -76,8 +81,14 @@ export const createWalletForRpc = async (signerPrivateKey: string, network: stri
     return hre.network.provider;
 };
 
+export const warpTime = async function (network: string, blocks = 20000, secondsBetweenBlocks = 270) {
+    const provider = await getProvider(network);
+    await provider.send('hardhat_mine', [`0x${blocks.toString(16)}`, `0x${secondsBetweenBlocks.toString(16)}`]);
```
valiafetisov, line 86: Don't you have `formatToHex` helper for this? Can you please use this helper everywhere, including `network.ts`, `rpc.ts`, `surplus-test.ts`?

core/helpers/hardhat.ts
```diff
@@ -87,3 +98,38 @@ export const resetBlockchainFork = async function (
     const provider = await createWalletForRpc(signerPrivateKey, network);
     return provider;
 };
+
+export const addToBalance = async (
+    network: string,
+    walletAddress: string,
+    mkrAmount: BigNumber,
+    daiAmount: BigNumber
+) => {
```
valiafetisov, line 107: I would do it as two different functions `overwriteDaiBalance` and `overwriteMkrBalance` there is no need to combine them together

core/helpers/hardhat.ts
```diff
@@ -87,3 +98,38 @@ export const resetBlockchainFork = async function (
...
+        provider,
+        'MCD_GOV',
+        '0x1',
+        walletAddress,
+        mkrAmount.shiftedBy(WAD_NUMBER_OF_DIGITS)
+    );
+
+    const daiBalanceHex = await daiContract.balanceOf(walletAddress);
+    const mkrBalanceHex = await mkrContract.balanceOf(walletAddress);
+    const daiBalance = new BigNumber(daiBalanceHex._hex).shiftedBy(-WAD_NUMBER_OF_DIGITS);
```
valiafetisov, line 131:
```suggestion
    const daiBalance = new BigNumber(daiBalanceHex._hex).shiftedBy(-DAI_NUMBER_OF_DIGITS);
```

core/helpers/hardhat.ts
```diff
@@ -87,3 +98,38 @@ export const resetBlockchainFork = async function (
...
+    const mkrBalance = new BigNumber(mkrBalanceHex._hex).shiftedBy(-WAD_NUMBER_OF_DIGITS);
```
valiafetisov, line 132:
```suggestion
    const mkrBalance = new BigNumber(mkrBalanceHex._hex).shiftedBy(-MKR_NUMBER_OF_DIGITS);
```

core/simulations/config.ts
```diff
@@ -0,0 +1,57 @@
+import causeDebt from './scripts/causeDebt';
+import resetFork from './scripts/resetFork';
+import warpTime from './scripts/warpTime';
+import increaseBalances from './scripts/addDaiAndMkrToWallet';
+
+export const SIMULATIONS = [
+    {
+        title: 'create debt auction',
```
valiafetisov, line 8:
```suggestion
        title: 'Create debt auction',
```

core/simulations/config.ts
```diff
@@ -0,0 +1,57 @@
...
+                entry: async () => await warpTime(),
+            },
+        ],
+    },
+    {
+        title: 'Fork block with active surplus auction',
+        steps: [
+            {
+                title: 'Reset blockchain fork',
+                // Few blocks before with various states of surplus auctions at 14078339,
```
valiafetisov, line 48: Few blocks before what?

core/simulations/index.ts
```diff
@@ -0,0 +1,46 @@
...
+        title: simulationName,
+        value: simulationName,
+    }));
+    const promtConfig: PromptObject = {
+        type: 'select',
+        name: 'value',
+        message: 'Select Simulation',
+        choices: promtChoices,
+    };
+    console.info(`Wallet Credentials: public key: ${HARDHAT_PUBLIC_KEY}, private key: ${HARDHAT_PRIVATE_KEY}`);
```
valiafetisov, line 26: I think this can be printed inside the function that resets the network and creates provider? It's kind of out of context here (and if you have provider, you don't have to hardcode public key)

core/simulations/scripts/causeDebt.ts
```diff
@@ -0,0 +1,11 @@
...
+import { causeDebt } from '../../helpers/causeDebt';
+import { HARDHAT_PRIVATE_KEY } from '../../helpers/constants';
+import { createWalletForRpc } from '../../helpers/hardhat';
+
+const provider = hre.network.provider;
+
+export default async () => {
+    await createWalletForRpc(HARDHAT_PRIVATE_KEY, 'custom');
+    await causeDebt('custom', provider);
+};
```
valiafetisov, line 11: Why do you need to call `createWalletForRpc` twice? Inside `createWalletForRpc` and then also here?

core/simulations/scripts/warpTime.ts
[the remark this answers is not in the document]
valiafetisov: > the initial idea is that all the scripts here do not require the arguments

I think the initial idea was to create functions already in a way what have unnecessary/default parameters set as defaults. But I don't see why do we want to create new function just to wrap another function with default parameters (which has exactly the same name)

In my opinion you can use `warpTime` function directly in the simulation config

core/simulations/index.ts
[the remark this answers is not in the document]
valiafetisov: The suggested output was specified in this comment: https://github.com/sidestream-tech/unified-auctions-ui/issues/420#issuecomment-1225615981

core/simulations/scripts/causeDebt.ts
KirillDogadin-std: i don't understand this question. i think there's a misnaming in your sentence. 

Could you link places where this is called?

core/helpers/hardhat.ts
KirillDogadin-std: `network.ts` does not have anything to use this on as i see.

core/hardhat.config.ts
KirillDogadin-std: essentialy this is related to the ci cd.

When adding it i did not discover how to point to the hardhat fork that is spawned by default. So instead i created a separate instance that has an adderess that we know. This fork is the `testnetwork`.

> also, this env var is not declared in the dockerfile)

[false](https://github.com/sidestream-tech/unified-auctions-ui/pull/426/files#diff-e95472413b240ae7729b7bea4f74662574ce035d769cf78f1191e47a277d79cbL21).

LukSteib (changes requested):
- [ ] Can you include the "Skip time" step to the "Create debt auction" simulation as well?

core/simulations/config.ts
```diff
@@ -0,0 +1,57 @@
...
+            {
+                title: 'Create debt auction',
+                entry: async () => await causeDebt(),
+            },
+            {
+                title: 'Add DAI and MKR to the wallet',
+                entry: async () => await increaseBalances(),
+            },
+            {
+                title: 'Expire the auction',
```
KirillDogadin-std, line 23:
```suggestion
                title: 'Skip time',
```

core/simulations/scripts/causeDebt.ts
valiafetisov: You first call it here:
https://github.com/sidestream-tech/unified-auctions-ui/blob/21ef5cd07a6cbc3b44a60063a51931ad80f457ee/core/simulations/config.ts#L12 inside `resetFork` "script"

Then you call it in the next step:
https://github.com/sidestream-tech/unified-auctions-ui/blob/21ef5cd07a6cbc3b44a60063a51931ad80f457ee/core/simulations/config.ts#L16 inside `causeDebt` "script"

core/simulations/scripts/causeDebt.ts
```diff
@@ -0,0 +1,11 @@
+import hre from 'hardhat';
+import { causeDebt } from '../../helpers/causeDebt';
+import { HARDHAT_PRIVATE_KEY } from '../../helpers/constants';
+import { createWalletForRpc } from '../../helpers/hardhat';
+
+const provider = hre.network.provider;
+
+export default async () => {
+    await createWalletForRpc(HARDHAT_PRIVATE_KEY, 'custom');
```
valiafetisov, line 9: - [ ] Why does `createWalletForRpc` hardhat helper function need to accept `HARDHAT_PRIVATE_KEY` every time it is called? Do we expect other wallets? Then the function wouldn't work anyway, because it doesn't return a wallet, but sets it globally. I think it makes sense to either hardcode `HARDHAT_PRIVATE_KEY` inside this function or at least make it an optional parameter
- [ ] Please always make `network` parameter to be the first argument like everywhere else (you can also make it default for all test helpers since we don't expect them to change)

core/simulations/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        choices: promtChoices,
+    };
+    while (true) {
+        const answer = await prompts(promtConfig);
+        const simulationConfig = SIMULATIONS.find(item => item.title === answer.value);
+        if (!simulationConfig) {
+            throw new Error(`Simulation config not found: ${answer.value}`);
+        }
+        for (const [i, step] of simulationConfig.steps.entries()) {
+            console.info(`Next step: ${step.title}`);
```
valiafetisov, line 32: The log still doesn't match [output proposed in this comment](https://github.com/sidestream-tech/unified-auctions-ui/issues/420#issuecomment-1225615981), right?

valiafetisov (commented):
- [x] Please let's remove `scripts` folder, as discussed here https://github.com/sidestream-tech/unified-auctions-ui/pull/426#discussion_r959307445 I find it confusing to create the same functions twice to wrap other functions with default parameters
  - [ ] Instead, please make sure env variables are not passed over, but set as default function arguments
- [ ] I suggest to split `core/simulations/config.ts` into `core/simulations/configs` folder with `*Simulation.ts`, one for each simulation, so we don't end up with a huge file in the end

core/hardhat.config.ts
KirillDogadin-std: https://hardhat.org/hardhat-network/docs/overview#how-does-it-work?

> It runs as either an in-process or stand-alone daemon, servicing JSON-RPC and WebSocket requests.

- currently it's a standalone because in this case we can configure the address it is at.
- when we do `npm run hardhat` the daemon starts and is located at localhost, so when developing we need this value to be `localhost:8545`
- when we are on ci pipeline this hardhat node is deployed as a separate service and thus we have to configure the env variable so that it points to the service in the drone. 
- The network is used for tests (see `package.json`) because otherwise the `hardhat test` will spawn its own instance of hardhat that apparently doesn't have an url to reach it through (from what i remember when implementing it some months ago)

TL;DR: we need this env var so that the tests know what address the node is at.

valiafetisov: > TL;DR: we need this env var so that the tests know what address the node is at.

Yes, exactly, I was not asking why do we need this variable. We need this variable in order to point tests toward service that runs RPC endpoint, that is correct. I were asking about the `testnetwork` config. 

> When adding it i did not discover how to point to the hardhat fork that is spawned by default. So instead i created a separate instance that has an adderess that we know. This fork is the `testnetwork`.

This was very confusing answer. After reading the documentation, I can finally understand what is happening:

- When we run `npx hardhat node`, the `hardhat` portion of the config is used and this creates a new RPC endpoint (in other words a new node is created for others to run against)
- When we run `npx hardhat --network testnetwork run ./simulations/index.ts`, the `testnetwork` portion of the config is created, but it doesn't create new RPC endpoint, it connects to the provided one and executes provided script over it

Since those two things will be run in two different processes, this now make more sense, but doesn't make it less confusing. Not sure if it was a good decision to merge those two folders together. Maybe it will make sense to at least store configs run by separate processes separately (eg, currently you [validate REMOTE_RPC_URL](https://github.com/sidestream-tech/unified-auctions-ui/blob/e683d4a085633e81414831523c2b89b210f2264a/core/hardhat.config.ts#L10-L12) for both processes, while you need it only in one)? `npx hardhat` have `--config` option to achieve separation into `remote.hardhat.config.ts` and `local.hardhat.config.ts`

core/simulations/configs/surplusAuctionSimulation.ts
```diff
@@ -0,0 +1,21 @@
...
+export default {
+    title: 'Fork block with active surplus auction',
+    steps: [
+        {
+            title: 'Reset blockchain fork',
+            // Block with various states of surplus auctions at 14078339,
+            entry: async () => {
+                console.info(
+                    `Wallet Credentials: public key: ${HARDHAT_PUBLIC_KEY}, private key: ${HARDHAT_PRIVATE_KEY}`
+                );
```
valiafetisov, line 12: As suggested, can we move this to the `resetBlockchainFork` and print it using `signerPrivateKey` variable and `provider.getAddress()`? (And don't forget to remove this logging from all other simulation configs)

In general, you should print information into the console where it is actually used, otherwise in case of the code change printed information will be invalid and misleading

core/src/rpc.ts
```diff
@@ -23,8 +25,8 @@ const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undef
 };
 
 export const setupRpcUrlAndGetNetworks = async function (
-    rpcUrl?: string,
-    isDev?: boolean
+    rpcUrl: string = LOCAL_RPC_URL,
```
valiafetisov, line 28: This function is not only used for test, so it's misleading to hardcode `LOCAL_RPC_URL` here. Core functionality is used without env variable `LOCAL_RPC_URL` whatsoever, so this env variable shouldn't appear in the regular (non-test-related) core functions

core/simulations/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        const answer = await prompts(promtConfig);
+        const simulationConfig = SIMULATIONS.find(item => item.title === answer.value);
+        if (!simulationConfig) {
+            throw new Error(`Simulation config not found: ${answer.value}`);
+        }
+        for (const [i, step] of simulationConfig.steps.entries()) {
+            console.info(`Next step is "${step.title}"`);
+            if (i !== 0) {
+                await keypress();
+            }
```
valiafetisov, line 35:
```suggestion
            if (i !== 0) {
                await keypress(step.title);
            }
```

Then please also update `keypress` to `Press enter to proceed to the next step "${stepTitle}"` to be more concise and print less lines as suggested in the https://github.com/sidestream-tech/unified-auctions-ui/issues/420#issuecomment-1225615981

core/simulations/index.ts
```diff
@@ -0,0 +1,44 @@
...
+        title: simulationName,
+        value: simulationName,
+    }));
+    const promtConfig: PromptObject = {
+        type: 'select',
+        name: 'value',
+        message: 'Select Simulation',
+        choices: promtChoices,
+    };
+    while (true) {
```
valiafetisov, line 25: Maybe the loop should start on the top of the file, so that when one simulation is completed, the other can be selected?

valiafetisov:
- [ ] See also https://github.com/sidestream-tech/unified-auctions-ui/pull/426#discussion_r962631981

core/simulations/configs/surplusAuctionSimulation.ts
KirillDogadin-std: `provider` is `hre.network.provider` of `EtheriumProvider` type, it's not a signer. To print public key in this function as you suggest one would have to extract the signer first. See the change.

core/simulations/index.ts
KirillDogadin-std: i don't understand nor the suggestion, nor the reasoning here:

  - "top of the file" start of the loop does not make sense to me: what about imports and helper function definitions?
    - if "top of the function" is actually meant, then effectively this is already the case because higher we only have the variable definitions that are the same for every loop
  - the other simulation can already be selected after the previous one is concluded.

valiafetisov: Alright, I got confused by the generic `const answer = await prompts(promtConfig)` which doesn't really say what it does. This makes code more readable:

```ts
const selectedSimation = await prompts(selectSimulationPrompt);
const simulationConfig = SIMULATIONS.find(simulation => simulation.title === selectedSimation.value);
```

core/hardhat.config.ts
KirillDogadin-std: i don't think that "removing" the default option (`hardhat.config.ts`) is an improvement. How about assuming that the remote option is the default and the local has to use different config? (corresponding change applied.)

core/simulations/index.ts
```diff
@@ -0,0 +1,43 @@
+import prompts, { PromptObject } from 'prompts';
+import { SIMULATIONS } from './config';
+import readline from 'readline';
+
+const keypress = async (title: string) => {
+    await prompts({
+        type: 'invisible',
+        name: 'irrelevant',
+        message: `Press Enter to continue to step ${title}`,
```
valiafetisov, line 9:
```suggestion
        message: `Press Enter to proceed to the next step "${title}"`,
```

core/simulations/configs/createDebtAuctionSimulation.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export default {
+    title: 'Create debt auction',
+    steps: [
+        {
+            title: 'Reset blockchain fork',
+            entry: async () => {
+                console.info(
+                    `Wallet Credentials: public key: ${HARDHAT_PUBLIC_KEY}, private key: ${HARDHAT_PRIVATE_KEY}`
+                );
```
valiafetisov, line 17: Please remove duplicative logs from all configs

core/simulations/index.ts
```diff
@@ -0,0 +1,43 @@
...
+    });
+};
+
+const selectAndRunSimulation = async () => {
+    readline.emitKeypressEvents(process.stdin);
+    const promtChoices = SIMULATIONS.map(simulationConfig => simulationConfig.title).map(simulationName => ({
+        title: simulationName,
+        value: simulationName,
+    }));
+    const promtConfig: PromptObject = {
```
valiafetisov, line 19: According to the https://github.com/sidestream-tech/unified-auctions-ui/pull/426#discussion_r962707075
```suggestion
    const selectSimulationPrompt: PromptObject = {
```

core/hardhat.config.ts
valiafetisov: > i don't think

Can you elaborate why? My suggestion was about making it less confusing through consistency of the terms `remote`/`local`, which your suggestion contradicts without explaining the reasoning

valiafetisov:
- [ ] See also https://github.com/sidestream-tech/unified-auctions-ui/pull/426/files/dffd5f658c3eea3295acf1ef59f9db5f981f65d8#r962788703

core/hardhat.config.ts
KirillDogadin-std: the reasoning is that the absence of the default config is confusing as well - i would expect a config that provides a valid fork / node to be the default option and the other configs to be the ones that cover specific cases.

on top of that, we would have to write out the config file arg for every single command `npx hardhat` - that is inconvenient.  It's possible to save some keystrokes there.

Also, if i think about naming then distinction between `local` and `remote` is not exactly clear as well:

- the `local` as we call it now is just there to point to the local node via `url` or in other words it shows "which node to use"
- the `remote` is there to configure on how we should create fork (`forking/url`) and does not provide the `url` parameter.

So if we stick to double-naming then i would say it should be `local.hardhat.config` and `fork.hardhat.config`

core/package.json
```diff
@@ -7,9 +7,12 @@
     "dev": "tsc-watch --noClear",
     "build": "tsc --build tsconfig.json",
     "start": "node .",
-    "test": "npx hardhat test --network testnetwork",
+    "test": "npx hardhat test --network testnetwork --config local.hardhat.config.ts",
     "lint": "eslint --ext \".js,.ts\" --ignore-path .gitignore . --max-warnings=0",
-    "fake-flop": "npx hardhat --network testnetwork run ./scripts/causeDebt.ts"
+    "hardhat:silent": "npx hardhat node 1>&/dev/null",
+    "hardhat": "npx hardhat node",
+    "scripts": "npx hardhat --network testnetwork run ./simulations/index.ts",
```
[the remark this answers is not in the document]
KirillDogadin-std:
```
Error HH8: There's one or more errors in your config file:

  * HardhatConfig.networks.hardhat can't have an url

File: core/package.json [hunk 24]
```diff
@@ -7,9 +7,12 @@
     "dev": "tsc-watch --noClear",
     "build": "tsc --build tsconfig.json",
     "start": "node .",
-    "test": "npx hardhat test --network testnetwork",
+    "test": "npx hardhat test --network testnetwork --config local.hardhat.config.ts",
```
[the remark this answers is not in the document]
KirillDogadin-std: that seems to be not allowed by hardhat ( see the error log in the other [suggestion](https://github.com/sidestream-tech/unified-auctions-ui/pull/426#discussion_r962917566) )

core/package.json
```diff
@@ -11,7 +11,7 @@
     "lint": "eslint --ext \".js,.ts\" --ignore-path .gitignore . --max-warnings=0",
     "hardhat:silent": "npx hardhat node 1>&/dev/null",
     "hardhat": "npx hardhat node",
-    "scripts": "npx hardhat --network testnetwork run ./simulations/index.ts",
+    "hardhat:simulations": "npx hardhat --network testnetwork --config local.hardhat.config.ts run ./simulations/index.ts",
     "simulate": "npm-run-all --parallel hardhat:silent scripts"
```
valiafetisov, line 15:
```suggestion
    "simulate": "npm-run-all --parallel hardhat:silent hardhat:simulations"
```

PR #444, opened 2022-08-31
Add `AnimatedArrow` component

### LINKED ISSUE #441: Create `AnimatedArrow` component
### Goal
In our new Vaults flow, we require an arrow, from which we can change the direction. Similarly to the `AnimatedNumber` component, we want to show a small animation when a user is viewing a page and the arrow changes direction. Therefore we want to have a arrow component, that shows an arrow that receives the property "direction". This can be "up", "right", "down" or "left". This value can change during the lifecycle of the arrow and when it does, the arrow should spin into the correct alignment. Please also take different sizing options into account.

### Context
In order to properly add the arrow we will first need to find a suitable arrow icon. You can find a free SVG icon (with a commercial license) that we are allowed to use. Once deciding on one design please use a [SVG Cleaner](https://jakearchibald.github.io/svgomg/) to remove any unneeded meta tags. Then you can import the arrow into the project.

### Tasks

- [x] Decide on an Arrow Icon to use inside of the project
- [x] Implement the new `AnimatedArrow` component
- [x] Add stories for all directions and an example in which the arrows direction is changed periodically

aomafarag:
Closes #441.

Checklist:

### Preview

screencast

Link to SVG: https://www.svgrepo.com/svg/19400/up-arrow


zoey-kaiser:
Some screenshots would be amazing!

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -0,0 +1,44 @@
...
+        UpArrow,
+    },
+    props: {
+        size: {
+            type: Number,
+            default: 16,
+        },
+        direction: {
+            type: String,
+            default: 'up',
```
zoey-kaiser, line 20: Can we type this with the directions and not just a string?

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -0,0 +1,44 @@
...
+<script lang="ts">
+import Vue from 'vue';
+import UpArrow from '~/assets/icons/uparrow.svg';
+
+export default Vue.extend({
+    components: {
+        UpArrow,
+    },
+    props: {
+        size: {
```
zoey-kaiser, line 14: I would maybe make the `size` prop a string and allow the dev to define what level of measurement to use. This would allow us to use px, rem or any other kind of unit. 

This change will need to be taken into account in the computed style data as well!

zoey-kaiser (changes requested):

zoey-kaiser (approved):

frontend/components/common/other/AnimatedArrow.stories.js
```diff
@@ -0,0 +1,50 @@
...
+        ...common,
+        template: '<AnimatedArrow direction="left" />',
+    }))
+    .add('Right', () => ({
+        ...common,
+        template: '<AnimatedArrow direction="right" />',
+    }))
+    .add('Big', () => ({
+        ...common,
+        template: '<AnimatedArrow size="64px" />',
```
valiafetisov, line 45: Normally, we would control size by setting a class or setting a style tag directly on the component. Why do you need to make a prop for that?

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -0,0 +1,46 @@
...
+            type: String,
+            default: '16px',
+        },
+        direction: {
+            type: String as Vue.PropType<ArrowDirections>,
+            default: 'up',
```
valiafetisov, line 22: Can we instead make it required? (I don't see where the component can be used without providing this prop)

valiafetisov (approved):

Issue #419, opened 2022-08-16
Create mock for the vault liquidation feature


zoey-kaiser:
### Goal
Mocked vault liquidation feature

### Context
When a vault becomes underwater, it can be liquidated and a collateral auction can be started. Beforehand the vault can be at risk for a certain amount of time. With this addition we hope to give users insight into what vaults are at risk and help them liquidate underwater vaults. 

### Tasks

- [x] Create mock
- [x] Discuss
- [ ] Create outline of what needs to be done

zoey-kaiser:
See mock: https://whimsical.com/makerdao-vaults-SgfmKedbUSoRS4FFvbnvTG

<details><summary>MOCK v1</summary>
<p>

![MakerDAO vaults](https://user-images.githubusercontent.com/30908158/184897385-ee39d59a-bd42-4662-b174-cf1f27fbec92.png)

</p>
</details>

valiafetisov:
- [ ] Liquidated vault becomes a collateral auction (if it wasn't clear so far), so instead of `Vault not found` there should be a _seamless_ transition/possibility to switch to the started collateral auction. This is one of the main UI problems how to combine those two, so I would start from answering this question
  - [ ] Moreover, we need to investigate if data about liquidated vault can be fetched after it was liquidated and adjust `not found` page accordingly
- [ ] `No wallet was selected to send the incentive` shouldn't be a panel from my perspective, because it's _optional_ non-primary action. The idea of the panels is to provide a clear path though the execution which is already quite complicated. We already have functionality for bidding where the user can receive outcome to another wallet – so you can use the same pattern here
- [ ] There is no transaction page? The user have to understand why it have to spend money _before clicking the button_ and even before connecting a wallet and engaging into panels. Currently there is just a bunch of numbers which doesn't make sense to someone seeing this page for the first time
- [ ] For advanced users there should be a way to prepare for new liquidatable vaults (ie pre-authorize, etc)

Therefore, the only important information for a vault is:
- How close it is to being "liquidatable"?
- How much would I get if I liquidate it?
  - Is it `Liquidation price`? Then maybe at least `Liquidation prize` is a better name? `Net profit`?
- When will those numbers change?
- In which direction will those numbers change if the change would happen right now?

The rest are technical details of the vault that doesn't make a difference to me, at least according to the [investigation](https://github.com/sidestream-tech/unified-auctions-ui/issues/363). `Risk score` from my perspective doesn't communicate enough information

[...]

PR #446, opened 2022-09-05
Vault and liquidation-related types

### LINKED ISSUE #447: Define types for liquidation of auctions
### Goal

Types are defined for liquidation of auctions functionality.

### Context

During the discussion in #363 it was agreed to define the types that are going to be used in the liquidation functionality.

### Tasks

- [x] Propose type defenitions that cover the information that is going to be used in the liquidations tool.
- [x] Discuss & adjust.

KirillDogadin-std:
Closes #447

Checklist:

### DIFF
--- a/core/src/types.ts
+++ b/core/src/types.ts
@@ -269,14 +269,11 @@ export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, Com
 
 type CollateralType = CollateralConfig['title'];
 
-export declare interface CollateralLiquidationLimit {
-    currentCollateralDebtDai: BigNumber;
-    maximumCollateralDebtDai: BigNumber;
-}
 export declare interface LiquidationLimits {
     maximumProtocolDebtDai: BigNumber;
     currentProtocolDebtDai: BigNumber;
-    collateralDebts: Record<CollateralType, CollateralLiquidationLimit>;
+    currentCollateralDebtDai: BigNumber;
+    maximumCollateralDebtDai: BigNumber;
 }
 
 export declare interface VaultCollateralParameters {
@@ -302,14 +299,14 @@ export declare interface VaultTransactionFees {
     transactionFeeLiquidationDai: BigNumber;
 }
 
-export declare interface VaultBaseInfo extends VaultBase, VaultAmount {}
+export declare interface Vault extends VaultBase, VaultAmount, LiquidationLimits {}
 export declare interface OraclePrices {
     currentUnitPrice: BigNumber;
     nextUnitPrice: BigNumber;
     nextPriceChange: Date;
 }
 
-export declare interface VaultTransaction extends VaultBaseInfo, VaultTransactionFees, OraclePrices {
+export declare interface VaultTransaction extends Vault, VaultTransactionFees, OraclePrices {
     liquidationRatio: number;
     collateralizationRatio: number;
     proximityToLiquidation: number;


core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
 }
 
 export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, CompensationAuctionTransactionFees {}
+
+type CollateralType = CollateralConfig['title'];
+export declare interface Collateral {
+    type: CollateralType;
```
valiafetisov, line 272: Please always use `collateralType` key for collateral type for consistency. I also don't understand why `Vault` need to have two keys supposedly containing the same thing? `type` and `collateralType`

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+}
+
+export declare interface LiquidationLimit {
+    maximumAmountAuctioned: BigNumber;
+    currentlyAuctioned: BigNumber;
+}
+export declare interface OraclePrices {
+    current: BigNumber;
+    next: BigNumber;
+}
```
valiafetisov, line 305: This seems to be unused by other types. Is it intended?

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    incentiveRelative?: BigNumber;
+    incentiveCombined?: BigNumber;
+    grossProfit?: BigNumber;
+    transactionFees?: BigNumber;
+}
+
+export declare interface LiquidationLimit {
+    maximumAmountAuctioned: BigNumber;
+    currentlyAuctioned: BigNumber;
+}
```
valiafetisov, line 301: Please move it above the line where you're using it

valiafetisov (commented):
- [x] Please extend instead of repeating the types
- [x] Please choose better namings for the `*Contract` types, since provided values are not the contract
- [ ] Please provide a single end type of the vault that includes everything that needs to be displayed according to the mock (including)
  - [x] Please make sure it includes everything that is displayed in the mock (eg `proximityToTheLiquidation`, `nextPriceChange`, etc)
  - [x] Please make sure it includes transaction fees, combines transaction fees, etc
  - [x] It might make sense to add `state` eg `loaded`, `liquidated`, etc
  - [x] Collateralisation ratio and liquidation ration [shown in the mock](https://github.com/sidestream-tech/unified-auctions-ui/issues/419) seems to be missing
- [ ] Please provide a single end "context" type per collateral (or merge everything into the vault type above) eg (`LiquidationLimit` is per collateral it can be delivered per collateral or can be delivered inside the valut type, depends on the implementation

KirillDogadin-std:
> Please make sure it includes transaction fees, combines transaction fees, etc

for current mock in #419 it only makes sense to have only one transaction fee which is liquidation of the vault. All the others that are responsible for the updating the prices and rates do not have the corresponding ui elements and if i recall correctly we did not agree on automatic execution of those functions. So i would say that there is only one transaction fee to mention.

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    maxDebtPerCollateralUnit: BigNumber;
+    liquidationLimits: LiquidationLimit;
+}
+export declare interface VaultAddress {
+    id: number;
+    address: string;
+    collateralType: CollateralType;
+}
+export declare interface VaultDebtAmount {
+    initialDebt: BigNumber;
```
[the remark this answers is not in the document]
KirillDogadin-std: debt = generated dai (initial debt) + stability fees.

Since this type represents the information that is initially fetched from the blockchain, `debt` should not be put here because contract does not store this information. Instead it stores the stability fees' rate and the initially extracted dai. These two can be used to compute the overall debt.

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    stabilityFeeRate: BigNumber;
+    maxDebtPerCollateralUnit: BigNumber;
+    liquidationLimits: LiquidationLimit;
+}
+export declare interface VaultAddress {
+    id: number;
+    address: string;
+    collateralType: CollateralType;
+}
+export declare interface VaultDebtAmount {
```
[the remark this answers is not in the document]
KirillDogadin-std: This type represents the information that can be extracted from the vat vault [mapping](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/vat.sol#L57). There, we need the collateral type to retrieve the information, hence the type of collateral and the amounts of collateral&debt are grouped together.

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    initialDebt: BigNumber;
+    collateralAmount: BigNumber;
+    collateralType: CollateralType;
+}
+export declare interface VaultBaseInfo extends VaultAddress, VaultDebtAmount {}
+export declare interface Vault extends VaultBaseInfo {
+    proximityToLiquidation: number;
+    nextPriceChange: Date;
+    state: 'loaded' | 'liquidated' | 'not-liquidatable';
```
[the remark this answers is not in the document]
KirillDogadin-std: if the liquidation limits are reached, we cannot liquidate the vault. 

Perhaps it's better to toggle the disabled state of inputs based on the computed props and the global liquidation limits. So removed it.

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    incentiveRelative?: BigNumber;
+    incentiveCombined?: BigNumber;
+    grossProfit?: BigNumber;
```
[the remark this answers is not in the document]
KirillDogadin-std: made them required.

core/src/types.ts
```diff
@@ -266,3 +266,40 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
 }
 
 export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, CompensationAuctionTransactionFees {}
+
+type CollateralType = CollateralConfig['title'];
+export declare interface Collateral {
+    type: CollateralType;
+    stabilityFeeRate: BigNumber;
+    maxDebtPerCollateralUnit: BigNumber;
+    liquidationLimits: LiquidationLimit;
```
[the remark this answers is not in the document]
KirillDogadin-std: > There are two liquidation limits that needs to be fetched: global and collateral-based

i was thinking that the global liquidation limit should reside somewhere in store and thus should not be included in the every single vault object.

core/src/types.ts
KirillDogadin-std: extracted collateral type into the separate interface.

core/src/types.ts
KirillDogadin-std: initially yes, as part of other comments' adjustments it's now used in the global `Vault` type

valiafetisov:
> So i would say that there is only one transaction fee to mention.

Fine with me, if there is only one transaction and it is extracted into a separate type for ETH and DAI amounts 👍 

> just to clarify since i dont think i understood this suggestion correctly: all the info fields come together in a single type so that the ui functionality is able to fetch it all from one single object. Is that correct interpretation?

The frontend doesn't need to know about any underlying blockchain logic, which should be incapsulated in the `core`. The frontend should only call `getVaults`/`getVault` to receive all required information to display relevant data, decide on participation and participate via executing `liquidateVault` (since we don't have authorization, allowance or wallet balances required to participate that usually comes via separate requests). This logic should be reflected in the types, and implementation-wise this should be taken care via caching.

core/src/types.ts
```diff
@@ -266,3 +266,51 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+
+export declare interface VaultTransactionFees {
+    transactionFeeLiquidationEth: BigNumber;
+    transactionFeeLiquidationDai: BigNumber;
+}
+
+export declare interface VaultBaseInfo extends VaultBase, VaultAmount {}
+export declare interface OraclePrices {
+    current: BigNumber;
+    next: BigNumber;
```
valiafetisov, line 304: If `OraclePrices` are used to expend into `Vault`, those keys should have better names
```suggestion
    currentUnitPrice: BigNumber;
    nextUnitPrice: BigNumber;
```

core/src/types.ts
```diff
@@ -266,3 +266,51 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    nextPriceChange: Date;
+}
+
+export declare interface Vault extends VaultBaseInfo, VaultTransactionFees, OraclePrices {
+    liquidationRatio: number;
+    collateralizationRatio: number;
+    proximityToLiquidation: number;
+    state: 'loaded' | 'liquidated';
+    incentiveRelativeDai: BigNumber;
+    incentiveCombinedDai: BigNumber;
```
valiafetisov, line 314:
```suggestion
    incentiveRelativeDai: BigNumber;
    incentiveConstantDai: BigNumber;
    incentiveCombinedDai: BigNumber;
```

core/src/types.ts
```diff
@@ -266,3 +266,51 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    grossProfitDai: BigNumber;
```
valiafetisov, line 315: Net profit as well?

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: > These two can be used to compute the overall debt.

Should we then add both `initialDebtDai` and `debtDai`?

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: > we need the collateral type to retrieve the information

It needs to retrieve the information meaning that input already contains `collateralType`, therefore the function don't need to output the same thing as given in the input, right?

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: > Perhaps it's better to toggle the disabled state of inputs based on the computed props and the global liquidation limits. So removed it.

It's actually better when this logic (liquidatable/not liquidatable) used once (eg inside the `core`) and then frontend doesn't need to reimplement it. Based on this, I would suggest to have 3 states: `liquidatable`, `not-liquidatable`, `liquidated`

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: I think it complicates the frontend logic, without a reason: in order to determine `liquidatable` or `not-liquidatable` (which you would otherwise need to do anyway in core) you would need to know those values. I suggest to include both global and collateral limits to the vault object itself. They can be cached to avoid fetching them for every vault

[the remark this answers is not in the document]
KirillDogadin-std: When we liquidate one vault , the cache instantly becomes invalid and we can encounter the "limit reached" state: The cached return will claim that we can liquidate a next vault while in real world it's forbidden. We have to keep the limits updated at all times it seems to me. So even if we keep this in core, we still would have to keep the global limit in its separate object to avoid repetition and multiple refetches. On top of that - duplication of information for every vault object seems suboptimal to me.

core/src/types.ts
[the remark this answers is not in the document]
KirillDogadin-std: well, we can, but not in this type because then it would require the rates fetched beforehand. Added this to the `Vault`

core/src/types.ts
```diff
@@ -266,3 +266,51 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+
+export declare interface VaultBase extends VaultCollateralType {
+    id: number;
+    address: string;
+}
+
+export declare interface VaultAmount {
+    initialDebtDai: BigNumber;
+    collateralAmount: BigNumber;
+    collateralType: CollateralType;
```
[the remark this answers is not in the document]
KirillDogadin-std: addressed as part of the comment posted before.

core/src/types.ts
```diff
@@ -266,3 +266,51 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+export declare interface VaultCollateralType {
+    collateralType: CollateralType;
+}
+
+export declare interface LiquidationLimit {
+    maximumAmountAuctionedCollateral: BigNumber;
+    currentlyAuctionedCollateral: BigNumber;
+}
+
+export declare interface VaultCollateralParameters extends LiquidationLimit, VaultCollateralType {
```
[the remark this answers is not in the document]
KirillDogadin-std: addressed as part of the comment posted before.

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: > duplication of information for every vault object seems suboptimal to me.

You don't need to duplicate it (eg store it every time), but enrich vault with those parameters. 

But it's up to implementation, we can also provide limits separately from the vault. But then all limits, as I don't see the reason why collateral-related limits are "duplicated" in every vault, while global are not – both collateral and global limits will be updated as soon as the vault is liquidated. Making two different update mechanisms for this is more than suboptimal. Removing limits cache on `liquidateVault` call and refetching `getVaults` should be sufficient. Implementing complicated refetching limits logic twice (on frontend and in the bot) is sub-optimal in my opinion

core/src/types.ts
```diff
@@ -266,3 +266,53 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+
+type CollateralType = CollateralConfig['title'];
+export declare interface VaultCollateralType {
+    collateralType: CollateralType;
+}
+
+export declare interface LiquidationLimit {
+    maximumAmountAuctionedCollateral: BigNumber;
+    currentlyAuctionedCollateral: BigNumber;
+}
```
valiafetisov, line 278: - [ ] Now you're missing the type for global (non-collateral) limits
- [ ] Naming can be simpler and include currency
```suggestion
export declare interface LiquidationLimits {
    maximumProtocolDebtDai: BigNumber;
    maximumCollateralDebtDai: BigNumber;
    currentProtocolDebtDai: BigNumber;
    currentCollateralDebtDai: BigNumber;
}
```

core/src/types.ts
```diff
@@ -266,3 +266,53 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+
+export declare interface VaultCollateralParameters extends LiquidationLimit {
+    stabilityFeeRate: BigNumber;
+    maxDebtPerCollateralUnit: BigNumber;
```
valiafetisov, line 282: Is't `maxDebtPerCollateralUnit` also `*Rate` (not `*Unit`) or even `*Ratio` as in other keys below?

core/src/types.ts
```diff
@@ -266,3 +266,53 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    liquidationRatio: number;
+    collateralizationRatio: number;
+    proximityToLiquidation: number;
+    state: 'liquidated' | 'liquidatable' | 'not-liquidateable';
+    incentiveRelativeDai: BigNumber;
+    incentiveConstantDai: BigNumber;
+    incentiveCombinedDai: BigNumber;
+    grossProfitDai: BigNumber;
+    netProfitDai: BigNumber;
+    debtDai: BigNumber;
```
valiafetisov, line 317: The type that assembles everything together and mostly computes (profits, fees, etc) historically called `*Transaction` in our codebase. You can also adhere to that and create a separate type `VaultTransaction`.

To help you will separation: `Vault` type will be used by twitter bot to announce that there is a vault in the `liquidatable` state, `VaultTransaction` type will be used by the keeper/frontend to participate in it

core/src/types.ts
KirillDogadin-std: No, i am pretty sure that it's the `perUnit` metric since the condition is `stabilityFeeRate * initialDebtDai > collateralAmount * maxDebtPerCollateralUnit(Rate)` https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/dog.sol#L181

and the documentation states / comment states that it's a "price" 
https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/vat.sol#L47

core/src/types.ts
```diff
@@ -266,3 +266,53 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+
+export declare interface VaultCollateralParameters extends LiquidationLimits {
+    stabilityFeeRate: BigNumber;
+    maxDebtPerCollateralUnit: BigNumber;
+}
+
+export declare interface VaultBase {
+    id: number;
+    address: string;
+    collateralType: CollateralType;
```
valiafetisov, line 287: I would also add `lastSyncedAt: Date` and `network: string` keys somewhere in those vault types

core/src/types.ts
valiafetisov: So can `maxDebtPerCollateralUnit` be called `minUnitPrice`? because `debt`( `DAI`) per collateral (eg `ETH`) is basically unit price which is the minimum, lower which the vault will become `liquidatable`?

core/src/types.ts
```diff
@@ -266,3 +266,53 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
+    initialDebtDai: BigNumber;
+    collateralAmount: BigNumber;
+}
+
+export declare interface VaultTransactionFees {
+    transactionFeeLiquidationEth: BigNumber;
+    transactionFeeLiquidationDai: BigNumber;
+}
+
+export declare interface VaultBaseInfo extends VaultBase, VaultAmount {}
```
valiafetisov, line 300: Let's call this `Vault` then?
```suggestion
export declare interface Vault extends VaultBase, VaultAmount {}
```
Otherwise [the suggestion](https://github.com/sidestream-tech/unified-auctions-ui/pull/446/files#r963578099) is not fully implemented

valiafetisov:
- [ ] Please answer [this comment](https://github.com/sidestream-tech/unified-auctions-ui/pull/446#discussion_r963561377) as well

core/src/types.ts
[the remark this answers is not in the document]
KirillDogadin-std: Ok, then i think the best option is to store them separately per collateral and for global. See the types.

core/src/types.ts
```diff
@@ -269,22 +269,27 @@ export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, Com
...
+    currentCollateralDebtDai: BigNumber;
+    maximumCollateralDebtDai: BigNumber;
+}
 export declare interface LiquidationLimits {
     maximumProtocolDebtDai: BigNumber;
-    maximumCollateralDebtDai: BigNumber;
     currentProtocolDebtDai: BigNumber;
-    currentCollateralDebtDai: BigNumber;
+    collateralDebts: Record<CollateralType, CollateralLiquidationLimit>
 }
```
valiafetisov, line 280: As suggested, I don't think this is a good idea to complicate this object and cache it ourselves. This can be simplified via using `memoizee` over `getCollateralLiquidationLimits(network, collateralType)` if `LiquidationLimits` will become part of the `VaultTransaction`

core/src/types.ts
```diff
@@ -269,22 +269,27 @@ export declare interface DebtAuctionTransaction extends DebtAuctionEnriched, Com
...
 
-export declare interface VaultCollateralParameters extends LiquidationLimits {
+export declare interface VaultCollateralParameters {
     stabilityFeeRate: BigNumber;
-    maxDebtPerCollateralUnit: BigNumber;
+    minUintPrice: BigNumber;
 }
 
-export declare interface VaultBase {
+export declare interface Vault {
```
valiafetisov, line 287: This type doesn't contain enough to announce it being `liquidatable`. Therefore I will still call this type `VaultBase`

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: How does it solves `Implementing complicated refetching limits logic twice`? Why do we need to fetch limits separately and compare them with vault amounts manually twice in the bot and the frontend? Or how do you imagine it?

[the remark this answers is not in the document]
KirillDogadin-std: > Why do we need to fetch limits separately

if the q is actually "why do we need to fetch limits into separate object", then the answer is - we then have a central place and a single source of truth where the limits can be looked up for every vault (collateral type).

If the q is indeed "Why do we need to fetch limits separately", then the answer is - because the contract stores them in separate places and we have to fetch one limit object per collateral + limit object for global scope.

___

we don't need to compare them in frontend or bot, since we have the states `liquidatable` and `not-liquidatable` - we are forced to adjust the state along with the refreshed limits when the vault gets liquidated. This happens in the core. I don't see where it happens twice here.

So fetching vaults triggers the mechanism, core logic sets every vault to have the corresponding state; Then after liquidating each vault we have to update the limits and state because otherwise the data will be invalid. For frontend and bot only `Vaults` object will be relevant while `Limits` will serve to determine the states under the hood and will not be exposed to frontend/bot.

core/src/types.ts
KirillDogadin-std: answered in the q linked.

core/src/types.ts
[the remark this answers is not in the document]
valiafetisov: > we are forced to adjust the state along with the refreshed limits when the vault gets liquidated. This happens in the core. I don't see where it happens twice here

- Invalidation of the cache will need to be implemented twice on frontend and bot, since you expect them to refetch values after the liquidation
- `Limits` will need to be fetched separately from the `vaults` array twice
  - As they will still need to be provided to the frontend to be displayed as separate numbers (see the latest mock) and will be required for the logs of the bot
- They will need to be transformed twice from `LiquidationLimits.collateralDebts` to another format which doesn't send a record of all collateral limits

I still don't understand why do you need to first _build up a cache by hand_, then use it to enrich the vault with the state (and still provide those two interdependent things separately), when you can just call the function and _cache the results using library_ which will do all the heavy lifting for you?

I still advice to flatten `LiquidationLimits` and merge them into `Vault`

core/src/types.ts
valiafetisov: Meaning [the suggestion](https://github.com/sidestream-tech/unified-auctions-ui/pull/446/files#r963578099) is not fully implemented yet as there is no `Vault` type yet

valiafetisov (approved):

PR #449, opened 2022-09-07
Add new surplus auction simulation

### LINKED ISSUE #448: Add surplus auction simulation
### Goal

Have a simulation that initiates the surplus auction on the forked blockchain hardhat node.

### Context

Currently we're able to set the block number on the fork to the one that has surplus auctions active. However this has a downside because the contracts can be updated to the newer versions.
On top of that, the auctioned amount of dai in surplus auction is not controlled. Customizable simulation script is required.

### Tasks

- [x] Introduce the new surplus creation simulation.

KirillDogadin-std:
Closes #448 

Checklist:


core/simulations/config.ts
```diff
@@ -1,5 +1,11 @@
 import debtAuctionSimulation from './configs/createDebtAuctionSimulation';
-import surplusAuctionSimulation from './configs/surplusAuctionSimulation';
+import surplusAuctionBlockchain from './configs/surplusAuctionSimulation';
 import wstethAuctionSimulation from './configs/wstethAuctionSimulation';
+import surplusAuctionSimulation from './configs/createSurplusAcutionSimulation';
 
-export const SIMULATIONS = [debtAuctionSimulation, surplusAuctionSimulation, wstethAuctionSimulation];
+export const SIMULATIONS = [
```
valiafetisov, line 6: One thing that I missed last time: let's create a type that those configs should adhere to and add it here

core/simulations/configs/createSurplusAcutionSimulation.ts
```diff
@@ -0,0 +1,33 @@
+import { HARDHAT_PUBLIC_KEY } from '../../helpers/constants';
+import BigNumber from '../../src/bignumber';
+import { resetNetworkAndSetupWallet, warpTime, addDaiToBalance, addMkrToBalance } from '../../helpers/hardhat';
+import { causeSurplus } from '../../helpers/auctionSimulators';
+
+export default {
```
valiafetisov, line 6: And type every config export

core/helpers/auctionSimulators.ts
```diff
@@ -0,0 +1,187 @@
...
+        return;
+    }
+    throw new Error(`Unexpected Sin value ${sin}, expected ${expected}`);
+};
+
+const overwriteProtocolOwnDaiBalance = async (amount: BigNumber) => {
+    const daiOwnerAddress = await getContractAddressByName(TEST_NETWORK, 'MCD_VOW');
+    await overwriteUintMapping('MCD_VAT', '0x5', daiOwnerAddress, amount.shiftedBy(RAD_NUMBER_OF_DIGITS));
+};
+const ensureOnAuctionDebt = async (expected: BigNumber) => {
```
valiafetisov, line 81: Typo?
```suggestion
const ensureOverallAuctionDebt = async (expected: BigNumber) => {
```

valiafetisov (commented):
Do you also want to create a test like `debt-test.ts` to validate that it works in the CI?

KirillDogadin-std:
> Do you also want to create a test like debt-test.ts to validate that it works in the CI?

with this pr's topic it would be `surplus-test.ts`

I suggest opening a separate issue and resolving it there for both simulations unless you're against it for some reason.

valiafetisov:
Alright, let's open a new issue then to add a full-blown test for each step of the auction!

valiafetisov:
- ✅  Was able to start and participate in the surplus auction using new simulation
  - ✅ setting allowance
  - ✅ bidding
  - ✅ collecting

Screenshot 2022-09-07 at 18 22 47

Although some problems got found which would also be addressed in the next PR:
- Price on uniswap is not fetched
- Next lowest bid is set to 0 (which is not accepted by the protocol). We need to set it to > 0, at least one 1 * 10<sup>-18</sup>
  - Since keeper will be using it to make first bid, we should make it more realistic, eg a uniswap price * gap to reach 1k net profit
- Auction state displays a date, not state
- After the auction ended it still counts down (not up) for some reason (possibly invalid determination of the `earliestEndDate`?)

valiafetisov (approved):

PR #452, opened 2022-09-08
Added helper to generate fake vaults

zoey-kaiser:
This PR expands on #446 and adds generation files to mock the new vault related types.

### DIFF
--- a/frontend/helpers/generateFakeVault.ts
+++ b/frontend/helpers/generateFakeVault.ts
@@ -4,7 +4,6 @@ import {
     Vault,
     VaultAmount,
     VaultBase,
-    VaultTransaction,
     VaultTransactionFees,
     VaultTransactionLiquidated,
     VaultTransactionNotLiquidated,


frontend/helpers/generateFakeVault.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    return {
+        initialDebtDai,
+        collateralAmount,
+    };
+};
+
+const generateFakerLiquidationLimits = function (): LiquidationLimits {
+    const maximumProtocolDebtDai = new BigNumber(faker.finance.amount());
+    const currentProtocolDebtDai = maximumProtocolDebtDai.dividedBy(2);
```
KirillDogadin-std, line 43: could we divide here and below by some natural number that is not higher than 5? Otherwise it seems like too convenient number that might cause us to miss some failing case

KirillDogadin-std (changes requested):

KirillDogadin-std (approved):

frontend/helpers/generateFakeVault.ts
```diff
@@ -0,0 +1,132 @@
...
+    const fakeVault = generateFakeVault();
+    const fakeTransactionFees = generateFakeVaultTransactionFees();
+    const fakeOraclePrices = generateFakeOraclePrices();
+
+    const liquidationRatio = faker.datatype.number({ min: 110, max: 150 });
+    const minUnitPrice = faker.datatype.number();
+    const collateralizationRatio = fakeVault.collateralAmount.multipliedBy(minUnitPrice).toNumber();
+    const proximityToLiquidation = liquidationRatio - collateralizationRatio;
+
+    const state = proximityToLiquidation < 0 ? 'liquidatable' : 'not-liquidateable';
```
valiafetisov, line 99: There is also `liquidated` state

zoey-kaiser: I was not sure how and where I would get the:

- Transaction Hash
- New CollateralAuction info (is it just `collateralType` and `id` or does it change?

from a liquidated vault. Would we need to fetch it separately (like with finished collateral auctions) or how do we plan to integrate this?

valiafetisov: Very good points. I suggest we add `auctionId` that only exist on the liquidated vault? I imagine this info would be possible to fetch from the clip `Kick` event even after the vault had been liquidated. There should also be the transaction hash and the date of the liquidation based on the same event.

@KirillDogadin-std please suggest changes to the types in this or a separate PR? Or directly via the suggestion feature?

KirillDogadin-std: #453

core/src/types.ts
```diff
@@ -320,13 +320,14 @@ export declare interface VaultTransactionBase extends Vault, VaultTransactionFee
 
 export declare interface VaultTransactionLiquidated extends VaultBase {
     state: 'liquidated';
-    liqudiationDate: Date;
-    transactionHash: string;
-    auctionId: string;
+    liqudiationDate?: Date;
+    transactionHash?: string;
+    auctionId?: string;
```
valiafetisov, line 325: Why was this changed? If the vault is liquidated it should have all those values, they are not optional

frontend/helpers/generateFakeVault.ts
```diff
@@ -96,16 +97,26 @@ export const generateFakeVaultTransaction = function (): VaultTransaction {
     const collateralizationRatio = fakeVault.collateralAmount.multipliedBy(minUnitPrice).toNumber();
     const proximityToLiquidation = liquidationRatio - collateralizationRatio;
 
-    const state = proximityToLiquidation < 0 ? 'liquidatable' : 'not-liquidateable';
+    const state: VaultTransactionState =
+        proximityToLiquidation < 0 ? faker.helpers.randomize(['liquidatable', 'liquidated']) : 'not-liquidatable';
```
valiafetisov, line 101: I am afraid that you will need to have a separate functions for the `*Liquidated` and `*NotLiquidated` states, which will then be randomised inside `generateFakeVaultTransaction`

valiafetisov (approved):

PR #454, opened 2022-09-09
Added `VaultLiquidationTransactionTable` and `VaultLiquidationTransactionFlow`

### LINKED ISSUE #437: Create `VaultLiquidationTransactionTable` component
### Goal
Introduce a new `VaultLiqudationTransactionTable` that displays the important liquidation information. In the mocks you can see a small icon. This should be an extension of the explain component, allowing the user to hover over the icon and see a popover with more information. For this you can modify our current `Explain` component to also accept icons as a placeholder, rather then just text.

### Assets
Screenshot 2022-08-30 at 13 51 58

aomafarag:
Closes #437.

Screenshot from 2022-09-09 18-20-27

Checklist:

### Missing

- [x] Add `AnimatedArrow` after merging
- [x] Add `Explain` texts


valiafetisov:
You can remove explainer over the `combined transaction fees` if you're asking about it since there is just one transaction. I would also call it `Transaction fee`

frontend/components/vault/VaultLiquidationTransactionFlow.stories.js
```diff
@@ -0,0 +1,52 @@
...
+import {
+    generateFakeVaultLiquidatedTransaction,
+    generateFakeVaultNotLiquidatedTransaction,
+} from '~/helpers/generateFakeVault';
+
+const common = {
+    components: { VaultLiquidationTransactionFlow },
+    data: () => ({
+        vaultTransaction: generateFakeVaultNotLiquidatedTransaction(),
+        network: 'mainnet',
```
valiafetisov, line 13: The network is also provided inside `vaultTransaction`, so no need to pass it as a separate prop

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,113 @@
...
+            default: true,
+        },
+    },
+    data() {
+        return {
+            isWalletConnected: false,
+        };
+    },
+    computed: {
+        vaultLink(): string | undefined {
```
valiafetisov, line 103: This should be an
```suggestion
        auctionLink(): string | undefined {
```

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,113 @@
...
+            if (!this.vaultTransaction || !this.vaultTransaction.auctionId) {
+                return undefined;
+            }
+            const link = generateLink(this.network, 'collateral');
+            const formattedAuctionId = this.vaultTransaction.auctionId.replace(':', '%3A');
```
valiafetisov, line 108: Please don't try to reimplement [`encodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent)

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,113 @@
...
+            <div slot="message">
+                <p>This vault has been liquidated into a collateral auction</p>
+                <div class="flex justify-end mt-2">
+                    <nuxt-link :to="vaultLink">
+                        <Button> View collateral auction {{ vaultTransaction.auctionId }} </Button>
+                    </nuxt-link>
+                </div>
+            </div>
+        </Alert>
+        <VaultLiquidationTransactionTable class="my-4" :vault-transaction="vaultTransaction" />
```
valiafetisov, line 14: Do you also need to show it if the vault is liquidated?

frontend/components/vault/VaultLiquidationTransactionTable.vue
```diff
@@ -0,0 +1,158 @@
+<template>
+    <div class="flex flex-col space-y-1 text-gray-700 dark:text-gray-100">
+        <div class="flex justify-between">
+            <div>Proximity to liquidation threshold</div>
+            <div>
+                <span v-if="vaultTransaction.proximityToLiquidation < 0" class="text-green-500 mr-1">
+                    (Ready for liquidation)
+                </span>
+                <span v-if="vaultTransaction.proximityToLiquidation">
```
valiafetisov, line 9: In which case there can be no `proximityToLiquidation`? Please remove this `v-if` together with empty `v-else`

frontend/components/vault/VaultLiquidationTransactionTable.vue
```diff
@@ -0,0 +1,158 @@
...
+                    {{ vaultTransaction.proximityToLiquidation }}%
+                </span>
+                <span v-else class="opacity-50"> </span>
+            </div>
+        </div>
+        <div class="flex justify-between">
+            <div>Next price update</div>
+            <div class="flex items-center space-x-1">
+                <template v-if="vaultTransaction.nextPriceChange">
```
valiafetisov, line 18: Same here and for every other key that is required by the types

frontend/components/vault/VaultLiquidationTransactionTable.vue
```diff
@@ -0,0 +1,158 @@
...
+            type: Object as Vue.PropType<VaultTransactionNotLiquidated>,
+            required: true,
+        },
+    },
+    computed: {
+        incentiveRelativePercentage(): String {
+            return (
+                formatToAutomaticDecimalPoints(
+                    this.vaultTransaction.incentiveRelativeDai.div(this.vaultTransaction.debtDai).times(100)
+                ) + '%'
```
valiafetisov, line 144: You shouldn't both compute and format things in one function. Instead, I suggest to compute the value here and then use `FormatPercentage` or other format _component_ to format it (also because the component will support animation in comparison to the `formatToAutomaticDecimalPoints`)

frontend/components/vault/VaultLiquidationTransactionTable.vue
```diff
@@ -0,0 +1,158 @@
...
+            );
+        },
+        getIsPriceGoingUpOrDown(): ArrowDirections {
+            if (!this.vaultTransaction.nextUnitPrice || !this.vaultTransaction.currentUnitPrice) {
+                return 'up';
```
valiafetisov, line 149: It's misleading to show it's going up, if we actually have no idea. I would expect it to be undefined and hidden in case they are not present/unknown/NaN

frontend/components/vault/VaultLiquidationTransactionFlow.vue
zoey-kaiser: It shows that the values are unknown. This is also what we do with the auctions.

frontend/components/vault/VaultLiquidationTransactionTable.vue
zoey-kaiser: When the vault was liquidated there is no more `proximityToLiquidation`. At least when the type is created it does not exist anymore. 

Aside form that I find having fallbacks better then the UI throwing an error if a user ever runs into an edge case.

zoey-kaiser: If the vault was liquidated we do not have these values anymore. There could be an edge case where this happens so I added it as an extra safety measure.

frontend/components/vault/VaultLiquidationTransactionTable.vue
zoey-kaiser: Same as above.

zoey-kaiser:
I am currently running into a bit of an issue. In our previously merged component `LiquidationLimitsCheck` all of the props are set to required. When I add it to the flow, I then run into issues when I display a `LiquidatedVault`, where many of the props are now undefined (as they don't exist on the vault anymore). 

Do I:
- Hide the panel when a vault is liquidated?
- Add fallback values and handle undefined? 

https://github.com/sidestream-tech/unified-auctions-ui/blob/vault-transaction-table/frontend/components/vault/VaultLiquidationTransactionFlow.vue#L44

valiafetisov:
It's of course better for the user to not suddenly hide anything, but in this case I think it's a good trade-off between UX and DX. In case vault is liquidated, I think it's fine to just hide everything in the page except for the title, close sign and the alert specifying what happened. The added value of the table that doesn't show anything and panels that doesn't work is very low.

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -15,11 +15,14 @@ export default Vue.extend({
     props: {
         direction: {
             type: String as Vue.PropType<ArrowDirections>,
-            required: true,
+            default: undefined,
         },
```
valiafetisov, line 19: If no direction is provided, this component doesn't make sense anymore, right? With this approach all props will be optional. Instead, required props are a good way to spot a problem during development and check (via our storybook test) if when this component was used correctly.

Actionable here: undo modifications to the component, add `v-if` to the parent instead

frontend/components/vault/VaultLiquidationTransactionFlow.vue
valiafetisov: See my suggestion here: https://github.com/sidestream-tech/unified-auctions-ui/pull/454#issuecomment-1247681874

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,129 @@
...
+                :collateral-type="vaultTransaction.collateralType"
+                :debt-dai="vaultTransaction.debtDai"
+                :incentive-relative-dai="vaultTransaction.incentiveRelativeDai"
+                :incentive-constant-dai="vaultTransaction.incentiveConstantDai"
+                :is-refreshing="isRefreshingLimits"
+                :is-explanations-shown="isExplanationsShown"
+                :is-correct.sync="areLimitsNotReached"
+                @refreshLimits="$emit('refreshLimits')"
+            />
+            <!-- LiquidationPanel -->
```
valiafetisov, line 55: Let's add it, since it was merged 🙂

frontend/components/vault/VaultLiquidationTransactionTable.vue
```diff
@@ -0,0 +1,159 @@
...
+        <div class="flex justify-between">
+            <div>Proximity to liquidation threshold</div>
+            <div>
+                <span v-if="vaultTransaction.proximityToLiquidation < 0" class="text-green-500 mr-1">
+                    (Ready for liquidation)
+                </span>
+                <span v-if="vaultTransaction.proximityToLiquidation">
+                    {{ vaultTransaction.proximityToLiquidation }}%
+                </span>
+                <span v-else class="opacity-50"> </span>
```
valiafetisov, line 12: If you're going to keep those `if`s, at least display `Unknown` here

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -20,6 +20,9 @@ export default Vue.extend({
     },
     computed: {
         style(): Object {
+            if (!this.direction) {
+                return undefined;
+            }
```
valiafetisov, line 25:
```suggestion
```

frontend/components/common/other/AnimatedArrow.vue
```diff
@@ -1,12 +1,12 @@
 <template>
-    <UpArrow :style="style" />
+    <UpArrow v-if="direction" :style="style" />
```
valiafetisov, line 2:
```suggestion
    <UpArrow :style="style" />
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -111,23 +111,23 @@ export default Vue.extend({
     props: {
         liquidationLimits: {
             type: Object as Vue.PropType<LiquidationLimits>,
+            default: undefined,
```
valiafetisov, line 114: Same here, please set back properties to be required. Here and for all other props in this file

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -111,23 +111,23 @@ export default Vue.extend({
...
+        },
+        collateralType: {
+            type: String,
             required: true,
         },
         debtDai: {
             type: BigNumber,
-            required: true,
+            default: undefined,
```
valiafetisov, line 122: Same

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -139,7 +139,10 @@ export default Vue.extend({
         },
     },
     computed: {
-        debtAndIncentives(): BigNumber {
+        debtAndIncentives(): BigNumber | undefined {
+            if (!this.debtDai || !this.incentiveConstantDai || !this.incentiveRelativeDai) {
+                return undefined;
+            }
```
valiafetisov, line 145: None of this should be possible according to the https://github.com/sidestream-tech/unified-auctions-ui/pull/454#issuecomment-1247681874

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,147 @@
...
+        Alert,
+        Button,
+        WalletConnectionCheckPanel,
+        Explain,
+    },
+    props: {
+        vaultTransaction: {
+            type: Object as Vue.PropType<VaultTransactionNotLiquidated>,
+            required: true,
+        },
```
valiafetisov, line 102: You accept `VaultTransactionNotLiquidated` type inside this panel, but inside the panel handles liquidated state. How is it possible?

Suggestion: keep the prop as is and hide the whole `VaultLiquidationTransactionFlow` in case the vault is liquidated, showing only the alert in the parent?

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,147 @@
...
+        auctionLink(): string | undefined {
+            if (!this.vaultTransaction || !this.vaultTransaction.auctionId) {
+                return undefined;
+            }
+            const link = generateLink(this.vaultTransaction.network, 'collateral');
+            return `${link}&auction=${encodeURIComponent(this.vaultTransaction.auctionId)}`;
+        },
+        wasLiquidated(): boolean {
+            return this.vaultTransaction.state === 'liquidated';
+        },
```
valiafetisov, line 144: I am not sure why types allow this check, since `VaultTransactionNotLiquidated` doesn't have this kind of state. Types doesn't work?

frontend/components/vault/VaultLiquidationTransactionFlow.vue
zoey-kaiser: I don't think this will be a big issue! I changed the type to `VaultTransaction` which has both sub types. I already correctly handled the finished auction state, just forgot to update the type!

frontend/components/vault/VaultLiquidationTransactionFlow.vue
zoey-kaiser: Types in my IDE never worked, but I am surprised it was not caught in the CI. However the comment above should fix it!

valiafetisov: > Types in my IDE never worked

OH WOW! Why was it never reported? Or did I miss it?

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,147 @@
...
+                    />
+                </Explain>
+                is a suggested value based on the current gas prices on the market; the “potential net profit” is also
+                approximate, since it is extrapolated from the exchange rates and may change during the transaction. If
+                you’re executing transactions with other participants at the same time, the one who pays a higher
+                transaction fee has more chances to receive the incentive. In case your transaction will be rejected,
+                it will only result in the loss of the transaction fee.
+            </template>
+        </TextBlock>
+        <div v-if="!wasLiquidated" class="mb-4">
```
valiafetisov, line 35: There is no space between the text and the panels

```suggestion
        <div v-if="!wasLiquidated" class="my-4">
```

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -0,0 +1,147 @@
...
+    },
+    props: {
+        vaultTransaction: {
+            type: Object as Vue.PropType<VaultTransaction>,
+            required: true,
+        },
+        liquidationLimits: {
+            type: Object as Vue.PropType<LiquidationLimits>,
+            required: true,
+        },
```
valiafetisov, line 106: The liquidation limits are actually part of the `vaultTransaction`, but I guess you will fix it when you will put together the container 🙂

valiafetisov (approved):

valiafetisov:
The other thing that will need to be updated soon the liquidated vault state: https://github.com/sidestream-tech/unified-auctions-ui/issues/460#issuecomment-1248295275

PR #457, opened 2022-09-09
Liquidation core functionality

### LINKED ISSUE #460: Implement vault liquidation core logic
### Goal

The PM is able to see specific vault and liquidate it using store actions

### Context

After the conclusion of the related investigation issue, where the [implementation proposal was outlined](https://github.com/sidestream-tech/unified-auctions-ui/issues/363#issuecomment-1234396013), we would implement this functionality in this issue. Current issue may also keep the discussion on the new edge cases.

The basic outline of the proposal:
- Fetch individual vaults from the blockchain directly
- Liquidate a vault
- Provide store functionality

### Assets

- Implementation proposal https://github.com/sidestream-tech/unified-auctions-ui/issues/363#issuecomment-1234396013
- Discussion with the DICU https://github.com/sidestream-tech/auction-ui/issues/433#issuecomment-1235232969
- Related TS types https://github.com/sidestream-tech/unified-auctions-ui/blob/main/core/src/types.ts#L270-L333

### Tasks

- [x] Discuss exact implementation order
  - [x] Eg order of implementation (simulation, liquidation)
- [ ] Implement simulation
- [ ] Implement fetching of a single vault
- [ ] Implement liquidation
- [ ] Implement store-related functionality
  - [ ] Provide test commands in the PR description

KirillDogadin-std:
Closes #460

Checklist:

___

As requested by @LukSteib - instructions to validate store functionality:

 - start simulations via `npm run simulate`
 - choose option in the prompt: `Fork blocks with vaults in various states`
   - this will immediately fork the block where there's a vault in some state, pay attention to the text output. It also will specify the vault id number
 - run in frontend/ dir `npm run dev`
 - open `localhost:3000` in browser, open browser console
 - run the following commands to interact with vaults:
   - `await $nuxt.$store.dispatch('vaults/fetchVault', YOUR_VAULT_NUMBER)` to first load information about the vault into the store
   - `$nuxt.$store.getters['vaults/getVaultById'](YOUR_VAULT_NUMBER)` to print information about the vault in your store
   - `await $nuxt.$store.dispatch('vaults/liquidateVault', YOUR_VAULT_NUMBER)` to liquidate the vault
   - in case you liquidate a vault, call `$nuxt.$store.getters['vaults/getVaultById'](YOUR_VAULT_NUMBER)` again, inspect the value at `pastLiquidations[0].auctionId` to know the id of the created auction, afterwards you'll be able to spot it in the collateral auctions table.


core/src/types.ts
```diff
@@ -299,16 +299,17 @@ export declare interface VaultTransactionFees {
...
-export declare interface Vault extends VaultBase, VaultAmount, LiquidationLimits {}
+export declare interface Vault extends VaultBase, VaultAmount, VaultCollateralParameters, LiquidationLimits {}
 export declare interface OraclePrices {
-    currentUnitPrice: BigNumber;
-    nextUnitPrice: BigNumber;
-    nextPriceChange: Date;
+    // can be set to NaN
+    currentUnitPrice: BigNumber | number;
+    nextUnitPrice: BigNumber | number;
+    nextPriceChange?: Date;
```
valiafetisov, line 307: Date can also be invalid btw, `new Date(NaN)`, but we don't handle it automatically in the components yet

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    length: 2,
+});
+
+const _fetchLiquidationRatio = async (network: string, collateralType: CollateralType) => {
+    const contract = await getContract(network, 'MCD_SPOT');
+    const collateralTypeAsHex = ethers.utils.formatBytes32String(collateralType);
+    const oracleAndLiquidationRatio = await contract.ilks(collateralTypeAsHex);
+    const liquidationRatioAsHex = oracleAndLiquidationRatio.mat;
+    const oracleAddress = oracleAndLiquidationRatio.pip;
+    const liquidationRatio = new BigNumber(liquidationRatioAsHex._hex).shiftedBy(-RAY_NUMBER_OF_DIGITS).toNumber();
```
valiafetisov, line 231: I am not sure it's a good idea to turn ratio that can be so precise into a number. Let's keep it as BigNumber

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    };
+};
+
+export const fetchVaultLiquidationIncentive = memoizee(_fetchVaultLiquidationIncentive, {
+    maxAge: CACHE_EXPIRY_MS,
+    promise: true,
+    length: 3,
+});
+
+const _isVaultLiquidated = async (network: string, vault: Vault) => {
```
valiafetisov, line 269: Do liquidated vaults stay within cdp manager and still can be fetched from there?

Anyway, the name of the function is misleading, I would call it `fetchLiquidatedParameters` and return undefined if there is nothing

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    const contract = await getContract(network, 'MCD_DOG');
+    const typeHex = ethers.utils.formatBytes32String(vault.collateralType);
+    const eventFilter = contract.filters.Bark(typeHex, vault.address, null, null, null, null, null);
+    const liquidationEvents = await contract.queryFilter(eventFilter);
+    if (liquidationEvents.length !== 0 && vault.initialDebtDai.eq(0)) {
+        // there was a liquidation and the vault was not used again
+        const latestEvent = liquidationEvents[liquidationEvents.length - 1];
```
valiafetisov, line 276: Btw, what happens if the vault was partially liquidated? Would there be two barking events? Should we then turn this data into an array of objects?

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+        return {
+            isLiquidated: true,
+            liquidationDate: await fetchDateByBlockNumber(network, latestEvent.blockNumber),
+            transactionHash: latestEvent.transactionHash,
+            auctionId: new BigNumber(latestEvent.args?.id._hex).toFixed(),
```
valiafetisov, line 281: This should contain collateral type in the format `ETH-A:1`

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+        debtDai
+    );
+
+    const { liquidationRatio, oracleAddress } = await fetchLiquidationRatio(network, vault.collateralType);
+    const collateralizationRatio = vault.collateralAmount
+        .multipliedBy(vault.minUnitPrice)
+        .multipliedBy(liquidationRatio)
+        .dividedBy(debtDai)
+        .toNumber();
+    const proximityToLiquidation = collateralizationRatio - liquidationRatio;
```
valiafetisov, line 313: Would this stay within 0-1? I thought it will be not subtraction but division for some reason

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    );
+    const osmPrices = await getOsmPrices(network, oracleAddress);
+    const { nextUnitPrice, nextPriceChange, currentUnitPrice } = osmPrices
+        ? osmPrices
+        : {
+              nextUnitPrice: new BigNumber(NaN),
+              currentUnitPrice: new BigNumber(NaN),
+              nextPriceChange: undefined,
+          };
+    const state = proximityToLiquidation < 0 ? 'liquidatable' : 'not-liquidatable';
```
valiafetisov, line 325: This logic doesn't account for the limits, right? If the limits are reached already, the transaction wouldn't go trough. If it will be reached after the execution, it will be done partially (so maybe the `incentiveRelativeDai` will be affected?)

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    return await enrichVaultWithTransactonInformation(network, vault);
+};
+
+export const getVaultTransaction = memoizee(_getVaultTransaction, {
+    maxAge: CACHE_EXPIRY_MS,
+    promise: true,
+    length: 2,
+});
+
+export const liquidateVault = async (network: string, vault: Vault, incentiveBeneficiaryAddress: string) => {
```
valiafetisov, line 372: Can we instead accept `collateralType` and `vaultAddress` here? `incentiveBeneficiaryAddress` should be optional and is taken form the signer if not provided

frontend/store/vaults.ts
```diff
@@ -0,0 +1,74 @@
+import { Vault, VaultTransaction } from 'auctions-core/src/types';
+import Vue from 'vue';
+import { fetchVault, fetchVaultsCount, liquidateVault } from 'auctions-core/src/vaults';
+import { ActionContext } from 'vuex';
+
+interface State {
+    vaultTransactions: Record<Vault['id'], VaultTransaction>;
+    vaultsCount?: number;
```
valiafetisov, line 8: Why do you need `vaultsCount` on the frontend? Is it displayed somewhere?

frontend/store/vaults.ts
```diff
@@ -0,0 +1,74 @@
...
+            console.error(`Failed to fetch vault count: ${e}`);
+        } finally {
+            commit('setIsVaultLoading', false);
+        }
+    },
+    async liquidateVault({ rootGetters, commit, getters }: ActionContext<State, State>, vaultId: number) {
+        const network = rootGetters['network/getMakerNetwork'];
+        const walletAddress = rootGetters['wallet/getAddress'];
+        const vaultTransaction = getters.vaultTransactions[vaultId];
+        commit('setIsVaultLoading', true);
```
valiafetisov, line 65: Please create separate variables for fetching and execution

frontend/store/vaults.ts
```diff
@@ -0,0 +1,74 @@
...
+    setVaultsCount(state: State, count: number) {
+        state.vaultsCount = count;
+    },
+};
+export const actions = {
+    async fetchVault({ commit, rootGetters }: ActionContext<State, State>, vaultId: number) {
+        const network = rootGetters['network/getMakerNetwork'];
+        commit('setIsVaultLoading', true);
+        try {
+            const vaultTransaction = await fetchVault(network, vaultId);
```
valiafetisov, line 41: What happens here if I try to fetch index that doesn't exist?

core/src/vaults.ts
KirillDogadin-std: > Would this stay within 0-1?

nope, not guaranteed. 

> I thought it will be not subtraction but division for some reason

when i think about it more, i come to conclusion that it's confusing for the user to see both options (subtraction as it is done now and relative value). "Collateralization ratio" is itself a metric that is designed to compare against liquidation ratio: mathematically it means that regardless of what we do ( `-` or `/` ) it comes down to the comparison of the same values: `debtDai` vs `collateralAmount * minUnitPrice`. So i suggest instead just show this proximity as follows:

  - in dai
  - how much more debt should be there to make vault liquidatable (`collateralAmount * minUnitPrice - debtDai`)

core/src/vaults.ts
KirillDogadin-std: if vault is partially liquidated, its debt is not 0. there's the check in the if statement for this.

Dilemma here: we have 2 states - which one should be assigned when there're 3 outcomes?

- I did not see logic in the contracts that forbids the owner from adding collateral and generating dai using the previously liquidated vault.
  - therefore i assume that presence of the event does not necessarily guarantee that that the vault is liquidated.
- Vault can be partially liquidated:
  - event is present, 
  - debt is non-0
  - this does not differ from the previous point where the vault is liquidated and re-used afterwards

Due to the 2 points above, i do not see how these cases can be distinguished and thus i did not add the third `partially-liquidated` state. Instead i assign the vault `not-liquidated` state.

In the ui after the partial liquidation the vault will still is present but will have `not-liquidatable` state due to the limits being hit.

We can distinguish wether the liquidation has been successful by the success of the liquidation function, then we can have the process that searches for the bark event with the specified auction id to fetch relevant to the final ui panel information. Shall i implement this?

core/src/vaults.ts
KirillDogadin-std: > Do liquidated vaults stay within cdp manager and still can be fetched from there?

yes.

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+    while (osmEvents.length === 0 && latestBlockNumber > rpcEventFilterSegmentIndex * EVENTS_PER_RPC_REQUEST) {
+        osmEvents = await osmContract.queryFilter(osmEventFilter, fromBlock, toBlock);
+        rpcEventFilterSegmentIndex += 1;
+        fromBlock = -EVENTS_PER_RPC_REQUEST * (rpcEventFilterSegmentIndex + 1) - EVENTS_ON_FIRST_RPC_REQUEST;
+        toBlock = -EVENTS_PER_RPC_REQUEST * rpcEventFilterSegmentIndex - EVENTS_ON_FIRST_RPC_REQUEST;
+    }
+    if (osmEvents.length === 0) {
+        return undefined;
+    }
+    const currentUnitCollateralPrice = new BigNumber(osmEvents[osmEvents.length - 1].args?.val).shiftedBy(
```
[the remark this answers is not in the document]
KirillDogadin-std: not relevant in this place after the slot-based price fetch.

as to answer the question: official docs do not mention anything, found this though https://community.infura.io/t/is-the-return-from-getlogs-in-order-by-blocknumber/3591 - seems like it's in order

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+});
+
+const _getOsmPrices = async (network: string, oracleAddress: string): Promise<OraclePrices | undefined> => {
+    const provider = await getProvider(network);
+    const latestBlockNumber = await fetchLatestBlockNumber(network);
+    let rpcEventFilterSegmentIndex = 0;
+
+    const osmContractInterface = await getContractInterfaceByName('OSM');
+    const osmContract = new ethers.Contract(oracleAddress, osmContractInterface, provider);
+    const osmEventFilter = osmContract.filters.LogValue(null);
```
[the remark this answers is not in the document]
KirillDogadin-std: not relevant after the slot-based price fetches.

to answer the question:

seems like it's possible if the filter is constructed via more low-level tools. https://docs.ethers.io/v5/concepts/events/

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+});
+
+const _fetchCollateralLiquidationLimitsAndLiquidatorAddress = async (network: string, type: string) => {
+    const contract = await getContract(network, 'MCD_DOG');
+    const typeHex = ethers.utils.formatBytes32String(type);
+    const { hole, dirt, clip } = await contract.ilks(typeHex);
+    return {
+        currentCollateralDebtDai: new BigNumber(dirt._hex).shiftedBy(-RAD_NUMBER_OF_DIGITS),
+        maximumCollateralDebtDai: new BigNumber(hole._hex).shiftedBy(-RAD_NUMBER_OF_DIGITS),
+        liquidatiorContractAddress: clip,
```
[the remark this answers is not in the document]
KirillDogadin-std: i use it in `_enrichVaultWithTransactonInformation`

core/src/helpers/blockNumber.ts
```diff
@@ -0,0 +1,16 @@
...
+const _fetchLatestBlockNumber = async (network: string) => {
+    const provider = await getProvider(network);
+    return await provider.getBlockNumber();
+};
+
+const fetchLatestBlockNumber = memoizee(_fetchLatestBlockNumber, {
+    promise: true,
+    length: 0,
+    maxAge: CACHE_EXPIRY_MS,
+});
```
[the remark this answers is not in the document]
KirillDogadin-std: file was not needed anymore and is removed.

frontend/store/vaults.ts
KirillDogadin-std: removed

frontend/store/vaults.ts
KirillDogadin-std: The error is thrown due to changes from the previous comments.

core/src/vaults.ts
valiafetisov: > in dai

This doesn't let the user _sort_ by the most risky vaults (something that we planned to do by default in the table). In your case, a small vault with 1.000 DAI "DAI proximity" will be much further from being liquidated than a vault with 10.000 DAI "DAI proximity". My suggestion is to keep it being a percentage: it can be the same value as you suggest by divided by the overall size of the vault in DAI: `(collateralAmount * minUnitPrice - debtDai) / collateralAmount * minUnitPrice`? (although I'm not sure if your formula is correct: don't we need to account for some fees/currentUnitPrice or debtDai is already derived from them?)

core/src/vaults.ts
```diff
@@ -0,0 +1,448 @@
...
+        network
+    );
+    const { nextUnitPrice, nextPriceChange, currentUnitPrice } = await getOsmPrices(
+        network,
+        oracleAddress,
+        vault.collateralType
+    );
+    let state: 'liquidatable' | 'not-liquidatable' = proximityToLiquidation.isLessThanOrEqualTo(0)
+        ? 'liquidatable'
+        : 'not-liquidatable';
```
valiafetisov, line 356: Instead of making it `let`, can you please split state logic into separate function(s)?

core/src/vaults.ts
valiafetisov: - [ ] Can we instead accept `collateralType` and `vaultAddress` here?

frontend/store/vaults.ts
```diff
@@ -0,0 +1,74 @@
...
+        commit('setIsVaultLoading', true);
+        try {
+            const vaultCount = await fetchVaultsCount(network);
+            commit('setVaultsCount', vaultCount);
+        } catch (e) {
+            console.error(`Failed to fetch vault count: ${e}`);
+        } finally {
+            commit('setIsVaultLoading', false);
+        }
+    },
```
valiafetisov, line 60: If you [removed `vaultCount`](https://github.com/sidestream-tech/unified-auctions-ui/pull/457#discussion_r970492513) you should also remove this action

core/src/types.ts
```diff
@@ -331,3 +340,11 @@ export declare interface VaultTransactionNotLiquidated extends VaultTransactionB
 
 export type VaultTransaction = VaultTransactionLiquidated | VaultTransactionNotLiquidated;
 export type VaultTransactionState = VaultTransaction['state'];
+
+export type PriceOracleType = 'dsThing' | 'default' | 'univ';
```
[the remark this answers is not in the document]
KirillDogadin-std: Actually these types are outdated and have to be removed, overlooked it.

core/src/types.ts
```diff
@@ -331,3 +340,11 @@ export declare interface VaultTransactionNotLiquidated extends VaultTransactionB
...
+export declare interface PriceOracleConfig {
+    currentPriceSlotAddress: string;
+    nextPriceSlotAddress?: string;
+    whitelistSlotAddress?: string;
+    hasDelay: boolean;
```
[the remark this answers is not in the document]
KirillDogadin-std: >  I would expect that this type is used in the collateral config

why? if the delay happens in the price oracle, then i would assume that we should define it in the price oracle config.

core/src/vaults.ts
KirillDogadin-std: made the change for this formula,

> although I'm not sure if your formula is correct: don't we need to account for some fees/currentUnitPrice or debtDai is already derived from them?

no, proximity to liquidation is not determined based on the fees, only on raw numbers of `initialDebt`, `stabilityFeeRate`, `collateralAmount`, `minUnitPrice`.  https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/dog.sol#L181

core/src/vaults.ts
KirillDogadin-std: reading is hard 😞

core/src/vaults.ts
KirillDogadin-std: >  Otherwise when and why do we label the vault as liquidated if their state is indistinguishable from the live ones?

`liquidated` is set if the debt is `0`:

  - in case it was liquidated: the debt is zero and we are correct in stating that is liquidated
  - in case it was liquidated once, then reused somehow and after sometime the debt is 0 due to the owner's actions: we claim it's liquidated. This is not entirely true but makes no difference as far as i see it.

core/src/vaults.ts
```diff
@@ -0,0 +1,378 @@
...
+import memoizee from 'memoizee';
+import fetchLatestBlockNumber from './helpers/blockNumber';
+
+const CACHE_EXPIRY_MS = 24 * 60 * 60 * 1000;
+const EVENTS_PER_RPC_REQUEST = 100000;
+const EVENTS_ON_FIRST_RPC_REQUEST = 200;
+
+const _fetchVaultBase = async (network: string, id: number): Promise<VaultBase> => {
+    const contract = await getContract(network, 'CDP_MANAGER');
+    const address = await contract.urns(id);
```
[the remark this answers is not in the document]
KirillDogadin-std: > What happens if the vault doesn't exist?

the method returns 0x0000 since it's a blockchain.

made this check happen

core/src/vaults.ts
KirillDogadin-std: i am not sure that it's desirable to remove `let`: since this place repeats the logic from blockchain, `state` is changed in places where the contract would throw error.

extracted to different function

core/src/types.ts
```diff
@@ -267,13 +274,15 @@ export declare interface DebtAuctionEnriched extends DebtAuctionActive {
...
-type CollateralType = CollateralConfig['title'];
+export type CollateralType = CollateralConfig['title'];
 
 export declare interface LiquidationLimits {
     maximumProtocolDebtDai: BigNumber;
     currentProtocolDebtDai: BigNumber;
     currentCollateralDebtDai: BigNumber;
     maximumCollateralDebtDai: BigNumber;
+    liquidationPenalty: BigNumber;
+    minimalAuctionedDai: BigNumber;
```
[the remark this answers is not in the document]
KirillDogadin-std: https://github.com/sidestream-tech/unified-auctions-ui/issues/460#issuecomment-1248180931

core/src/types.ts
```diff
@@ -84,14 +84,37 @@ export declare interface UniswapV2LpTokenCalleeConfig {
...
     symbol: string;
     decimals: number;
     exchange: RegularCalleeConfig | UniswapV2LpTokenCalleeConfig;
+    oracle: CollateralPriceSourceConfig;
+}
+
+interface OracleConfigBase {
+    hasDelay: boolean;
+    slotPriceValueBeginsAtPosition: number;
+}
```
valiafetisov, line 104: - [ ] Why `currentPriceSlotAddress` is not part of the `OracleConfigBase` if it's part of the both `OracleCurrentAndNextPrices` and `OracleCurrentPriceOnly`?

- [ ] Is it now just 2 types, not 3?

core/src/types.ts
```diff
@@ -84,14 +84,37 @@ export declare interface UniswapV2LpTokenCalleeConfig {
...
+
+export declare interface OracleCurrentAndNextPrices extends OracleConfigBase {
+    currentPriceSlotAddress: string;
+    nextPriceSlotAddress: string;
+}
+
+export declare interface OracleCurrentPriceOnly extends OracleConfigBase {
+    currentPriceSlotAddress: string;
+    currentPriceValiditySlotAndOffset: ValueSlotAddressAndOffset;
 }
```
valiafetisov, line 114: Do you want to explicitly specify types to
- depend on those types in the code instead of `'key' in oracle` checks
- refer to it in the readme for clarity?

```suggestion
export declare interface OracleCurrentAndNextPrices extends OracleConfigBase {
    type: 'OracleTypeOne'
    currentPriceSlotAddress: string;
    nextPriceSlotAddress: string;
}

export declare interface OracleCurrentPriceOnly extends OracleConfigBase {
    type: 'OracleTypeTwo'
    currentPriceSlotAddress: string;
    currentPriceValiditySlotAndOffset: ValueSlotAddressAndOffset;
}
```

(just a suggestion)

core/src/vaults.ts
```diff
@@ -0,0 +1,446 @@
...
+        return new BigNumber(NaN);
+    }
+    const slot = oracle.nextPriceSlotAddress;
+    const nextPriceFeed = await provider.getStorageAt(oracleAddress, slot);
+    const valueSplitPosition = oracle.slotPriceValueBeginsAtPosition;
+    const isPriceValid = nextPriceFeed.substring(0, valueSplitPosition);
+    if (parseInt(isPriceValid, 16) === 1) {
+        return new BigNumber(`0x${nextPriceFeed.substring(valueSplitPosition)}`);
+    }
+    return new BigNumber(NaN);
```
valiafetisov, line 184: As [previously suggested](https://github.com/sidestream-tech/unified-auctions-ui/pull/457#discussion_r971857568): 
- [ ] please add a comment in the code explaining what's happening here (for me it looks like a bunch of hacks with no meaning)
- [ ] please add section to the "collateral onboarding" guide to explain how to determine which collateral type it is and where to get those values 
- [ ] please provide a method to validate that those values are still valid (eg right now it is tested on a specific block, but how to check if those are still correct if when we run it on the latest block?)

core/src/vaults.ts
```diff
@@ -0,0 +1,446 @@
...
+    // logic from https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/dog.sol#L186
+    // detemines if the vault is liquidatable and what amount of debt can be covered.
+    let state: 'liquidatable' | 'not-liquidatable' = proximityToLiquidation.isLessThanOrEqualTo(0)
+        ? 'liquidatable'
+        : 'not-liquidatable';
+    const amountDaiCanBeAuctionedGloballyDai = vault.maximumProtocolDebtDai.minus(vault.currentProtocolDebtDai);
+    const amountDaiCanBeAuctionedCollateralDai = vault.maximumCollateralDebtDai.minus(vault.currentCollateralDebtDai);
+    const minimumDebtCovered = amountDaiCanBeAuctionedCollateralDai.isLessThan(amountDaiCanBeAuctionedGloballyDai)
+        ? amountDaiCanBeAuctionedCollateralDai
+        : amountDaiCanBeAuctionedGloballyDai;
```
valiafetisov, line 328: Can this also be written with `BigNumber.min`?

core/src/vaults.ts
```diff
@@ -0,0 +1,446 @@
...
+    if (amountDaiCanBeAuctionedGloballyDai.eq(0) || amountDaiCanBeAuctionedCollateralDai.eq(0)) {
+        state = 'not-liquidatable';
+    }
```
valiafetisov, line 331: Why exactly 0 is a separate edge case? Don't we need to check if it's smaller OR equal than that?

core/src/vaults.ts
```diff
@@ -0,0 +1,446 @@
...
+            if (
+                auctionedAmountDai
+                    .multipliedBy(vault.stabilityFeeRate)
+                    .isGreaterThanOrEqualTo(vault.minimalAuctionedDai)
+            ) {
+                state = 'not-liquidatable';
+            }
+        }
+    }
+    return { state, auctionedAmountDai };
```
valiafetisov, line 354: What is `auctionedAmountDai`? Amount of DAI that will be auctioned in case of the partial auction which can be smaller than `debtDai`? Maybe `debtThatCanBeAuctionedDai`?

I suggest to split this into two functions for readability: one that computes `debtThatCanBeAuctionedDai`, the other one which determines state checking if `debtThatCanBeAuctionedDai` bigger than 0 and the minimum

valiafetisov: Also, I don't know if you know noticed it, but limits also account for the incentives

core/src/vaults.ts
```diff
@@ -0,0 +1,446 @@
...
+        nextPriceChange = new Date((lastPriceUpdateTimestampInSeconds + priceUpdateFrequencyInSeconds) * 1000);
+    }
+    return nextPriceChange;
+};
+
+const _getOsmPrices = async (
+    network: string,
+    oracleAddress: string,
+    collateralType: CollateralType
+): Promise<OraclePrices> => {
```
valiafetisov, line 227: Let's remove oracle logic into a separate file, it's too long and complicated and surely deserves a better structure

core/src/types.ts
KirillDogadin-std: > Is it now just 2 types, not 3?

core/src/vaults.ts
KirillDogadin-std: > but how to check if those are still correct if when we run it on the latest block?)

i don't think this one is feasible. Since we receive bytes as the result of memory access, we cannot be sure that the accessed value is still the one we're looking for. The only case where we can likely say that the logic is broken is to fetch 0 value (since price 0 is not possible???).

core/src/vaults.ts
KirillDogadin-std: From the contract i'vd read that this difference cannot fall below 0. But comparing with <= instead of === is safer ;)

core/src/vaults.ts
KirillDogadin-std: > debtThatCanBeAuctionedDai

actually `initialDebtThatCanBeAuctionedDai`

KirillDogadin-std: > Also, I don't know if you know noticed it, but limits also account for the incentives

this seems out of context and i can't figure what you're referring to here. could you elaborate?

core/README.md
```diff
@@ -33,3 +40,13 @@ The process of adding new collaterals depends on the token type used. This is du
         - The file should be named using the name from `1.`
         - The file should export `CalleeFunctions`
         - The file should be imported in the [`calleeFunctions/index.ts`](./src/calleeFunctions/index.ts)
+3. Adding price oracle configurations for the token:
+    1. Get the source code of the price oracle contract:
+       - read `ilks(collateralType)` from [`Spot` contract](https://etherscan.io/address/0x65c79fcb50ca1594b025960e539ed7a9a6d434a3#code)
+    2. Read the contract and determine the slot address of the variable:
+       - Depending on the variable type it can be stored in different slots for memory consumption minimization
+       - https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html documentation is helpful
+       - Generally the slot number increments with each (big enough) variable in top-to-bottom manner
```
valiafetisov, line 49:
```suggestion
       - Generally a slot number can be determined by counting definition of variables in the contract source code, but there are exceptions, [please read the docs on the solidity version the contract was compiled with](https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html)
```

core/README.md
```diff
@@ -33,3 +40,13 @@ The process of adding new collaterals depends on the token type used. This is du
...
+       - Experimenting with blockchain fork (e.g. hardhat) helps: try to fetch the value you're looking for / overwrite it / ... and validate that it's correct via some public method or comparing against your expectation.
```
valiafetisov, line 50: This doesn't explain how to actually experiment with it.

Currently, even me with internal knowledge of the repo, contracts and the process, wouldn't be able to add a new collateral because this instruction doesn't explain it practical enough. Where to look? What commands to execute? How to understand they are correctly executed? What `private variables` need to be extracted? What to do with the slot address after it was determined? What is the difference between `CurrentAndNextPrice` and the other type? What is `hasDelay` value and how to determine it? Where is the example of the config that needs to be added? What is `offset`? What is `slotPriceValueBeginsAtPosition`? etc, I can keep going with every other value.

Please explain it in details that even external developer can add a new collateral / fix existing one, provide links to the source code. Provide helper functions/scripts to determine slots if this will simplify the process

core/README.md
```diff
@@ -2,6 +2,13 @@
 
 Set of typescript functions, tests and simulation to work with MakerDAO auctions.
 
+## Collateral price extraction
+
+The functionality of the tool includes fetching collateral prices from the blockchain in order to determine wether the vault is or will be undercollateralized.
+The price is fetched from the private variables of the contract. Since the variables are private, they are only accessible via memory address. Different contracts have different memory layouts.
+On top of that the information that is contained in the contracts differs - e.g. some of them do not contain future prices.
+Therefore it is necessary to specify the memory addresses for each collateral type since there's one price source (contract) per collateral. The fetching is done based on the provided values.
```
valiafetisov, line 10: This text is missing the context and obviously shouldn't be above `Simulations` and `Environment Variables` sections. It's mostly important to know this information in case new collateral is added, so I would add it as subsection to `Collateral onboarding` or remove completely as it doesn't add a value besides describing the obvious (already enforced by the types)

core/src/oracles.ts
```diff
@@ -0,0 +1,135 @@
...
+     * The price is stored in the same slot as a validity marker:
+     *   - split the received value into validity marker part and price part
+     * If the price is valid - return it, otherwise provide a NaN
+     **/
+    if (oracle.type !== 'CurrentAndNextPrice') {
+        return new BigNumber(NaN);
+    }
+    const slot = oracle.nextPriceSlotAddress;
+    const nextPriceFeed = await provider.getStorageAt(oracleAddress, slot);
+    const valueSplitPosition = oracle.slotPriceValueBeginsAtPosition;
```
valiafetisov, line 35: Why do you need to rename variables? This is always a bad practice: you should call the same thing same everywhere to increase readability

core/src/oracles.ts
```diff
@@ -0,0 +1,135 @@
...
+     * If yes, fetch the price by direct memory access (via slot address)
+     * The price is stored in the same slot as a validity marker:
+     *   - split the received value into validity marker part and price part
+     * If the price is valid - return it, otherwise provide a NaN
+     **/
+    if (oracle.type !== 'CurrentAndNextPrice') {
+        return new BigNumber(NaN);
+    }
+    const slot = oracle.nextPriceSlotAddress;
+    const nextPriceFeed = await provider.getStorageAt(oracleAddress, slot);
```
valiafetisov, line 34: If this variable contains both price and the validity, maybe that's how it should be called?
```suggestion
    const nextPriceAndValidityHex = await provider.getStorageAt(oracleAddress, slot);
```

core/src/oracles.ts
```diff
@@ -0,0 +1,135 @@
...
+         * The price is stored in the same slot as a validity marker:
+         *   - split the received value into validity marker part and price part
+         * If the price is valid - return it, otherwise provide a NaN
+         **/
+        const currentPriceFeed = await provider.getStorageAt(oracleAddress, oracle.currentPriceSlotAddress);
+        const valueSplitPosition = oracle.slotPriceValueBeginsAtPosition;
+        const isPriceValid = parseInt(currentPriceFeed.substring(0, valueSplitPosition), 16) === 1;
+        return isPriceValid
+            ? new BigNumber(`0x${currentPriceFeed.substring(valueSplitPosition)}`)
+            : new BigNumber(NaN);
```
valiafetisov, line 81: This code seems to be same/similar to `getNextOraclePrice`. What's the difference? Can the code be reused?

core/src/oracles.ts
```diff
@@ -0,0 +1,135 @@
...
+         *   - extract the byte that contains validity marker only
+         * If the price is valid - return it, otherwise provide a NaN
+         **/
+        const currentPriceFeed = await provider.getStorageAt(oracleAddress, oracle.currentPriceSlotAddress);
+        const valueSplitPosition = oracle.slotPriceValueBeginsAtPosition;
+        const storageValue = await provider.getStorageAt(oracleAddress, oracle.currentPriceValiditySlotAndOffset.slot);
+        const isPriceValid = parseInt(storageValue[oracle.currentPriceValiditySlotAndOffset.offset], 16) === 1;
+        return isPriceValid
+            ? new BigNumber(`0x${currentPriceFeed.substring(valueSplitPosition)}`)
+            : new BigNumber(NaN);
```
valiafetisov, line 62: This code seems to be same/similar to `getNextOraclePrice`. What's the difference? Can the code be reused?

core/src/oracles.ts
```diff
@@ -0,0 +1,135 @@
...
+    const nextPrice = await getNextOraclePrice(collateralConfig.oracle, provider, oracleAddress);
+    const currentPrice = await getCurrentOraclePrice(collateralConfig.oracle, provider, oracleAddress);
+    const nextPriceChange = await getNextOraclePriceChange(collateralConfig, provider, oracleAddress);
+
+    const currentUnitCollateralPrice = new BigNumber(currentPrice).shiftedBy(-WAD_NUMBER_OF_DIGITS);
+    const nextUnitCollateralPrice = new BigNumber(nextPrice).shiftedBy(-WAD_NUMBER_OF_DIGITS);
+
+    return {
+        currentUnitPrice: currentUnitCollateralPrice,
+        nextUnitPrice: nextUnitCollateralPrice || new BigNumber(NaN),
```
valiafetisov, line 126: In which case this will be falsy? I can see that `nextUnitCollateralPrice` will always contain `new BigNumber()`

KirillDogadin-std: removed the right part

core/README.md
```diff
@@ -33,3 +40,13 @@ The process of adding new collaterals depends on the token type used. This is du
         - The file should be named using the name from `1.`
         - The file should export `CalleeFunctions`
         - The file should be imported in the [`calleeFunctions/index.ts`](./src/calleeFunctions/index.ts)
+3. Adding price oracle configurations for the token:
+    1. Get the source code of the price oracle contract:
```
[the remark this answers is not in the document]
KirillDogadin-std: This and the previous comments should now be answered by the subbullets.

core/README.md
KirillDogadin-std: added a section

core/README.md
```diff
@@ -33,3 +33,31 @@ The process of adding new collaterals depends on the token type used. This is du
...
+  3.1. If you're not authorized - you have to overwrite the mapping that stores the authorizations before reading or calling any of the functions. The process is no different from overwriting any other value.
+4. Overwrite the value on the running hardhat fork, write your script that does the following:
+   - get the provider object. (e.g. call `getProvider` from `provider.ts`)
+   - determine the contract name (or add it). Use `contract.ts` file.
+   - Call the overwrite function, most likely `/helpers/hardhat` package might be of help here since it already has some funcitonality for overwrites defined.
+   - validate that the overwrite was successful:
+       - if you were whitelisting yourself (giving auth to your wallet) - call a function of the contract that requires this auth:
+           - `overwriteUIntMapping`
+           - `overwriteUIntValue`
+       - if you are whitelisted, call the function of the contract (`await contract.function_name(...params)`) that returns the price. Compare it with the value you have submitted.
```
valiafetisov, line 62: This part seems like you're explaining how to write a (test) script. But it's not specific enough. Eg `determine the contract name (or add it). Use contract.ts file` – which contract name, how to determine it?

Can you please instead provide the script(s)? The script that will ask collateral name via prompt, print its OSM address as a url to etherscan and the instruction to count something specific, then wait for the count, then ask for something else, eg method name, validate it, and finally print the oracle config. Otherwise, I am afraid we will keep iterating this readme indefinitely. As [most of the questions asked previously](https://github.com/sidestream-tech/unified-auctions-ui/pull/457#discussion_r974307134) still not answered:
- What `private variables` need to be extracted? Do you have their names, their types, other ways of understanding that?
- What to do with the slot address after it was determined? In which exact variable to put it, in which format?
- What is the difference between `CurrentAndNextPrice` and the other type? How to determine which one is which?
- What is `hasDelay` value and how to determine it?
- Where is the example of the config that needs to be added?
- What is `offset`? How to determine it?
- What is `slotPriceValueBeginsAtPosition`? How to determine it?

valiafetisov (commented):
I see 3 options to move forward:
1. We return back into discussing implementation of price fetching via events and the limitations of such approach (since this implementation is much simpler and can be generic) and if we can initially release feature with it
2. You write an "onboarding" script via prompts that asks for specific things, then validates them, then outputs the final oracle config
3. You rewrite the readme with step-by-step code examples and more clarity about config structure (after two iterations this seems to be the option which will sink much more time than others, as it is in general very hacky and hard to explain and validate)

core/src/vaults.ts
valiafetisov: > limits also account for the incentives

Do you account for combined incentive when determining if auction will be executed partially? Because when vault is liquidated the limits are increased not only on the amount of debt, but also on the amount of incentives. Do you account for it or am I wrong?

core/src/vaults.ts
valiafetisov: So the only option is to manually ballpark received number or how exactly do you manually validate it?

Then, should we record the address of the oracle in the config to make sure it hasn't changed? And make a script that alerts user if the contract is different (eg in the CI)?

core/src/getCollateralPriceOracleConfig.ts
```diff
@@ -0,0 +1,210 @@
...
+    const collateralType = await promptCollateralType();
+    const { contract, address } = await getOracleAddressAndContract(collateralType);
+    console.info(`Contract address: ${address}`);
+    if (basicInfo.hasWhitelist) {
+        const { whitelistSlot, whitelistFunction } = await promptWhitelistSlotAndFunction();
+        await addToWhitelist(address, parseInt(whitelistSlot));
+        const isWhitelisted = await callFunction(contract, whitelistFunction, HARDHAT_PUBLIC_KEY);
+        if (isWhitelisted === '0x00') {
+            throw new Error('Failed to whitelist the wallet on the fork');
+        }
```
KirillDogadin-std, line 195: @valiafetisov here the logic should overwrite the mapping, but it doesn't do it AND does not throw an error.
 - does not seem that it's the issue with address calculation because
   - calculating the address of the whitelisting flag for `spot` contract instead of hardhat public key provides the expected/correct result (flag in the auth mapping is `1`)
   - if i set the value by `hardhat_setStorageAt` to 1 and read it by `eth_getStorageAt`, i still get 0 while keeping track of the same address being used.

KirillDogadin-std: If you will want to run it yourself, here's the current input log to reach the error:

```
> npx hardhat run ./src/getCollateralPriceOracleConfig.ts                                                                                                         Y liquidation_core_functionality 21.Sep.22__12:44 ‚ôøÔ∏è
Forking at the latest block
New RPC endpoint has started on http://localhost:8545
Using wallet with:
         - public key: 0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266
         - private key: 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80
- Does the contract have whitelisting for method execution? : yes
- Does contract provide future price? : yes
- State the collateral type. E.g. ETH-A : ETH-A
Contract address: 0x81FE72B5A8d1A857d176C3E7d5Bd2679A9B85763

    For this step you must guess the slot number.
    1. Find the function that returns the variable, provide its name.
    2. Find the variable you're interested in the the smart contract.
    3. Count variables from top to bottom - what's the number of the variable you're interested in? Provide it.
    4. If the variable is less than 32B (e.g. uint256) provide the offset and length to indicate the start of the variable.

    Example for point (4):
        boolean takes up less than 32B of memory, the let's assume the returned value from slot address looks like '0x1000000000000000000000000000000000000000000000000000000000000000'.
        The offset is 3, length is 1.


- What is the slot of the whitelist mapping?: 5
- What is the function to check the whitelist?: bud
0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266

/unified-auctions-ui/core/src/getCollateralPriceOracleConfig.ts:193
            throw new Error('Failed to whitelist the wallet on the fork');
                  ^
Error: Failed to whitelist the wallet on the fork
    at run (/unified-auctions-ui/core/src/getCollateralPriceOracleConfig.ts:193:19)

Comment (KirillDogadin-std) [reply core/src/getCollateralPriceOracleConfig.ts:194]: Aaaah, great job rubber-ducking, @valiafetisov , i knew you're the best helper here.

`npx hardhat --network testnetwork --config local.hardhat.config.ts run ./src/getCollateralPriceOracleConfig.ts` instead of `npx hardhat run ./src/getCollateralPriceOracleConfig.` fixes it

**Review by LukSteib — COMMENTED** [review]
Functional test based on store actions:
* able to fetch vault information (see [1] below)
* able to liquidate vault (see [2] below) 

[1]
image

[2]
image

---
Some questions:
* I would expect any incentives (in DAI) being send to the wallet I used to call `bark()` but there are none. Is this expected?
* I would expect the parameters of the vault to change after it is liquidated (e.g. the collateralAmount) but they don't. Is this expected and parameters effectively only change after the auction ends? 
* I would expect a new auction being displayed via our UI after successfully liquidating the vault but it seems like it doesn't. Expected?

**Comment by KirillDogadin-std** [comment]
> I would expect any incentives (in DAI) being send to the wallet

please check your VAT as well, i think it's there.

> I would expect the parameters of the vault to change after it is liquidated (e.g. the collateralAmount) but they don't. Is this expected and parameters effectively only change after the auction ends?

for now it's not updated because the result of the vault fetching is cached - should i remove the caching for this process, @valiafetisov ?

> I would expect a new auction being displayed via our UI after successfully liquidating the vault but it seems like it doesn't. Expected?

Not expected. How did you validate this though? for me there're pages of eth auctions with different indexes and without the previous point (updating the vault in the store to newest state) it's not easy to tell what's the auction index.

**Comment by LukSteib** [comment]
> please check your VAT as well, i think it's there.

yes, you are right. We might adjust the overall UI flow accordingly. Will take it with me; no need to tackle in this PR.

> Not expected. How did you validate this though? for me there're pages of eth auctions with different indexes and without the previous point (updating the vault in the store to newest state) it's not easy to tell what's the auction index.

Agree. Claim was made on a best effort guess by comparing the collateral amounts of the active ETH auctions with the one of the liquidated vault (where I couldn't find a close enough match). But probably a vague statement.

**Comment by valiafetisov** [comment]
> for now it's not updated because the result of the vault fetching is cached - should i remove the caching for this process, @valiafetisov ?

- [ ] All parameters of the vault _that can be changed_ shouldn't be cached. Not only because it can be liquidated, but also because the owner of the vault may change them (return some DAI or add more collateral) without even touching our UI

- [ ] After the liquidation has happened, the vault should be refetched into the store

File: core/src/getCollateralPriceOracleConfig.ts [hunk 40]
```diff
@@ -0,0 +1,177 @@
...
+    const collateralType = await promptCollateralType();
+    const { contract, address } = await getOracleAddressAndContract(collateralType);
+    console.info(`Contract address: ${address}`);
+    if (basicInfo.hasNextPrice) {
+        const whitelistSlot = 5;
+        const whitelistFunction = 'bud';
+        await addToWhitelist(address, whitelistSlot);
+        const isWhitelisted = await callFunction(contract, whitelistFunction, HARDHAT_PUBLIC_KEY);
+        if (isWhitelisted === '0x00') {
+            throw new Error('Failed to whitelist the wallet on the fork');
```
valiafetisov, line 167: Please provide more logging to make clear what's happening (there is no message that whitelisting even started, no message about which config type is selected, etc). 

Would the developer know what to do when faced with such error message?

core/src/oracles.ts
```diff
@@ -0,0 +1,145 @@
...
+    return nextPriceChange;
+};
+
+const _getOsmPrices = async (
+    network: string,
+    oracleAddress: string,
+    collateralType: CollateralType
+): Promise<OraclePrices> => {
+    const provider = await getProvider(network);
+    const collateralConfig = COLLATERALS[collateralType];
```
valiafetisov, line 125: Please prefer using `getCollateralConfigByType` which automatically throws in case there is no such collateral

core/src/oracles.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const nextPrice = await getNextOraclePrice(collateralConfig.oracle, provider, oracleAddress);
+    const currentPrice = await getCurrentOraclePrice(collateralConfig.oracle, provider, oracleAddress);
+    const nextPriceChange = await getNextOraclePriceChange(collateralConfig, provider, oracleAddress);
+
+    const currentUnitCollateralPrice = new BigNumber(currentPrice).shiftedBy(-WAD_NUMBER_OF_DIGITS);
+    const nextUnitCollateralPrice = new BigNumber(nextPrice).shiftedBy(-WAD_NUMBER_OF_DIGITS);
```
valiafetisov, line 132: - [ ] Instead of doing it here, please make sure that every function in our codebase returns tokens that are properly formatted already (otherwise it's a mess)
- [ ] Please always use correct unit, if you format DAI, use `DAI_NUMBER_OF_DIGITS`

core/src/vaults.ts
```diff
@@ -0,0 +1,361 @@
...
+    const currentProtocolDebtDaiHex = await contract.Dirt();
+    const currentProtocolDebtDai = new BigNumber(currentProtocolDebtDaiHex._hex).shiftedBy(-RAD_NUMBER_OF_DIGITS);
+    return {
+        currentProtocolDebtDai,
+        maximumProtocolDebtDai,
+    };
+};
+
+export const fetchGlobalLiquidationLimits = memoizee(_fetchGlobalLiquidationLimits, {
+    maxAge: CACHE_EXPIRY_MS,
```
valiafetisov, line 98: - [ ] Those values can change more often than in 24 hours. I would say 60 seconds would be enough to avoid refetching of this value with every vault fetch. The only reason we have those caches is when we need to fetch this value in the array of say 100 vaults and we don't want to do 99 extra requests
- [ ] Please also check all other caches

core/src/vaults.ts
```diff
@@ -0,0 +1,361 @@
...
+        ...vaultAmount,
+        ...globalLiquidationLimits,
+        currentCollateralDebtDai,
+        maximumCollateralDebtDai,
+        liquidationPenaltyRatio,
+        minimalAuctionedDai,
+    };
+};
+export const fetchVault = memoizee(_fetchVault, {
+    maxAge: CACHE_EXPIRY_MS,
```
valiafetisov: I don't think this should ever be cached

KirillDogadin-std:
@LukSteib now you're able to make sure that the vault's liquidated in the store and you know which auction id to look for since the logs are provided for the liquidated vault.

core/src/vaults.ts
```diff
@@ -0,0 +1,361 @@
...
+    promise: true,
+    length: 3,
+});
+
+const _fetchLiquidatedParameters = async (network: string, vault: Vault) => {
+    const contract = await getContract(network, 'MCD_DOG');
+    const typeHex = ethers.utils.formatBytes32String(vault.collateralType);
+    const eventFilter = contract.filters.Bark(typeHex, vault.address, null, null, null, null, null);
+    const liquidationEvents = await contract.queryFilter(eventFilter);
+    if (liquidationEvents.length !== 0 && vault.initialDebtDai.eq(0)) {
```
valiafetisov, line 198: Always prefer returning earlier and fetching less

```suggestion
    if (!vault.initialDebtDai.isZero()) {
        return;
    }
    const eventFilter = contract.filters.Bark(typeHex, vault.address, null, null, null, null, null);
    const liquidationEvents = await contract.queryFilter(eventFilter);
    if (liquidationEvents.length === 0) {
        return;
    }
```

core/src/getCollateralPriceOracleConfig.ts
```diff
@@ -0,0 +1,177 @@
...
+const run = async () => {
+    await keypress('Press enter to start collateral config generation');
+    await resetNetworkAndSetupWallet(undefined, HARDHAT_PRIVATE_KEY);
+    const basicInfo = await promptBasicInformation();
+    const config: CollateralPriceSourceConfig = basicInfo.hasNextPrice
+        ? CONFIG_WITH_NEXT_PRICE
+        : CONFIG_WITHOUT_NEXT_PRICE;
+    const collateralType = await promptCollateralType();
+    const { contract, address } = await getOracleAddressAndContract(collateralType);
+    console.info(`Contract address: ${address}`);
```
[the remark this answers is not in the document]
KirillDogadin-std: That's an oracle address, there're 1 per collateral, if i generate the name it's not guaranteed that chainlog has the same one on their side.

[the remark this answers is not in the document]
KirillDogadin-std: I propose

```diff
- console.info(`Contract address: ${address}`);
+ console.info(`Oracle contract address: ${address}`);
```

LukSteib (approved):
Was able to 
* fetch vault and inspect its parameters
* liquidate vault and inspect updated parameters
* see the related auction in the UI

Hence approved from my side ✔️

core/src/vaults.ts
KirillDogadin-std: is this outdated? i struggle to find this line 🥇

core/src/oracles.ts
valiafetisov: I see that you've returned those two lines, but I don't see that they've been moved to the respective functions. Where did they go?

core/src/vaults.ts
```diff
@@ -0,0 +1,335 @@
...
+        id,
+        address,
+        collateralType,
+        lastSyncedAt: new Date(),
+        network,
+    };
+};
+
+export const fetchVaultBase = memoizee(_fetchVaultBase, {
+    maxAge: CACHE_EXPIRY_MS,
```
valiafetisov, line 47: As far as I understand, those values would never change (except for the `lastSyncedAt`), right? So maybe you can remove maxAge here, but move `lastSyncedAt` to the parent?

core/src/vaults.ts
```diff
@@ -0,0 +1,335 @@
...
+
+const _fetchLiquidatedParameters = async (network: string, vault: Vault) => {
+    const contract = await getContract(network, 'MCD_DOG');
+    const typeHex = ethers.utils.formatBytes32String(vault.collateralType);
+    if (!vault.initialDebtDai.isZero()) {
+        return;
+    }
+    const eventFilter = contract.filters.Bark(typeHex, vault.address, null, null, null, null, null);
+    const liquidationEvents = await contract.queryFilter(eventFilter);
+    if (liquidationEvents.length !== 0) {
```
valiafetisov, line 201: Please prefer returning earlier, [as suggested before](https://github.com/sidestream-tech/unified-auctions-ui/pull/457#discussion_r977347070)
```suggestion
    if (liquidationEvents.length === 0) {
        return;
    }
```

core/test/vault-test.ts
```diff
@@ -0,0 +1,381 @@
...
+
+const compareVaultTransactionsNotLiquidated = (
+    expected: VaultTransactionNotLiquidated,
+    actual: VaultTransactionNotLiquidated
+) => {
+    expect(expected.state).to.eq(actual.state);
+    expect(expected.collateralType).to.eq(actual.collateralType);
+    expect(expected.address).to.eq(actual.address);
+    expect(expected.network).to.eq(actual.network);
+    expect(expected.nextPriceChange?.toISOString()).to.eq(actual.nextPriceChange?.toISOString());
```
valiafetisov, line 20: Would [`deep-equal-in-any-order`](https://www.chaijs.com/plugins/deep-equal-in-any-order/) help here?

core/test/vault-test.ts
```diff
@@ -0,0 +1,381 @@
...
+            maximumProtocolDebtDai: new BigNumber('100000000'),
+            currentProtocolDebtDai: new BigNumber(0),
+            currentCollateralDebtDai: new BigNumber(0),
+            maximumCollateralDebtDai: new BigNumber(65000000),
+            transactionFeeLiquidationEth: new BigNumber('0.000669987000446658'),
+            transactionFeeLiquidationDai: new BigNumber('1.130545198449229148108694994096671174'),
+            debtDai: new BigNumber('32198315.88349647029073587039891649505768393987481484'),
+            collateralAmount: new BigNumber('60266.446946772373610407'),
+            currentUnitPrice: new BigNumber('1712.2106886'),
+            nextUnitPrice: new BigNumber('1712.2106886'),
```
valiafetisov, line 89: Does it mean it's a `CONFIG_WITHOUT_NEXT_PRICE`? Or in this case, the value would be `NaN`?

core/test/vault-test.ts
```diff
@@ -0,0 +1,381 @@
...
+        };
+        compareVaultTransactionsNotLiquidated(expectedObject, vaultTransactionLiquidatable);
+    });
+    it('Fetches liquidated vault', async () => {
+        await resetNetwork(14955398);
+
+        const vault = await fetchVault(TEST_NETWORK, 7370);
+        const vaultTransaction = (await getVaultTransaction(TEST_NETWORK, vault)) as VaultTransactionLiquidated;
+        expect(vaultTransaction.state).to.eq('liquidated');
+        expect(vaultTransaction.pastLiquidations[0].auctionId).to.eq('ETH-A:726');
```
valiafetisov, line 100: Can we try to check this auction as well and at least compare the debt and the collateral amount?

core/src/oracles.ts
KirillDogadin-std: the `shiftedBy` now is executing right at the place where the prices are fetched (there're 2 places in this file), there're no additional variable assignments - it's just `return ...shiftedBy()`.

core/test/vault-test.ts
KirillDogadin-std: what you mean by "it"? presense of two prices = `CONFIG_WITH_NEXT_PRICE`

valiafetisov: So this vault have type `CONFIG_WITH_NEXT_PRICE` even if the next price is the same?

core/test/vault-test.ts
```diff
@@ -0,0 +1,390 @@
...
+        );
+        const updatedVault = (await fetchVault(TEST_NETWORK, 27435)) as Vault;
+        expect(updatedVault.collateralAmount.toFixed()).to.eq('0');
+        expect(updatedVault.initialDebtDai.toFixed()).to.eq('0');
+        const updatedVaultTransaction = (await getVaultTransaction(
+            TEST_NETWORK,
+            updatedVault
+        )) as VaultTransactionLiquidated;
+        expect(updatedVaultTransaction.state).to.eq('liquidated');
+        const liquidationAuction = await fetchAuctionByCollateralTypeAndAuctionIndex(TEST_NETWORK, 'ETH-C', 191);
```
valiafetisov, line 170: The idea is to not hardcode values of the auction here, but to validate liquidatedVault parameters: eg `auctionId`

- [ ] Instead of `'ETH-C', 191`, please parse required values from `updatedVaultTransaction` using `parseAuctionId`

core/test/vault-test.ts
```diff
@@ -0,0 +1,390 @@
...
+        expect(liquidationAuction.collateralAmount).to.deep.equalInAnyOrder(expectedObject.collateralAmount);
```
valiafetisov, line 171: Please compare `liquidationAuction` with the value from `vaultTransactionNotLiquidated` for both:
- [ ] `collateralAmount`
- [ ] `debtDai`

KirillDogadin-std: Good point, opened up the thing where the gross profit of liquidation is not taking the liquidation penalty int account

core/src/vaults.ts
valiafetisov: This comment was not addressed

core/test/vault-test.ts
valiafetisov: The first part was not addressed
```suggestion
        expect(liquidationAuction.collateralAmount).to.deep.equalInAnyOrder(vaultTransactionNotLiquidated.collateralAmount);
```

valiafetisov (commented):
- [ ] Do we also test that the combined profit equals the amount of DAI in VAT after the liquidation?

core/test/vault-test.ts
KirillDogadin-std: yes, and that is technically possible.

core/src/vaults.ts
KirillDogadin-std: > parent

to the child you mean? the vault type that inherits from the base which has the data updated regularly.

core/test/vault-test.ts
KirillDogadin-std: there're other fields such as `lastSyncedAt` that are skipped. Also it would heavily influence the debugging process: comparing as it is done rn allows to pre-print the `BigNumber.toFixed()` to spot the exact expected value.

valiafetisov: In general, I expect some kind of helper that utilises `json pointer` to just list which values should be compared can add more clarity here and also help with debugging

core/src/vaults.ts
valiafetisov: No, to the parent! Currently, the `lastSyncedAt` will be cached forever and will be shown on the frontend as `last fetched at`, which will be misleading, since we refetch all other things about the vault. The proposal is to change the VaultBase type to exclude `lastSyncedAt` since it's the only value that is changeable in this function and move this to the parent logic (enrichment?)

We can leave this to tackle in the next PR, since it's a minor thing

valiafetisov (approved):

Issue #363, opened 2022-07-18
Investigate functionality to start collateral auctions


valiafetisov:
### Goal

Get understanding how difficult is it to get all information/rights/etc to start new auctions.

### Context

Starting new auctions or, in maker terms, `barking` on _underwater vaults_ is the functionality that is missing from a complete end-to-end auction flow. Currently, we only provide functionality to facilitate fair market participation in started collateral auctions, and depend on other community members to start auctions. While it is actually profitable by itself and might be required for the newly onboarded collaterals.

### Assets

- `dog.bark` method to liquidate a vault https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/dog.sol#L156-L237
  - `clicp.kick` method to start an auction https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/clip.sol#L220-L266
- Docs https://docs.makerdao.com/smart-contract-modules/dog-and-clipper-detailed-documentation
- Data insights CU and their API https://data.makerdao.network
- Old graphql API for vault analytics https://api.makerdao.com

### Tasks

- [ ] Outline investigation results

LukSteib:
Just wanted to add the following resources here which are nice IMO: 
* https://maker.blockanalitica.com/vaults-at-risk/
  * overview on vaults at risk
* https://maker.blockanalitica.com/simulations/vaults-at-risk/
  * UI to simulate change in OSM price and its effect on vaults at risk

KirillDogadin-std:
> overview on vaults at risk

```sh
curl 'https://papi.blockanalitica.com/maker/vaults-at-risk/' --compressed
```

just sends back the vaults at risk with no auth required, so i guess we could negotiate if it's allowed to borrows this api's functionality in our ui.

The only q becomes wether it only displays vaults at risk or also the ones that are already undercollateralized.

KirillDogadin-std:
__Barking on the auction__

Happens in the following manner:

1. Extracts vault parameters ( stored collateral and the debt) from vat.
2. Make sure that the value stored in the vault is less than the debt, otherwise throws error `Dog/not-unsafe`
3. Check if the trade is possible and overall value that is being auctioned currently does not exceed the limit (which is set in contract)
4. if the value that can still be auctioned is less than the value contained in the vault, partial liquidation is initiated. The check is performed so that the liquidated volume is not to small (regulated by `dust` parameter in collateral)
[...]

PR #461, opened 2022-09-12
Add `VaultLiquidationLimitsCheckPanel` component and stories

### LINKED ISSUE #439: Create `VaultLiquidationLimitsCheckPanel` component
### Goal
Add a new `LiqudationsLimitCheckPanel`, that checks the global liqudations limits. The panel will receive two values, the global and collateral limit. 

@valiafetisov how do we know if the two provided values are above the global limit? 

##### Global limits reached:
Screenshot 2022-08-30 at 13 56 20

##### Global limits not reached:
Screenshot 2022-08-30 at 13 56 58

aomafarag:
Closes #439.

Screenshot from 2022-09-13 14-57-04

Checklist:

### Missing

- Add explanation texts:
  - [x] for numbers
  - [x] for panel


frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,119 @@
...
+            return this.liquidationLimits.maximumProtocolDebtDai
+                .minus(this.liquidationLimits.currentProtocolDebtDai)
+                .minus(this.debtDai);
+        },
+        collateralDifference(): BigNumber {
+            return this.liquidationLimits.maximumCollateralDebtDai
+                .minus(this.liquidationLimits.currentCollateralDebtDai)
+                .minus(this.debtDai);
+        },
+        isGlobalLimitsReached(): Boolean {
```
zoey-kaiser, line 96: I think this name is a bit misleading. I would think from the naming that it only checks for the global Limits. Why don't we rename it to either:

`areLimitsReached` or `areGlobalAndCollateralLimitsReached` to make it a bit more readable..

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,119 @@
...
+            if (this.collateralDifference.isNegative()) {
+                return true;
+            }
+            return false;
+        },
+        currentStateAndTitle(): PanelProps {
+            if (this.isGlobalLimitsReached) {
+                return {
+                    name: 'incorrect',
+                    title: 'Current global liquidation limits are reached',
```
zoey-kaiser, line 109: @valiafetisov do we want to remove the `global` in the panel messages as we have:

1.) Global limits
2.) Collateral limits

I think this would make it easier to understand.

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,119 @@
...
+                };
+            }
+            return {
+                name: 'correct',
+                title: 'Current global liquidation limits are not reached',
```
zoey-kaiser, line 114: same here

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,119 @@
...
+
+<script lang="ts">
+import Vue from 'vue';
+import BigNumber from 'bignumber.js';
+import BaseButton from '~/components/common/inputs/BaseButton.vue';
+import TextBlock from '~/components/common/other/TextBlock.vue';
+import Explain from '~/components/common/other/Explain.vue';
+import FormatCurrency from '~/components/common/formatters/FormatCurrency.vue';
+import BasePanel from '~/components/common/other/BasePanel.vue';
+import { LiquidationLimits } from '~/../core/src/types';
```
zoey-kaiser, line 52: please change this import to:

```suggestion
import { LiquidationLimits } from 'auctions-core/src/types';
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,119 @@
...
+        isGlobalLimitsReached(): Boolean {
+            if (this.globalDifference.isNegative()) {
+                return true;
+            }
+            if (this.collateralDifference.isNegative()) {
+                return true;
+            }
+            return false;
+        },
+        currentStateAndTitle(): PanelProps {
```
zoey-kaiser, line 105: You also need to add a state for when no `LiqudationLimits` are provided. Unless...

We decide to let the user continue with the transaction and MetaMask will inform them when they try to execute. 

What do you think @valiafetisov?

frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
```diff
@@ -0,0 +1,49 @@
+import { storiesOf } from '@storybook/vue';
+import faker from 'faker';
+import BigNumber from 'bignumber.js';
+import VaultLiquidationLimitsCheckPanel from './VaultLiquidationLimitsCheckPanel';
+import { generateFakeVaultNotLiquidatedTransaction } from '~/helpers/generateFakeVault';
+
+const fakeVaultLiquidationTransaction = generateFakeVaultNotLiquidatedTransaction();
```
zoey-kaiser, line 7: Why don't you just generate a fake collateral 

```ts 
import { getAllCollateralTypes } from 'auctions-core/src/constants/COLLATERALS.ts';

const COLLATERALS = getAllCollateralTypes();
faker.helpers.randomize(COLLATERALS)
```


and then export the function to generate mock `LiqudationLimits`? That seems a lot easier than all this.

frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
```diff
@@ -0,0 +1,49 @@
...
+            this.isRefreshing = true;
+            setTimeout(() => {
+                this.isRefreshing = false;
+            }, 1000);
+        },
+    },
+    template: '<VaultLiquidationLimitsCheckPanel v-bind="$data" @refreshLimits="refreshLimits" />',
+};
+
+storiesOf('Panels/VaultLiquidationLimitsCheckPanel', module)
```
zoey-kaiser, line 37: Lets add some more stories:

- `LiqudationLimits` undefined
- `isExplanationsShown` to false
- `isRefreshing` to true

zoey-kaiser (changes requested):

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
valiafetisov: I think the message should be different in case collateral or global limits were reached. In the mock it is just the example for one case

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
valiafetisov: I think they have to be provided according to the types (so that vault without them shouldn't exist, unless it's a liquidated vault). But we can still have `inactive state` for this case just for the stories

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
zoey-kaiser: Agreed! However I would add something after the "please wait". Right now I would be wondering, please wait for what?

My proposal therefore: "..., please wait for them to lower"

valiafetisov: (you can give a priority for the global limit in case both are reached to not handle the `&&` edge case)

valiafetisov: Then let's move the whole `please wait` part to the explanation of this panel

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
valiafetisov: Then let's make the prop `liquidationLimits` optional, but keep the types.ts intact

frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
```diff
@@ -0,0 +1,71 @@
...
+    .add('Expert Mode', () => ({
+        ...common,
+        data() {
+            return {
+                ...common.data(),
+                isExplanationsShown: false,
+            };
+        },
+    }))
+    .add('Global Limits Reached', () => ({
```
valiafetisov, line 45: Can you please add another story for collateral liquidations?

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+        BaseButton,
+        BasePanel,
+        TextBlock,
+        Explain,
+        FormatCurrency,
+    },
+    props: {
+        liquidationLimits: {
+            type: Object as Vue.PropType<LiquidationLimits>,
+            default: undefined,
```
valiafetisov, line 66: Please see the edit https://github.com/sidestream-tech/unified-auctions-ui/pull/461#discussion_r968408806

Actually we expect to receive the whole `VaultTransaction` in the transaction page, so it wouldn't be possible to receive empty `liquidationLimits`, but empty keys of the object

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+        collateralDifference(): BigNumber {
+            return this.liquidationLimits.maximumCollateralDebtDai
+                .minus(this.liquidationLimits.currentCollateralDebtDai)
+                .minus(this.debtDai);
+        },
+        currentStateAndTitle(): PanelProps {
+            if (!this.liquidationLimits) {
+                return {
+                    name: 'inactive',
+                    title: 'No liquidation limits are provided',
```
valiafetisov, line 100: The error messages should always be directed to the user. Eg 
```suggestion
                    title: 'Liquidation limits are unknown',
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+                };
+            }
+            if (this.globalDifference.isNegative()) {
+                return {
+                    name: 'incorrect',
+                    title: 'Current global liquidation limits are reached, please wait for them to lower',
```
valiafetisov, line 106:
```suggestion
                    title: 'Current global liquidation limits are reached',
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+                };
+            }
+            if (this.collateralDifference.isNegative()) {
+                return {
+                    name: 'incorrect',
+                    title: `Current ${this.collateralType} liquidation limits are reached, please wait for them to lower`,
```
valiafetisov, line 112:
```suggestion
                    title: `Current ${this.collateralType} liquidation limits are reached`,
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+        <TextBlock v-if="isExplanationsShown"></TextBlock>
+        <div class="flex justify-between mt-4">
+            <span>Current global limit</span>
+            <span v-if="liquidationLimits"
+                ><Explain :text="liquidationLimits.maximumProtocolDebtDai.toFixed()"></Explain> -
```
valiafetisov, line 8: Please don't use `toFixed` (which will display all possible decimal places) but the appropriate `Format` component 

(here and everywhere where you use `toFixed`)

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+        <TextBlock v-if="isExplanationsShown"></TextBlock>
+        <div class="flex justify-between mt-4">
+            <span>Current global limit</span>
+            <span v-if="liquidationLimits"
```
valiafetisov, line 7: As discussed, this should never be the case, instead, please check each key in a computed prop

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+                ><Explain :text="liquidationLimits.maximumProtocolDebtDai.toFixed()"></Explain> -
+                <Explain :text="liquidationLimits.currentProtocolDebtDai.toFixed()"></Explain> -
```
valiafetisov, line 9: Please provide explanations to the numbers (please propose them if there are none so far). Eg
```
The amount of DAI needed to cover debt and fees of all active auctions. The internal name for this parameter is [dog.Dirt()](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/dog.sol#L78)
```

More info can be found here https://github.com/sidestream-tech/unified-auctions-ui/issues/363#issuecomment-1234396013

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,122 @@
...
+                <Explain :text="debtDai.toFixed()"></Explain> =
```
valiafetisov, line 10: Eg `The amount DAI that will be auctioned after the liquidation`

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,150 @@
...
+            <span>Current {{ collateralType }} limit</span>
+            <span v-if="!isCollateralLimitMissing"
+                ><Explain :text="format(liquidationLimits.maximumCollateralDebtDai)"
+                    >The maximum amount of DAI allowed to cover the debt and fees of all active
+                    {{ collateralType }} auctions</Explain
+                >
+                -
+                <Explain :text="format(liquidationLimits.currentCollateralDebtDai)"
+                    >The amount of DAI needed to cover the debt and fees of all active
+                    {{ collateralType }} auctions</Explain
```
valiafetisov, line 34:
```suggestion
                    >The amount of DAI needed to cover the debt and liquidation insensitives of active
                    {{ collateralType }} auctions. In maker terms it is called <a href="https://docs.makerdao.com/smart-contract-modules/dog-and-clipper-detailed-documentation#limits-on-dai-needed-to-cover-debt-and-fees-of-active-auctions" target="_blank">ilk.dirt</a></Explain
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,150 @@
...
+                >
+                -
+                <Explain :text="format(debtDai)">The amount of DAI that will be auctioned after liquidation</Explain>
```
valiafetisov, line 37:
```suggestion
                <Explain :text="format(debtDai)">The amount of DAI that will be auctioned after the current liquidation plus liquidation insensitives of this auction</Explain>
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,150 @@
...
+        isGlobalLimitMissing(): boolean {
+            return !this.liquidationLimits.maximumProtocolDebtDai || !this.liquidationLimits.currentProtocolDebtDai;
+        },
+        isCollateralLimitMissing(): boolean {
+            return (
+                !this.liquidationLimits.maximumCollateralDebtDai || !this.liquidationLimits.currentCollateralDebtDai
+            );
+        },
+        currentStateAndTitle(): PanelProps {
+            if (this.isGlobalLimitMissing && this.isCollateralLimitMissing) {
```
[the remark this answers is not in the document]
valiafetisov: No, you don't need to. This is a single edge case that should make panel inactive in case ANY of the limits are missing

valiafetisov:
I'm sorry, in all suggestions I've made I mixed up `incentive` with `insensitive`. Please fix that!

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -0,0 +1,196 @@
...
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+        <TextBlock v-if="isExplanationsShown"
+            >The amount of DAI that will be auctioned in case the current vault is liquidated must be less than the
+            currently available DAI to cover the debt and liquidation incentives of all active auuctions in general
+            (global limit) and all active {{ collateralType }} auctions in particular ({{
+                collateralType
+            }}
+            limit).</TextBlock
+        >
```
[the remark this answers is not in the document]
valiafetisov:
```suggestion
        <TextBlock v-if="isExplanationsShown"
            >The limits restrict the amount of simultaneously running auctions. There are two types of limits: global limit
            which restricts maximum total amount of DAI owned by all active auctions; and collateral limit (in this case
            {{ collateralType }}) amount of DAI owned by active auctions of the same collateral type.
            The limits go down when the liquidated vaults are successfully auctioned out in the collateral auctions.
            </TextBlock
        >
```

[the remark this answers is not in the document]
valiafetisov: Yes, you're right. Thanks for reading through it!

valiafetisov (approved):

Issue #375, opened 2022-07-25
Investigate UniV3 automatic routing support


LukSteib:
### Goal
Make use of uni v3 automatic routing feature

### Context
For a lot of the collateral types we are supporting the uniswap v3 callee. However the routes for the token swap are not determined automatically.
Since uniswap provides an automatic routing feature this issue is meant to investigate whether/how we are able to support it and get rid of hardcoding the routes.

### Tasks
- [ ] Investigate how we might support univ3 automatic routing feature

### Assets
* docs: https://docs.uniswap.org/sdk/guides/auto-router

valiafetisov:
Ok, the number of splits can be controlled via `maxSplits`, see [my commit here](https://github.com/sidestream-tech/unified-auctions-ui/pull/467/commits/ce84e3c01e4f14fcec1a27d86fbe8a98901d2798).

LukSteib:
> Ok, the number of splits can be controlled via maxSplits, see [my commit here](https://github.com/sidestream-tech/unified-auctions-ui/pull/467/commits/ce84e3c01e4f14fcec1a27d86fbe8a98901d2798).

What is the final outcome of this investigation? Is there a need for us to take action?

My current interpretation:
* we can use the auto-router to determine the best possible swapping route (ie. moving away from having the hardcoded paths)
  * however we constrain the auto-router suggestion to not split routes
* we can use the auto-router proposed swapping route (e.g. WETH-USDC-DAI) to pass as calldata to exchangeCallee

valiafetisov:
> What is the final outcome of this investigation?

Your summary is correct. It seems that we will be able to both a) get the prices and also b) pass routing information to the callee. Proposed next step is to:
- [ ] integrate `a)` into the dashboard to see if the prices or the routes differ (so it makes sense to integrate it)
- [ ] create concept and the mock to show how to integrate this functionality 
    - the question of how to work with multiply callees per collateral can be relative here (is it user-selectable?)
    - do we replace or make hardcoded routes selectable?
    - how do we display routes?
- [ ] integrate `a)` and `b)` into the actual execution flow

Proposed concept is: automatic routing can be integrated as a "separate route" where we can have 1. Uniswap v3 hardcoded route; 2. Uniswap v3 automatic route with X slippage; 3. Uniswap v3 automatic route with Y slippage. Then, we fetch all prices and propose the most profitable in the UI, but a user can still select another one manually.

[...]

PR #467, opened 2022-09-15
Added UniSwap Autoroutes to dashboard

valiafetisov:
- Addresses https://github.com/sidestream-tech/unified-auctions-ui/issues/375


zoey-kaiser:
I am trying to implement the auto router right now. I have read through everything said in the parent issue and the one about autorouting in general. I feel like I am still missing some important pieces of information that I currently do not understand:

- Currently the input amount is always shifted by the ETH number of digits. How do we shift if for other collateral types/ do we even need to? (Seen [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/026f6d6deb6157ee9777f2ce1cba3edd872cb168/core/src/calleeFunctions/helpers/uniswapAutoRouter.ts#L28))
- The route object is very very big, which keys do I need to specifically add to the table? Based on the issue I would assume: the names of all the objects in under the key `route` (which as far as I see are `V3`, `trade` and a few others) as well as the estimated gas price. Which one of the keys do I use for that? There are so many different options. 
- In my current implementation I also get the error that no network with the name passed from the store can be found, which baffels me as the other functions seem to work. 

I feel like I was sort of thrown into deep water with this issue, as I had no prior infos about it and I could really use some pointers on where to get started!

Also can be please prio #476, as I needed to import the changes into this PR to properly test the dashboard again!

valiafetisov:
> Currently the input amount is always shifted by the ETH number of digits. How do we shift if for other collateral types/ do we even need to?

You can use `getCollateralConfig` to get it's config and use `decimals` to format it properly

> * The route object is very very big, which keys do I need to specifically add to the table? Based on the issue I would assume: the names of all the objects in under the key `route` (which as far as I see are `V3`, `trade` and a few others) as well as the estimated gas price. Which one of the keys do I use for that? There are so many different options.

- `route.quote.toFixed(2)` will print the quote (according to [the docs linked in the issue](https://docs.uniswap.org/sdk/guides/auto-router))
- something like `route.route[0].tokenPath.map(p => p.symbol)`? I imagine since `maxSplits` set to 0, the `route.route` will contain only a single item. Correct me if I'm wrong

> In my current implementation I also get the error that no network with the name passed from the store can be found, which baffels me as the other functions seem to work

Sounds like the exact problem that was fixed in the previously merged PR https://github.com/sidestream-tech/unified-auctions-ui/pull/476 you call your function too early?

zoey-kaiser:
Screenshot 2022-10-06 at 12 35 05

LukSteib (changes requested):
- [ ] let's add a reference to uniswap v3 to the column headings. Proposal: Append existing titles with `(UniV3)`

* Rather a question @valiafetisov: Is this already the effect of exchange fees being factored in here for the auto router that we see. E.g. when comparing the auto route quote for `ETH-A` with the market value I am wondering about the price difference as essentially the same route is being used AFAIK.

image

frontend/store/collaterals.ts
```diff
@@ -98,12 +100,17 @@ export const actions = {
...
+            const autoRouteData = await getUniswapAutoRoute(network, collateral.symbol);
+            const autoRouteExchanges = autoRouteData.route[0].tokenPath.map(p => p.symbol);
+
             const updated = {
                 ilk: collateral.ilk,
                 marketUnitPrice,
                 secondsBetweenPriceDrops: calcParameters.secondsBetweenPriceDrops,
                 priceDropRatio: calcParameters.priceDropRatio,
                 tokenAddress,
+                autoRouteQuote: new BigNumber(autoRouteData.quote.toFixed(2)),
```
valiafetisov, line 112: Why do you turn it into a `BigNumber` in order to turn it back to string with the worse method (`toString`) in the template? Suggestion: keep it as string

frontend/store/collaterals.ts
```diff
@@ -98,12 +100,17 @@ export const actions = {
                     priceDropRatio: error.toString(),
                 };
             });
+            const autoRouteData = await getUniswapAutoRoute(network, collateral.symbol);
+            const autoRouteExchanges = autoRouteData.route[0].tokenPath.map(p => p.symbol);
```
valiafetisov, line 104: Please make this function as well as the extraction of the route safe. In case there is no connection/no token/no tokenPath, etc this whole page should still work. Suggestion: extract required parameters _inside_ `getUniswapAutoRoute`, try/catch it here (see `.catch` all over the place here)

core/src/types.ts
```diff
@@ -63,6 +63,8 @@ export declare interface CollateralRow extends CollateralConfig, Partial<MakerPa
     marketUnitPrice?: BigNumber | string;
     tokenAddress?: string;
     tokenAddressError?: string;
+    autoRouteQuote?: BigNumber;
+    autoRouteExchanges?: string | undefined[];
```
valiafetisov, line 67: Is this what you meant? `?` already specifies that
```suggestion
    autoRouteExchanges?: string[];
```

frontend/store/collaterals.ts
zoey-kaiser: I had done this to be able to determine if the value provided is a valid quote number (BigNumber) or an error message (string).

This way I can add the popup as well to display full error messages.

frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -34,11 +34,17 @@ export default Vue.extend({
         },
     },
     computed: {
+        parsedValue(): number | BigNumber {
+            if (typeof this.value === 'string') {
+                return parseInt(this.value);
+            }
+            return this.value;
+        },
```
valiafetisov, line 42: Why is there changes to the AnimatedNumber in this PR? Seems unrelated to me (and again converts BigNumber into number which we agreed to avoid in the past)

frontend/helpers/generateFakeCollateral.ts
```diff
@@ -13,6 +13,8 @@ export const generateFakeCollaterals = function () {
             secondsBetweenPriceDrops: faker.datatype.number(120),
             priceDropRatio: new BigNumber(faker.datatype.number({ min: 0.5, max: 1, precision: 0.0001 })),
             tokenAddress: faker.finance.ethereumAddress(),
+            autoRouteQuote: faker.finance.amount(),
```
valiafetisov, line 16: Please keep it a `BigNumber` instead
```suggestion
            autoRouteQuote: new BigNumber(faker.finance.amount()),
```

frontend/store/collaterals.ts
```diff
@@ -98,12 +99,24 @@ export const actions = {
...
                 };
             });
+            const autoRouteData = await fetchAutoRouteInformation(network, collateral.symbol).catch(error => {
+                console.error(error);
+                return {
+                    autoRouteError: error.toString(),
+                    autoRouteQuote: undefined,
+                    autoRouteExchanges: undefined,
+                };
+            });
```
valiafetisov, line 109: Nit: You can move `try/catch` into the `fetchAutoRouteInformation` itself, so that the code defining the keys stays in one place (in case you need to change something, you don't need to edit two places)

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
```diff
@@ -0,0 +1,65 @@
...
+    walletAddress?: string
+) {
+    const autoRouteData = await getUniswapAutoRoute(network, collateralSymbol, inputAmount, walletAddress);
+    const autoRouteExchanges = autoRouteData.route[0].tokenPath.map(p => p.symbol);
+
+    return {
+        autoRouteQuote: autoRouteData.quote.toFixed(2),
+        autoRouteExchanges,
+        autoRouteError: undefined,
+    };
```
valiafetisov, line 64: - Please prefer to alway return BigNumber in the core and only format it in the template

Nit:
- No need to prefix keys with the function name
- Use of consistent terms is encouraged `totalPrice`, `route`, `errorMessage`

```suggestion
    return {
        totalPrice: autoRouteData.quote,
        route,
        errorMessage: undefined,
    };
```

valiafetisov (commented):
Code approved (although it's my PR, so I can't properly do that)

zoey-kaiser:
> Code approved (although it's my PR, so I can't properly do that)

Ill do it for you ❤️

zoey-kaiser (approved):

LukSteib:
I want to raise again the question I had some time ago (see: https://github.com/sidestream-tech/unified-auctions-ui/pull/467#issuecomment-1269998042)
Especially for the blue chip collaterals like `ETH` and `WBTC` prices based on auto router are significantly higher than the base case. Do we understand why? 
* Are fees factored in for the prices we get from the auto router? 
* Is it due to the fact that auto router uses different pools? 

One more aspect: For `CRVV1ETHSTETH` we display the error. My assumption is that we are passing `CRVV1ETHSTETH` as the token to be converted to DAI to the autorouter. However as per the [callee contract](https://etherscan.io/address/0x71f2198849F3B1372EA90c079BD634928583f2d2#code) what actually happens on uniswap is the swap from `WETH` -> `DAI` (based on my understanding). Can we circumvent the error by taking this into consideration?

zoey-kaiser:
> I want to raise again the question I had some time ago (see: [#467 (comment)](https://github.com/sidestream-tech/unified-auctions-ui/pull/467#issuecomment-1269998042)) Especially for the blue chip collaterals like `ETH` and `WBTC` prices based on auto router are significantly higher than the base case. Do we understand why?
> 
>     * Are fees factored in for the prices we get from the auto router?
> 
>     * Is it due to the fact that auto router uses different pools?
> 
> 
> One more aspect: For `CRVV1ETHSTETH` we display the error. My assumption is that we are passing `CRVV1ETHSTETH` as the token to be converted to DAI to the autorouter. However as per the [callee contract](https://etherscan.io/address/0x71f2198849F3B1372EA90c079BD634928583f2d2#code) what actually happens on uniswap is the swap from `WETH` -> `DAI` (based on my understanding). Can we circumvent the error by taking this into consideration?

@valiafetisov do you think you could give some insight or give me a starting point? I would not know how to answer the questions at the moment 😢

valiafetisov:
> I would not know how to answer the questions at the moment 😢

Feel free to read the documentation, source code and shoot particular questions here to get started

LukSteib:
> There is [a bunch of other things happens in the callee](https://github.com/makerdao/exchange-callees/blob/3c4055571c0ab17efbd9e7ed809862bc47d8c2fd/src/CurveLpTokenUniv3Callee.sol#L139-L157) before it is exchanged into DAI with UniswapV3, most notably, CURVE_POOL.remove_liquidity_one_coin() is executed to unstake ETH – basically remove ETH from the pool.

I see. And what are we passing to the auto router? The `steCRV` ERC 20 token? Is this correct then? 

> I think those are all good questions, but I'm not sure if they are in the scope of this particular PR.

For me it is fine to put these considerations out of scope of this issue. However, since it will be the determining factor later on wether to use auto router or not we should understand how these prices come about.

LukSteib:
> It seems correct, although [doesn't seem tradable on the uniswap app as well](https://user-images.githubusercontent.com/3393626/196634703-7f57f8a4-38b5-4a4f-8a1a-7e3f647a35a6.png) and is [not verified there](https://user-images.githubusercontent.com/3393626/196636174-3b6a9711-e169-4a79-a8a1-bb50f1f134a6.png) – maybe something related to the proxies as well?

This again refers to my initial concern. The point to use a curve pool as an intermediate step among the chain of things that happen via the callee is to get a ERC 20 token that is exchangable on uniswap in the first place. So from my understanding we would need to pass the address of the token that actually will be exchanged on uniswap. Looking at [this example](https://ethtx.info/mainnet/0x15ec1fd73850b13dccd45d753e91945ead3de88f418eac00e5ecf89ce6ed0bdc/) I think it is [WETH](https://etherscan.io/address/0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2).


> We can merge this PR and directly open an issue to investigate those questions with a clear goal

Agree.

LukSteib:
>  the amount of WETH tokens you receive when unwrapping is not equal to the amount of steCRV tokens you have: this is confirmed by the numbers I've seen while developing the callee.

Then to resolve the current error for `CRVV1ETHSTETH` and make it usable with auto router functionality the question boils down to: How to determine the correct amount of the correct token to be passed to the auto router :) 
But we can pass this discussion to the follow up issue. 


Approved this PR

LukSteib (approved):

valiafetisov:
> make it usable with auto router functionality the question boils down to: How to determine the correct amount of the correct token to be passed to the auto router :)

There are collaterals which require special callees and `CRVV1ETHSTETH` is one of them. Later on when we will add support for auto-routing it will only be added to "simple" collaterals which have `UniswapV3Callee` callee. We might also add it to other callees that use uniswap under the hood (`CurveLpTokenUniv3Callee`, `WstETHCurveUniv3Callee`), but this can be done later in a separate issue.

PR #468, opened 2022-09-21
Minor surplus fixes

### LINKED ISSUE #450: Fix surplus-related minor issues
### Goal

Test surplus participation using simulation, document and fix related problems

### Context

After the new surplus simulation was implemented in https://github.com/sidestream-tech/unified-auctions-ui/pull/449, we're discovered few issues already documented [in the comment there](https://github.com/sidestream-tech/unified-auctions-ui/pull/449#issuecomment-1239647134). Now we want to test it further and fix related problems.

### Tasks

- [ ] Test surplus participation via the UI using new simulation
- [ ] Test surplus participation via the bot using new simulation
- [ ] Please pay attention to various edge cases (eg testing participation if user is the latest bidder and if not the latest bidder)
- [ ] Document found problems, add them in the comments here
- [ ] Fix found issues, inclusing
  - [x] Fix uniswap price fetching when there is no bids
  - [x] Set `nextMinimumBid` to make sense (eg a uniswap price * gap to reach 1k net profit)
  - [x] `SurplusAuctionBidTransactionTable` "state" should display state, not just date
  - [x] `SurplusAuctionBidTransactionTable` end state should count down, but count up (ie Ended 2s ago, 3s ago, etc. First guess: can be caused by invalid determination of the `earliestEndDate`)

aomafarag:
Closes #450.

Checklist:

Issues to be fixed:
- [x] Fix uniswap price fetching when there is no bids
- [x] Set `nextMinimumBid` to make sense (eg a uniswap price * gap to reach 1k net profit)
- [x] `SurplusAuctionBidTransactionTable` "state" should display state, not just date
- [x] `SurplusAuctionBidTransactionTable` end state should count down, but count up (ie Ended 2s ago, 3s ago, etc. First guess: can be caused by invalid determination of the `earliestEndDate`)

### DIFF
--- a/core/package.json
+++ b/core/package.json
@@ -9,7 +9,7 @@
     "start": "node .",
     "test": "npx hardhat test --network testnetwork --config local.hardhat.config.ts",
     "lint": "eslint --ext \".js,.ts\" --ignore-path .gitignore . --max-warnings=0",
-    "hardhat:silent": "npx hardhat node 1>&/dev/null",
+    "hardhat:silent": "npx hardhat node &> /dev/null",
     "hardhat": "npx hardhat node",
     "hardhat:simulations": "npx hardhat --network testnetwork --config local.hardhat.config.ts run ./simulations/index.ts",
     "simulate": "npm-run-all --parallel hardhat:silent hardhat:simulations"
--- a/core/src/surplus.ts
+++ b/core/src/surplus.ts
@@ -205,11 +205,14 @@ export const enrichSurplusAuction = async (
     network: string,
     auction: SurplusAuctionActive
 ): Promise<SurplusAuctionTransaction> => {
-    const nextMinimumBid = await getNextMinimumBid(network, auction);
+    let nextMinimumBid = await getNextMinimumBid(network, auction);
     const unitPrice = auction.bidAmountMKR.div(auction.receiveAmountDAI);
     const marketUnitPrice = await getMarketPriceMkr(network, auction.bidAmountMKR);
     const marketUnitPriceToUnitPriceRatio = unitPrice.minus(marketUnitPrice).dividedBy(marketUnitPrice);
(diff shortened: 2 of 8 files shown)


valiafetisov:
> Set `nextMinimumBid` to make sense (eg a uniswap price * gap to reach 1k net profit)

Why was it not addressed in this PR? I understand that it's not the UI problem, but you basically need to fix `nextMinimumBid` in the core to set a valid default instead of 0

valiafetisov:
Initially, there is no `bidAmountMKR` as there is no bids yet. But you only need to check if `nextMinimumBid` is zero to set it to another default, since 0 is not a valid bid amount

valiafetisov (approved):

LukSteib (changes requested):
- [ ] For me uniswap price still not fetched properly (see 1 below)


1
image

valiafetisov:
Not sure I fully understand without a link to the source code where the problem is. What is the proposed solution?

valiafetisov:
> the auctions are not enriched in that first case

Ok, the problem is reproduced. Do you want to investigate why is it not enriched in this case and propose a solution?

valiafetisov:
Since we want to show "unit price" and not the total price for the whole collateral, we can (only in case when there is no bids yet) compute `marketUnitPrice` using arbitrary number, eg `1`. The price will be different in case the bid amount will be significantly bigger, but I think it's better to show at least some kind of unit price as a reference than no unit price at all. Since then the user will still need to go to external platform to get it. Do you agree, @LukSteib?

LukSteib:
> Do you agree, @LukSteib

Agree to the propsal by @valiafetisov. Better to use arbitrary number (in this case `1`) than none.

valiafetisov (approved):

LukSteib (approved):

PR #469, opened 2022-09-22
Added `VaultsFlow` and `VaultContainer`

### LINKED ISSUE #433: Add `vaults` frontend components
### Goal
In #419 we made a mock for the frontend for the new `vaults` page. This issue aims to provide a clean overview of any new components we need as well as any old components we can reuse for this new feature.

I believe we should wait to start making the components until we decided on what the types should look like. I think this will make the development a lot easier as most components require the VaultTypes to function properly. 

<details><summary>Finished Mock</summary>
<p>

MakerDAO vaults

</p>
</details>

### Components Overview

- [ ] VaultsFlow
- [x] Vaults
  - [x] #436
  - [x] #434
- [x] Vault Page
  - [x] #435
  - [x] VaultEventsTable
  - [x] VaultLiquidatedFlow
- [x] VaultLiquidationTransactionFlow
  - [x] #437
  - [x] WalletConnectPanel (already exists)
  - [x] #439
  - [x] #440
- [x] #441 
- [x] #442 

We are currently still writing all the required texts. You can find the current draft [here](https://docs.google.com/document/d/1Ae-xIj5fR53z0VHCK0sObITKOyO2IULk4g1-a3sKAc8/edit?usp=sharing).

zoey-kaiser:
Closes #433 

Still blocked by #457 

Checklist:


LukSteib (changes requested):
- [ ] We currently use the heading `Collateral auctions` in the head. Let's change it to `Vault liquidation`

zoey-kaiser:
@valiafetisov and @KirillDogadin-std. Lukas and I did some testing, these are the current issues we face!

Frontend Issues:

- Component type errors for some types that were changed from number to BigNumber
- When navigating from panel step 1 to step 0, the URL does not get reset
- On step 2, the proximity to Liquidation needs to be formatted
- When loading into a vault via URL, the vault does not get fetched correctly (error message: network is undefined) 

Issues with Core:
- Liquidation proximity for vault #28187 should be 33% but is displayed as 0.33%

KirillDogadin-std:
> Liquidation proximity for vault #28187 should be 33% but is displayed as 0.33%

yeah, i guess that's a matter of preference and implementations:

 - it's named `proximityToLiquidation` and not `proximityToLiquidationPercent`
 - reason to store it as value between 1 and 0: no conversion will be needed if this value is used in the mathematical operations.

Possible solutions:

 a. add the helper to frontend `formatFloatToPercent`
 b. adjust the core logic so that it actually goes and provides the value in percents right away.

for the above stated reason i would go with (a) option. Lmk if it's too unagreeable.

valiafetisov:
> a. add the helper to frontend

There is already a helper for that called `FormatPercentage`

> I cannot build the core right now so I am coding blind, which makes debugging super hard

@zoey-kaiser was it resolved? Or what is the problem exactly? Please provide logs + reproduction instructions

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -75,6 +75,7 @@
                 <span>DAI</span>
             </div>
         </div>
+        <!--
```
valiafetisov, line 78: Technically, you can trigger refetching of the vault, which will also get fresh limits

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -146,6 +142,16 @@ export default Vue.extend({
         wasLiquidated(): boolean {
             return this.vaultTransaction.state === 'liquidated';
         },
+        liquidationLimits(): LiquidationLimits {
+            return {
```
valiafetisov, line 146: I think it's possible to pass a type which has more properties than needed. So you can pass the whole `VaultTransaction` to the prop that asks for `LiquidationLimits`, no need to hand-pick certain keys

frontend/store/vaults.ts
```diff
@@ -50,13 +59,21 @@ export const actions = {
...
         }
     },
-    async liquidateVault({ rootGetters, commit, getters, dispatch }: ActionContext<State, State>, vaultId: number) {
+    async liquidateVault(
+        { rootGetters, commit, getters, dispatch }: ActionContext<State, State>,
+        { vaultId, walletAddress }: { vaultId: number; walletAddress?: string }
+    ) {
         const network = rootGetters['network/getMakerNetwork'];
-        const walletAddress = rootGetters['wallet/getAddress'];
+        const executeWalletAddress = walletAddress || rootGetters['wallet/getAddress'];
```
valiafetisov, line 67: Actually, you don't need to get wallet address from the store, `rootGetters['wallet/getAddress']` because `liquidateVault` core function doesn't require `walletAddress` argument and will substitute with the signer vault address (which is better in case we have store-related problems)

frontend/components/vault/VaultFlow.vue
```diff
@@ -0,0 +1,170 @@
...
+    methods: {
+        explanationsTrigger(event: boolean): void {
+            if (event === true && this.$refs.mainText) {
+                (this.$refs.mainText as Vue).$el.scrollIntoView({ block: 'start', behavior: 'smooth' });
+            }
+            this.$emit('update:isExplanationsShown', event);
+        },
+        liquidate() {
+            this.step = 2;
+        },
```
valiafetisov, line 152: The name of this method is misleading (as well as the emitted event name) Let's find a different name for it (eg `participate`/`openLiquidationTransaction`)

zoey-kaiser:
- Notifier when liquidating
- Dark mode input field
- When switching from a liquidated vault to newly created auction, re-fetch auction as soon as you navigate there
- fix helper functions
- Recheck profit calculation: In tests we calculated 212,8 but it was 310,14
- CRUVE Vaults are not found (expected tho)

LukSteib (approved):

KirillDogadin-std (approved):

PR #476, opened 2022-09-26
Fix token information not showing on dashboard

### LINKED ISSUE #472: Fix token information on dashboard
### Goal
Dashboard page displays token information as expected

### Context
Currently the dashboard page https://unified-auctions.makerdao.com/dashboard doesn't seem to work properly. Token information is stuck in an infinite loading loop. 
Let's investigate and fix

### Assets
image

### Tasks
- [ ] Fix dashboard related problem as described above

aomafarag:
Closes #472.

Screenshot from 2022-09-26 13-20-23

Checklist:

### DIFF
--- a/frontend/components/dashboard/CollateralTable.vue
+++ b/frontend/components/dashboard/CollateralTable.vue
@@ -14,6 +14,20 @@
             {{ ilk }}
         </div>
         <div slot="symbol" slot-scope="symbol" class="Element">{{ symbol }}</div>
+        <div slot="token" slot-scope="tokenAddress, record" class="Element" :class="{ Loading: isLoading(record) }">
+            <div v-if="isLoading(record)" class="flex items-center">
+                <LoadingIcon class="h-3 w-3 animate animate-spin fill-current dark:text-gray-300 mr-2" />
+                <span>Loading...</span>
+            </div>
+            <div v-else>
+                <format-address v-if="tokenAddress" :value="tokenAddress" shorten type="address" />
+                <Popover v-else placement="top" :content="record.tokenAddressError" trigger="hover">
+                    <p class="inline-block w-20 text-red-500 truncate">
+                        <span>{{ record.tokenAddressError }}</span>
+                    </p>
+                </Popover>
+            </div>
+        </div>
         <div
             slot="marketUnitPrice"
             slot-scope="marketUnitPrice, record"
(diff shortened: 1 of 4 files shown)


valiafetisov:
> missing a function for refetching data

The dashboard doesn't need to be refetched, it's just a debugging page.

- Then how refetching solves the problem of initial loading?
- When does the `setup` called initially? Is it called on the dashboard page load or on any page load? – if on any, then it should be fixed and called only on the dashboard page load

>  Also, are the errors in the screenshot (Error: missing revert data in...) normal?

Depends on the network. On the mainnet all collaterals should have uniswap price

valiafetisov:
> same goes for virtually all stores

Yes, because we expect that values on the other stores might change. The dashboard _mostly_ presents things that doesn't change that often – list of collaterals and their addresses, prices (that we only need to know are tradable or not).

> When the dashboard container is mounted

Then why it doesn't get displayed anymore?

valiafetisov:
> once the page is refreshed

What do you mean by `refreshed`? When it `is lost` is it lost from the store (probably `reset`?) or only from the component?

frontend/components/dashboard/CollateralTable.vue
```diff
@@ -60,23 +74,6 @@
                 %
             </span>
         </div>
-        <div slot="icon" slot-scope="record" class="Element">
-            <CurrencyIcon :currency-symbol="record.symbol" />
-        </div>
```
valiafetisov, line 65: Why did you need to remove `CurrencyIcon` and move `token address` column around to fix the problem? Is it unrelated improvement or a leftover?

frontend/containers/DashboardContainer.vue
```diff
@@ -21,6 +21,12 @@ export default Vue.extend({
     components: {
         DashboardAuctionsView,
     },
+    props: {
+        network: {
+            type: String,
+            default: undefined,
+        },
+    },
```
valiafetisov, line 29: You actually don't need to pass it as a prop, since it's already a container and can use `...mapGetters(network, { network: 'getMakerNetwork' })` directly

frontend/components/dashboard/CollateralTable.vue
valiafetisov: Alright! Did it also change the order of columns in the UI or had no effect on it?

valiafetisov (approved):

LukSteib (approved):

PR #479, opened 2022-09-27
Simulation Vault liquidation

### LINKED ISSUE #464: Create a simulation for vault liquidation
### Goal

A developer is able to start a simulation which will create a vault and turn it into `liquidatable` state.

### Context

In order to test the new collateral types, but also vault logic, we need to create a reusable simulation steps in order to test related functionality. For this, let's implement a creation of a new vault and make this logic universal across collaterals. Eg:
- Create a vault via `createActiveVault(collateralType)` test logic
  - get tokens of a specific collateral symbol (via overwriting a mapping?)
  - set allowance, collateral auth
  - open vault, deposit collateral
  - ? withdraw all available DAI
- Make vault `liquidatable` by:
  - ? `drip`, ie `collectProtocolFees(collateralType)`
  - or setting OSM price (I guess it's more difficult to make it universal?)

Note:
- This can be implemented as part of the https://github.com/sidestream-tech/unified-auctions-ui/issues/460 or after it, in order to test vault creation

### Tasks

- [x] Come of with the final proposal on how to make it universal across collateral types 
- [x] Implement simulation

KirillDogadin-std:
Closes #464

Checklist:


core/simulations/configs/vaultLiquidation.ts
```diff
@@ -0,0 +1,137 @@
...
+            title: 'Deposit Collateral to vault',
+            entry: async () => {
+                const latestVaultId = await getLatestVaultIndex();
+                const vault = await fetchVaultBase(TEST_NETWORK, latestVaultId);
+                await depositCollateralToVat(
+                    TEST_NETWORK,
+                    vault.address,
+                    vault.collateralType,
+                    COLLATARAL_OWNED.dividedBy(1.2)
+                );
```
KirillDogadin-std, line 105: advice requested @valiafetisov 

This step fails with `GemJoin/failed-transfer` that corresponds to failed transfer in the [join.sol](https://github.com/makerdao/dss/blob/master/src/join.sol) the allowance has been given out in the previous step.

Vat operation (generating the collateral in vat) comes before transfer and hence does not fail.

I'm short on ideas what is going wrong here

KirillDogadin-std: - i have this in the wallet
- __I__ have the amount in the VAT initially, but i do `join` so that it ends up in the __vault's__ VAT.
- i mean: possible to bypass this by adding the vat directly into the vault's VAT, but then it's not a 100% realistic simulation

core/simulations/configs/vaultLiquidation.ts
```diff
@@ -0,0 +1,137 @@
...
+            entry: async () => {
+                const collateralType = await getCollateralType();
+                const addressJoin = await getContractAddressByName(
+                    TEST_NETWORK,
+                    getJoinNameByCollateralType(collateralType)
+                );
+                const collateralConfig = getCollateralConfigByType(collateralType);
+                const tokenContractAddress = await getContractAddressByName(TEST_NETWORK, collateralConfig.symbol);
+                const contract = await getErc20Contract(TEST_NETWORK, tokenContractAddress, true);
+                await contract.approve(addressJoin, COLLATARAL_OWNED.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0));
```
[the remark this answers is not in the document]
KirillDogadin-std: this is a valid remark, but wether i adjust it depends on your answer to the other comment. (this line can just disappear)

core/simulations/configs/vaultLiquidation.ts
valiafetisov: Sorry, I just skimmed it though last time and didn't get the whole flow, now I get it. My suggestion to move forward: to validate every step (that money is in the account, then that allowance was given, etc) and also try to authorize. After I given the authorization, I received a different type of error (but on the same simulation step):

<details><summary>SafeMath: subtraction overflow error</summary>

```
✔ Press Enter to proceed to the next step "Deposit Collateral to vault" … 
transaction 3d793d28-2a28-4281-8e7e-cc9c286acbf2: Transaction is preparing to be mined...
transaction 3d793d28-2a28-4281-8e7e-cc9c286acbf2: Transaction error: "Error: VM Exception while processing transaction: reverted with reason string SafeMath: subtraction overflow (cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ])"

/Users/user/Projects/sidestream/unified-auctions-ui/core/node_modules/@ethersproject/logger/src.ts/index.ts:261
        const error: any = new Error(message);
                           ^
Error: cannot estimate gas; transaction may fail or may require manual gas limit [ See: https://links.ethers.org/v5-errors-UNPREDICTABLE_GAS_LIMIT ] (error={"reason":"Error: VM Exception while processing transaction: reverted with reason string 'SafeMath: subtraction overflow'","code":"UNPREDICTABLE_GAS_LIMIT","method":"estimateGas","transaction":{"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"maxFeePerGas":{"type":"BigNumber","hex":"0x59682f02"},"to":"0x24e459F61cEAa7b1cE70Dbaea938940A7c5aD46e","data":"0x3b4da69f000000000000000000000000bcca2b4c57ac039f14e2f125e3bfaffbcda6356600000000000000000000000000000004e5dc52f5e4c61ff1f453902aaaaaaaaa","type":2,"accessList":null},"error":{"reason":"processing response error","code":"SERVER_ERROR","body":"{\"jsonrpc\":\"2.0\",\"id\":60,\"error\":{\"code\":-32603,\"message\":\"Error: VM Exception while processing transaction: reverted with reason string 'SafeMath: subtraction overflow'\",\"data\":{\"message\":\"Error: VM Exception while processing transaction: reverted with reason string 'SafeMath: subtraction overflow'\",\"data\":\"0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001e536166654d6174683a207375627472616374696f6e206f766572666c6f770000\"}}}","error":{"code":-32603,"data":{"message":"Error: VM Exception while processing transaction: reverted with reason string 'SafeMath: subtraction overflow'","data":"0x08c379a00000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001e536166654d6174683a207375627472616374696f6e206f766572666c6f770000"}},"requestBody":"{\"method\":\"eth_estimateGas\",\"params\":[{\"type\":\"0x2\",\"maxFeePerGas\":\"0x59682f02\",\"maxPriorityFeePerGas\":\"0x59682f00\",\"from\":\"0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266\",\"to\":\"0x24e459f61ceaa7b1ce70dbaea938940a7c5ad46e\",\"data\":\"0x3b4da69f000000000000000000000000bcca2b4c57ac039f14e2f125e3bfaffbcda6356600000000000000000000000000000004e5dc52f5e4c61ff1f453902aaaaaaaaa\"}],\"id\":60,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","url":"http://127.0.0.1:8545"}}, tx={"data":"0x3b4da69f000000000000000000000000bcca2b4c57ac039f14e2f125e3bfaffbcda6356600000000000000000000000000000004e5dc52f5e4c61ff1f453902aaaaaaaaa","to":{},"maxFeePerGas":{"type":"BigNumber","hex":"0x59682f02"},"maxPriorityFeePerGas":{"type":"BigNumber","hex":"0x59682f00"},"from":"0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266","type":2,"nonce":{},"gasLimit":{},"chainId":{}}, code=UNPREDICTABLE_GAS_LIMIT, version=abstract-signer/5.6.2)
    at Logger.makeError (/Users/user/Projects/sidestream/unified-auctions-ui/core/node_modules/@ethersproject/logger/src.ts/index.ts:261:28)
    at Logger.throwError (/Users/user/Projects/sidestream/unified-auctions-ui/core/node_modules/@ethersproject/logger/src.ts/index.ts:273:20)
    at /Users/user/Projects/sidestream/unified-auctions-ui/core/node_modules/@ethersproject/abstract-signer/src.ts/index.ts:301:31
    at async Promise.all (index 7)
```
</details>

KirillDogadin-std: >  My suggestion to move forward: to validate every step

👍🏻 

> I given the authorization to this collateral

So, reading our code i see that we have 3 authorizations: 
  - to the clipper,
  - to the flapper,
  - to the flopper.

each of those is responsible to manage the auctions.

I have 2 questions:
  - which function did you call via ui? (i need it in order to find out which contract got your auth at the end)
  - did you guess that the auth is needed here or do you have a specific place to reference that shows that it's needed?

valiafetisov: > did you guess that the auth is needed here or do you have a specific place to reference that shows that it's needed?

I just called whatever auth we have on the frontend, I think it's `MCD_VAT.hope(clipperAddress)` aka `authorizeCollateral`. We require this authorisation to withdraw collateral from VAT in `Bidding with DAI` flow. Maybe you need to do it for the vault-related contracts instead? I can't see any auth in the code, but I had this in the past that it's hard to trace it.

I also suggested to fetch VAT balance, balance and allowance once you change them to validate they hold expected values, since it's nice to debug everything step-by-step and make sure that the current step doesn't fail because the previous didn't actually work

valiafetisov: In [dai.js source](https://github.com/makerdao/dai.js/blob/76f7ddc842d48e0eebdb4368f84e693509468cac/packages/dai-plugin-mcd/src/CdpManager.ts#L138-L179) they are calling quite a different method, if I understand correctly. If it's that complicated, we can directly modify vault amount if it doesn't break any further logic and let us liquidate the vault in the end

KirillDogadin-std: Ok, turns out that indeed auth is not required here. Fixed the problem. The bad part was that i messed up the units. (WAD shifts)

valiafetisov:
> i would then prompt the amount of collateral in the beginning and print the limits

Let's automatically pick a random number between the limits

> how do you like the approach with simulation steps calling functions to get some "global parameter" that should be accessible to all steps?

It's fine, although, you can improve simulation loop to pass returned value from one step to the next one (or even better, always pass `context` object), eg: 
```ts
{
    title: 'Select the collateral',
    entry: async () => {
        await resetNetworkAndSetupWallet(14052140);
        const vaultIndex = await getLatestVaultIndex.clear();
        const collateralType = await getCollateralType.clear();
        return { vaultIndex, collateralType }
    },
},
{
    title: 'Set collateral in VAT',
    entry: async (context: any) => {
        await setCollateralInVat(context.collateralType, COLLATARAL_OWNED_WAD);
        await assertVatCollateralBalance(
            TEST_NETWORK,
            HARDHAT_PUBLIC_KEY,
            context.collateralType,
            COLLATARAL_OWNED_WAD
        );
    },
},
```
Or just use global vars in the file (not nice since they will not be zeroed between runs)

core/helpers/assertions.ts
```diff
@@ -0,0 +1,67 @@
+import BigNumber from '../src/bignumber';
+import getContract, { getErc20Contract } from '../src/contracts';
+import { CollateralType } from '../src/types';
+import { ethers } from 'ethers';
+import { WAD_NUMBER_OF_DIGITS } from '../src/constants/UNITS';
+
+export const assertVatCollateralBalance = async (
```
valiafetisov, line 7: Let's not create a new `assertions` file with a single-purpose functions. Instead, we can:
- create functions (if they not already exist) that fetch and return what we need
- do assertion directly in the function that does modification (or in the test)

valiafetisov:
Would this script work for all collateral types including `CRVV1ETHSTETH`? Otherwise, what collaterals doesn't work? Do you want to do a test that will check each?

core/helpers/assertions.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (allowance.toFixed(0) !== expectedAllowance.toFixed(0)) {
+        throw new Error(
+            `Unexpected allowance: Expected ${expectedAllowance.toFixed()}, actual: ${allowance.toFixed()}`
+        );
+    }
+    console.info(`Allowance of ${authenticatedAddress} from ${walletAddress} is ${expectedAllowance.toFixed()} WAD`);
+};
+
+export const assertAuth = async (network: string, accessorAddress: string, accessedAddress: string) => {
+    const vat = await getContract(network, 'MCD_VAT');
```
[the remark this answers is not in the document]
KirillDogadin-std: for specific cases only. will add the generic function and reuse it.

core/helpers/assertions.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+export const assertBalance = async (
+    network: string,
+    erc20TokenAddress: string,
+    walletAddress: string,
+    expectedBalance: BigNumber
+) => {
+    const token = await getErc20Contract(network, erc20TokenAddress);
+    const balanceHex = await token.balanceOf(walletAddress);
```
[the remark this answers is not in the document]
KirillDogadin-std: added one

core/simulations/configs/vaultLiquidation.ts
```diff
@@ -0,0 +1,167 @@
...
+                getLatestVaultIndex.clear();
+                getCollateralType.clear();
+            },
+        },
+        {
+            title: 'Set collateral in VAT',
+            entry: async () => {
+                const collateralType = await getCollateralType();
+                await setCollateralInVat(collateralType, COLLATARAL_OWNED_WAD);
+                await assertVatCollateralBalance(
```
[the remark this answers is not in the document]
KirillDogadin-std: this comment would be clear if it wouldn't be attached to some codelines. Instead it is and thus is a bit confusing.

I assume you're suggesting just generally reducing the amount of steps and putting the whole setup into a single function.

KirillDogadin-std:
> Would this script work for all collateral types including CRVV1ETHSTETH? Otherwise, what collaterals doesn't work? Do you want to do a test that will check each?

`CRVV1ETHSTETH` - no, disabled it here

> Otherwise, what collaterals doesn't work?
  - the ones that are not having the funds in join contracts to give collateral out. Did not overwrite the balance because i don't know how to do it reliably (ERC20 contracts still could have different memory layouts and setting the balance is tricky there. If needed, let's do it in the other issue).
  - the ones that are artificially disabled by setting "maximum possible debt" to 0
Corresponding errors are introduced

core/helpers/hex.ts
```diff
@@ -11,3 +12,9 @@ export const concat = (prefix: string, postfix: string) => {
 export const stripZeros = (val: string) => {
     return ethers.utils.hexStripZeros(val);
 };
+
+export const randomBigNumber = (min: BigNumber, max: BigNumber) => {
+    const generatedBytes = ethers.utils.hexlify(ethers.utils.randomBytes(32));
+    const offset = new BigNumber(generatedBytes).mod(max.minus(min));
+    return offset.plus(min);
```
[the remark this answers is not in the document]
valiafetisov: Also, I would call this function `randomBigNumberBetween` and I'm not sure if it has something to do with `hex`, so let's move it somewhere else (you can have a new helper for calculations)

core/simulations/steps/createVaultForCollateral.ts
```diff
@@ -0,0 +1,160 @@
...
+    const addressJoin = await getContractAddressByName(TEST_NETWORK, getJoinNameByCollateralType(collateralType));
+    const joinContractHasSufficientFunds = await isBalanceGreaterThan(
+        TEST_NETWORK,
+        tokenContractAddress,
+        addressJoin,
+        collateralOwned
+    );
+    if (!joinContractHasSufficientFunds) {
+        throw new Error('Join contract does not have sufficient funds');
+    }
```
valiafetisov, line 55: Can you please explain why is this error again? Can't we just add more tokens to the join contract? Or this will effectively mean that we need to modify every token contract, which we tried to avoid?

core/simulations/steps/createVaultForCollateral.ts
```diff
@@ -0,0 +1,160 @@
...
+    const contract = await getContract(TEST_NETWORK, 'MCD_VAT');
+    const typeHex = ethers.utils.formatBytes32String(collateralType);
+    const { dust } = await contract.ilks(typeHex);
+    const { minUnitPrice } = await fetchVaultCollateralParameters(TEST_NETWORK, collateralType);
+    const minCollateralInVault = new BigNumber(new BigNumber(dust._hex).toFixed(0))
+        .shiftedBy(-RAD_NUMBER_OF_DIGITS)
+        .div(minUnitPrice);
+
+    await checkAvailableDebtForAmountAndMinUnitPrice(collateralType, minCollateralInVault, minUnitPrice);
+    return new BigNumber(minCollateralInVault.multipliedBy(1.1).toFixed(0));
```
valiafetisov, line 93: Why is it multiplied by `1.1`?

core/src/vaults.ts
```diff
@@ -333,3 +333,40 @@ export const liquidateVault = async (
...
+        differenceDebtDai.shiftedBy(DAI_NUMBER_OF_DIGITS).toFixed(0),
+    ];
+    await executeTransaction(network, 'CDP_MANAGER', 'frob', argumentList);
+};
+
+export const collectStabilityFees = async (network: string, collateralType: CollateralType) => {
+    return await executeTransaction(network, 'MCD_JUG', 'drip', [ethers.utils.formatBytes32String(collateralType)]);
+};
+
+export const changeCollateralInVault = async (
```
valiafetisov, line 360: Nit:
```suggestion
export const withdrawCollateralFromVault = async (
```

core/test/vault-test.ts
```diff
@@ -407,3 +409,33 @@ describe('Sound values are extracted', () => {
...
+            } catch (e) {
+                if (e instanceof Error && e.message.startsWith('Join contract does not have sufficient funds')) {
+                    return;
+                }
+                throw e;
+            }
+
+            const vault = await fetchVault(TEST_NETWORK, vaultId);
+            expect(vault.collateralAmount.toFixed(0)).to.eq(collateralOwned.toFixed(0));
+        });
```
valiafetisov, line 439: Can you please also liquidate it here?

valiafetisov (commented):
- [x] Nit: please add one empty line between functions

core/simulations/steps/createVaultForCollateral.ts
KirillDogadin-std: >  Can't we just add more tokens to the join contract? 

yes, this will effectively mean that we need to modify every token contract so it would possess enough funds to send to the user's wallet.

core/simulations/steps/createVaultForCollateral.ts
KirillDogadin-std: to not be the minimal. But it could. removed.

core/src/vaults.ts
KirillDogadin-std: nope, not true. You can send in negative or positive number and either put or extract collateral.

core/simulations/steps/createVaultForCollateral.ts
valiafetisov: That's actually a problem for testing newly onboarded collateral which have no vaults. We might actually need to find a way to transfer the tokens into the join address anyway 🙁 (but not in this PR already)

core/test/vault-test.ts
```diff
@@ -407,3 +416,38 @@ describe('Sound values are extracted', () => {
...
+            try {
+                collateralOwned = await getCollateralAmountInVat(collateralType);
+            } catch (e) {
+                if (e instanceof Error && e.message.startsWith('Cannot borrow more dai with the collateral')) {
+                    return;
+                }
+                throw e;
+            }
+            try {
+                vaultId = await createVaultForCollateral(collateralType, collateralOwned, collateralConfig.decimals);
```
valiafetisov, line 436: Instead of passing it as additional arg it's enough to have `collateralType` arg to determine amount of decimals using `getCollateralConfigByType`

core/simulations/configs/vaultLiquidation.ts
```diff
@@ -0,0 +1,112 @@
...
+import { HARDHAT_PUBLIC_KEY, TEST_NETWORK } from '../../helpers/constants';
+import createVaultForCollateral, { getCollateralAmountInVat } from '../steps/createVaultForCollateral';
+
+const getSupportedCollaterals = () => {
+    const supportedCollaterals: Record<string, CollateralConfig> = JSON.parse(JSON.stringify(COLLATERALS));
+    delete supportedCollaterals['CRVV1ETHSTETH-A'];
+    return supportedCollaterals;
+};
+
+export const SUPPORTED_COLLATERALS = getSupportedCollaterals();
```
valiafetisov, line 17: It's a bad practice to assign something to a variable and then export it. I suggest that you export `getLiquidatableCollateralTypes` function that returns list of types and is used in other files
```suggestion
const UNSUPPORTED_COLLATERAL_TYPES = ['CRVV1ETHSTETH-A'];

export const getLiquidatableCollateralTypes = () => {
    return getAllCollateralTypes().filter(collateralType => !UNSUPPORTED_COLLATERAL_TYPES.includes(collateralType));
};
```

valiafetisov (commented):
- [x] As discussed, please document in the comment the list of the collaterals that currently throw an error during the test execution

KirillDogadin-std:
running grep on logs from tests
```
REPORT: AAVE-A - Can't borrow dai
REPORT: BAL-A - Can't borrow dai
REPORT: BAT-A - Can't borrow dai
REPORT: COMP-A - Can't borrow dai
REPORT: ETH-A - success
REPORT: ETH-B - success
REPORT: ETH-C - success
REPORT: GUSD-A - Can't borrow dai
REPORT: KNC-A - Can't borrow dai
REPORT: LINK-A - success
REPORT: LRC-A - Can't borrow dai
REPORT: MANA-A - success
REPORT: PAXUSD-A - Can't borrow dai
REPORT: RENBTC-A - Join doesnt have money
REPORT: TUSD-A - Can't borrow dai
REPORT: UNI-A - Can't borrow dai
REPORT: USDC-A - Can't borrow dai
REPORT: USDC-B - Can't borrow dai
REPORT: USDT-A - Can't borrow dai
REPORT: WBTC-A - Join doesnt have money
REPORT: WBTC-B - Join doesnt have money
REPORT: WBTC-C - Join doesnt have money
REPORT: YFI-A - success
REPORT: ZRX-A - Can't borrow dai
REPORT: MATIC-A - success
REPORT: WSTETH-A - success
REPORT: UNIV2DAIETH-A - Can't borrow dai
REPORT: UNIV2USDCETH-A - success
REPORT: UNIV2ETHUSDT-A - Can't borrow dai
REPORT: UNIV2WBTCDAI-A - Can't borrow dai
REPORT: UNIV2WBTCETH-A - Can't borrow dai
REPORT: UNIV2LINKETH-A - Can't borrow dai
REPORT: UNIV2UNIETH-A - Can't borrow dai
REPORT: UNIV2AAVEETH-A - Can't borrow dai
REPORT: UNIV2DAIUSDT-A - Can't borrow dai
```

core/simulations/steps/createVaultForCollateral.ts
```diff
@@ -0,0 +1,158 @@
...
+    const contract = await getContract(TEST_NETWORK, 'MCD_VAT');
+    const typeHex = ethers.utils.formatBytes32String(collateralType);
+    const { dust } = await contract.ilks(typeHex);
+    const { minUnitPrice } = await fetchVaultCollateralParameters(TEST_NETWORK, collateralType);
+    const minCollateralInVault = new BigNumber(new BigNumber(dust._hex).toFixed(0))
+        .shiftedBy(-RAD_NUMBER_OF_DIGITS)
+        .div(minUnitPrice);
+
+    await checkAvailableDebtForAmountAndMinUnitPrice(collateralType, minCollateralInVault, minUnitPrice);
+    return new BigNumber(minCollateralInVault.multipliedBy(1.1).shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed());
```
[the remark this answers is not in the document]
KirillDogadin-std: > What's the purpose of new BigNumber(value.toFixed())

bug, fixed. 

Should round down

KirillDogadin-std:
for some strange reasons there're collaterals for which the test fails but the simulation works out. Could you please validate this on your setup?

as of list of current state:

```
REPORT: cannot borrow collateral, AAVE-A
REPORT: cannot borrow collateral, BAL-A
REPORT: cannot borrow collateral, BAT-A
REPORT: cannot borrow collateral, COMP-A
REPORT: success, ETH-A
REPORT: test fails, ETH-B
REPORT: test fails, ETH-C
REPORT: no funds in join, GUSD-A
REPORT: cannot borrow collateral, KNC-A
REPORT: test fails, LINK-A
REPORT: cannot borrow collateral, LRC-A
REPORT: test fails, MANA-A
REPORT: cannot borrow collateral, PAXUSD-A
REPORT: no funds in join, RENBTC-A
REPORT: cannot borrow collateral, TUSD-A
REPORT: test fails, UNI-A
REPORT: cannot borrow collateral, USDC-A
REPORT: cannot borrow collateral, USDC-B
REPORT: cannot borrow collateral, USDT-A
REPORT: no funds in join, WBTC-A
REPORT: no funds in join, WBTC-B
REPORT: no funds in join, WBTC-C
REPORT: test fails, YFI-A
REPORT: cannot borrow collateral, ZRX-A
REPORT: test fails, MATIC-A
REPORT: test fails, WSTETH-A
REPORT: test fails, WSTETH-B
REPORT: test fails, UNIV2DAIETH-A
REPORT: test fails, UNIV2USDCETH-A
REPORT: cannot borrow collateral, UNIV2ETHUSDT-A
REPORT: test fails, UNIV2WBTCDAI-A
REPORT: test fails, UNIV2WBTCETH-A
REPORT: cannot borrow collateral, UNIV2LINKETH-A
REPORT: test fails, UNIV2UNIETH-A
REPORT: cannot borrow collateral, UNIV2AAVEETH-A
REPORT: cannot borrow collateral, UNIV2DAIUSDT-A
REPORT: test fails, UNIV2DAIUSDC-A
```

valiafetisov:
> for some strange reasons there're collaterals for which the test fails but the simulation works out

If the tests run async, maybe it's the problem with the `getLastVaultId`? Although I can see that you suppose to reset the network between the tests (maybe that is the problem?)

KirillDogadin-std:
- removing resetting of network helped.
  - reason of why this breaks tests not descovered
- added two collateral types to disabled list, listed the reasons as comments as well: 
  - collateral can have liquidation limits hit, can be fixed if we:
    - discover how to overwrite value on blockchain of `dog.sol` contract. The value is `hole` of collateral config type in the `ilks` mapping
    - discover how to adjust stability fee rate accumulation

latest state:
```
REPORT AAVE-A - cannot borrow dai
REPORT BAL-A - cannot borrow dai
REPORT BAT-A - cannot borrow dai
REPORT COMP-A - cannot borrow dai
REPORT ETH-A - success
REPORT ETH-B - success
REPORT ETH-C - success
REPORT GUSD-A - cannot borrow dai
REPORT KNC-A - cannot borrow dai
REPORT LINK-A - success
REPORT LRC-A - cannot borrow dai
REPORT MANA-A - success
REPORT PAXUSD-A - cannot borrow dai
REPORT RENBTC-A - join contract does not have the funds
REPORT TUSD-A - cannot borrow dai
REPORT UNI-A - cannot borrow dai
REPORT USDC-A - cannot borrow dai
REPORT USDC-B - cannot borrow dai
REPORT USDT-A - cannot borrow dai
REPORT WBTC-A - join contract does not have the funds
REPORT WBTC-B - join contract does not have the funds
REPORT WBTC-C - join contract does not have the funds
REPORT YFI-A - success
REPORT ZRX-A - cannot borrow dai
REPORT MATIC-A - success
REPORT WSTETH-A - success
REPORT UNIV2DAIETH-A - cannot borrow dai
REPORT UNIV2USDCETH-A - success
REPORT UNIV2ETHUSDT-A - cannot borrow dai
REPORT UNIV2WBTCDAI-A - cannot borrow dai
REPORT UNIV2WBTCETH-A - cannot borrow dai
REPORT UNIV2LINKETH-A - cannot borrow dai
REPORT UNIV2UNIETH-A - cannot borrow dai
REPORT UNIV2AAVEETH-A - cannot borrow dai
REPORT UNIV2DAIUSDT-A - cannot borrow dai

**Comment by valiafetisov** [comment]
Successful creation and liquidation of a vault:
```
ETH-A
ETH-B
ETH-C
LINK-A
MANA-A
RENBTC-A
WBTC-A
WBTC-B
WBTC-C
YFI-A
MATIC-A
WSTETH-A
UNIV2USDCETH-A
```
Collaterals which we don't support yet:
```
CRVV1ETHSTETH-A // We don't support fetching this vault type yet
UNIV2DAIUSDC-A // Liquidation limit too high (fails with "Dog/liquidation-limit-hit")
WSTETH-B // Does not accumulate stability fee rate (fails with "Dog/not-unsafe")
```

No longer supported collaterals (`MCD_VAT.ilks(collateralType).line === 0` ie [Debt Ceiling](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/vat.sol#L48) is 0)
```
AAVE-A
BAL-A
BAT-A
COMP-A
GUSD-A
KNC-A
LRC-A
PAXUSD-A
TUSD-A
UNI-A
USDC-A
USDC-B
USDT-A
ZRX-A
UNIV2DAIETH-A
UNIV2ETHUSDT-A
UNIV2WBTCDAI-A
UNIV2WBTCETH-A
UNIV2LINKETH-A
UNIV2UNIETH-A
UNIV2AAVEETH-A
UNIV2DAIUSDT-A
```

valiafetisov (approved):

valiafetisov (approved):

LukSteib:
Some questions:

> WSTETH-B // Does not accumulate stability fee rate (fails with "Dog/not-unsafe")

- Do we have an idea why this might be the case? What's strange to me is that there shouldn't be much of a difference between working `WSTETH-A` and `WSTETH-B`.

- Basically what we are doing to get a vault in a liquidatable state is a) skipping time and b) calling `drip` to collect stability fees, correct? What makes me wonder: Why do all of the other simulations we've heard about / seen in the past take to route of manipulating OSM price to achieve the same outcome (aka getting the vault in a liquidatable state)? 

---

For improvement PR
- [ ] support simulation for `CRVV1ETHSTETH-A`
- [ ] re-name the simulation in CLI from "Simulate liquidation Auctions" to "Open and liquidate vault"

LukSteib (approved):

PR #480, opened 2022-09-27
Small fixes for dark mode

zoey-kaiser:
Fixes:

- Mismatched Background color for loading component in dark mode
- LandingBlock buttons are now displayed with white text in dark mode
- CollateralAuctionTable got fixed widths for some values to ensure the table does not change in sizes when navigating between steps
- Fixed an issue where the icons in alerts was in a different position in the dark mode vs light mode


LukSteib (commented):
- LandingBlock text color ✔️ 

> CollateralAuctionTable got fixed widths for some values to ensure the table does not change in sizes when navigating between steps

So the fix basically prevents the table from shrinking when opening the side panel? Like indicated in s1 below.

> Mismatched Background color for loading component in dark mode
> Fixed an issue where the icons in alerts was in a different position in the dark mode vs light mode

Hard for me to review as I cannot follow what exactly was fixed here


s1:
image

valiafetisov (approved):
Making the table fixed is fine for me if it properly works on mobile. @zoey-kaiser thanks for taking initiative here!

LukSteib (approved):

zoey-kaiser:
I decided to remove the fixed table width. After re-evaluating the change, I noticed it did in fact (as raised by @valiafetisov) introduce problems with the mobile view in specific situations.

Therefore I believe it does not make sense to add it.

The dark mode fixes still apply!

valiafetisov:
> I noticed it did in fact introduce problems with the mobile view in specific situations

Can you please be more specific? I think the `antdv` table suppose to work well on mobile in general and also support fixed columns. Would this solve mentioned "specific situations"? Or what are they to start with?

zoey-kaiser:
> Can you please be more specific? I think the `antdv` table suppose to work well on mobile in general and also support fixed columns. Would this solve mentioned "specific situations"? Or what are they to start with?

Sure thing! The issue I ran into was that I set a fixed width of 75px for the index column. On the Desktop view this worked fine, but on mobile the word index was then broken into two lines, which I thought was not a good thing.

valiafetisov:
> but on mobile the word index was then broken into two lines

Maybe it would work if you add few more pixels? Otherwise this seems like still an improvement for all other columns. I suggest that we still apply fixed width to all tables (if horizontal scroll on mobile works well)

zoey-kaiser:
> Maybe it would work if you add few more pixels? Otherwise this seems like still an improvement for all other columns. I suggest that we still apply fixed width to all tables (if horizontal scroll on mobile works well)

I take back what I said about mobile being the main issue. The biggest issue is the fact that we display the table in two different versions, with and without explanations. The two tables are widely different in size, which leads to the big problems. 

If I set the size of every column to be too big, then we get a scroll bar on the version with explanations. If I set it too small, the button column looks way too big on the version without. The only solution for this issue would be to use relative sizes instead of absolute sizes, which in the end does not solve the problem of the columns changing sizes when switching from one view to another. 

The table natively chooses the best relative size depending on the table size, which is why I would want to avoid playing around with that, as it will only cause us more problems.

However, what I did change was the `state` column in the collateral auctions table. In that table we had a column called "Auction ends". In the new auctions, we redesigned this to be a more neutral state column. I brought the old collateral table in line with this for more consistency. We also show other states then ends in in the column, which makes the change useful and appropriate IMO. If we do not like this change, I can revert it!

Screenshot 2022-10-17 at 16 11 52

valiafetisov:
> If I set it too small, the button column looks way too big on the version without

Wouldn't it be solved via `min-width` instead of `width`? Even if antdv table doesn't support it out of the box, I think you can provide a custom cell renderer function to implement this (or directly set the cell style yourself?) (:

zoey-kaiser:
Agreed on in Daily Call:

- move table fixed widths into a separate PR and Issue (I will create one later)

Fixed in this PR:

- Mismatched Background color for loading component in dark mode 
- LandingBlock buttons are now displayed with white text in dark mode
- Fixed an issue where the icons in alerts was in a different position in the dark mode vs light mode
- Made `CollateralAuctionsTable` "state" column be more in line with the design of the other tables

valiafetisov (approved):

PR #481, opened 2022-09-27
Improved number formatting

### LINKED ISSUE #474: Introduce thousand separators for big numbers
### Goal
Numbers are displayed with thousand separators throughout our UI

### Context
The newly introduced barking on underwater vaults feature displays a lot of big numbers. Especially in the limits section (see screenshot below).
To improve the UX and make these numbers more understandable, we want to introduce thousand separators.

### Tasks
- [ ] Introduce thousand separators (ie. display e.g. `1,000,000.45` instead of `1000000.45`)
- [ ] Include dedicate story
- [ ] Ensure that generally our `AnimatedNumbers` do not break

### Assets
image

zoey-kaiser:
closes #474 

I decided to use a custom regex expression over native JavaScript number functions, because the native functions only support the number type. As we do a lot of other custom formatting, the end result is always a string, resulting in them being useless. In addition to this formatting the number before we add our automatic decimal place messes up the end results, so I thought it was easier to apply the thousand commas last with regex rather then before. It also helps with the fact that we use numbers, BigNumbers and strings in some places, allowing me to simplify the amount of types the function needs to support. 

Screenshot 2022-09-27 at 15 37 24

Checklist:


frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -32,8 +33,18 @@ export default Vue.extend({
...
+            default: false,
+        },
     },
     computed: {
+        safeValue(): number {
+            if (BigNumber.isBigNumber(this.value)) {
+                return this.value.toNumber();
+            }
+            return this.value;
+        },
```
valiafetisov, line 47: Why is this needed? We should not covert big numbers right before we want to display them because this might produce:
- big numbers being displayed using scientific notation
- small numbers loosing precision

frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -32,8 +33,18 @@ export default Vue.extend({
             type: Number,
             default: undefined,
         },
+        formatted: {
+            type: Boolean,
+            default: false,
+        },
```
valiafetisov, line 39: If we want to introduce this feature, why not setting it on by default? Then, please rename it into `disableThousandSeparators`, since `formatted` doesn't really tell what it does (the number is already formatted anyway)

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -180,7 +183,7 @@ export default Vue.extend({
     },
     methods: {
         format(value: BigNumber): string {
-            return formatToAutomaticDecimalPoints(value);
+            return formatWithThousandSeparators(formatToAutomaticDecimalPoints(value));
```
valiafetisov, line 186: I would use `formatWithThousandSeparators` inside `formatToAutomaticDecimalPoints` and make it accept second parameter with options `formatToAutomaticDecimalPoints(value, { disableThousandSeparators: true })`

core/src/helpers/formatToAutomaticDecimalPoints.ts
```diff
@@ -58,6 +58,10 @@ function limitedValue(value: number | BigNumber): number | BigNumber {
     return value;
 }
 
+export function formatWithThousandSeparators(value: string): string {
+    return value.replace(/\B(?<!\.\d*)(?=(\d{3})+(?!\d))/g, ',');
```
valiafetisov, line 62: As suggested in the call, why not use `intl` for that? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat and make it country-specific out of the box?

zoey-kaiser: See PR description: 

> I decided to use a custom regex expression over native JavaScript number functions, because the native functions only support the number type. As we do a lot of other custom formatting, the end result is always a string, resulting in them being useless. In addition to this formatting the number before we add our automatic decimal place messes up the end results, so I thought it was easier to apply the thousand commas last with regex rather then before. It also helps with the fact that we use numbers, BigNumbers and strings in some places, allowing me to simplify the amount of types the function needs to support.

frontend/components/common/formatters/AnimatedNumber.vue
zoey-kaiser: We did this beforehand as well. Beforehand we passed the `format(value)` as the value to the animated Number component. The external `animated-number` component does not support BigNumber, so we need to format it to a number before we can pass it. 

Before this PR, we passed it with our format function, which outputted a string, that could still be parsed as a int. With the new function, a string is returned that cannot be formatted back into a number, resulting in an error. 

This is why I added a `safeValue`, which is always a number. This number is the base number, the animated-number component uses. Then I pass the format function to it, which it calls and then the decimal place is added and the string is correctly formatted.

valiafetisov: > so we need to format it to a number before we can pass it

Not true. The formatting function as well as the `value` prop [accepts string or number](https://www.npmjs.com/package/animated-number-vue#props). Therefore it's preferable to send string.

> This is why I added a `safeValue`

As discussed many times in this repo, we can not use `.toNumber()` under any circumstances since we risk loosing precision and/or formatting a number with scientific notation, which most people wouldn't understand

core/src/helpers/formatToAutomaticDecimalPoints.ts
valiafetisov: Nit: I would add a _shorter_ comment above the line with regexp explaining this decision

frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -32,8 +32,17 @@ export default Vue.extend({
             type: Number,
             default: undefined,
         },
+        disableThousandSeparators: {
+            type: Boolean,
+            default: false,
+        },
     },
     computed: {
+        formattedValue(): string {
```
zoey-kaiser, line 41: We need this formattedValue computed property, as we need to pass a certain type of number to the external `AnimatedNumber` component. 

It takes in either a string or a number, which is why I decided to use our `formatToAutomaticDecimalPoints` function twice, as it ensures that we do not lose any kind of precision when converting the BigNumber to a string or number.

frontend/components/common/formatters/AnimatedNumber.vue
```diff
@@ -43,7 +52,9 @@ export default Vue.extend({
     },
     methods: {
         format(value: number | BigNumber): string {
-            return formatToAutomaticDecimalPoints(value, this.decimalPlaces);
+            return formatToAutomaticDecimalPoints(value, this.decimalPlaces, {
```
zoey-kaiser, line 55: I then use the function again here to add the final decimal points and correctly format the numbers between the start and end value (when the animation plays)

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -238,7 +238,7 @@ export default Vue.extend({
     },
     methods: {
         format(value: BigNumber): string {
-            return formatToAutomaticDecimalPoints(value);
+            return formatToAutomaticDecimalPoints(value, undefined, { formatWithThousandSeparators: true });
```
valiafetisov, line 241: If it's `true` by default, you don't need to change this call, right?

core/src/helpers/formatToAutomaticDecimalPoints.ts
```diff
@@ -58,11 +58,25 @@ function limitedValue(value: number | BigNumber): number | BigNumber {
     return value;
 }
 
+// We choose regex over native JS solutions, as using native solutions often misformatted the decimal points, leading to inconsistent number formatting
```
valiafetisov, line 61: This comment is strange and is not the same as in https://github.com/sidestream-tech/unified-auctions-ui/pull/481/files#r983341267. There the main point there is: `native functions only support the number type`, here: `native solutions often misformatted the decimal points` – which doesn't even sounds convincing. What do you mean by `often misformatted`?

zoey-kaiser: I updated the comment to be more inline!

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
valiafetisov: > you don't need to change this call

This wasn't addressed, right? (there is even no `formatWithThousandSeparators` anymore)

I also wonder why wasn't catch by the types?

zoey-kaiser: Hmm, I am wondering that as well. The tests should have failed IMO. I fixed it now!

valiafetisov (approved):

LukSteib (approved):

PR #482, opened 2022-09-27
Create `WithdrawDaiFromVat` component and stories

### LINKED ISSUE #473: Add `WithdrawDaiFromVat` panel to liquidation flow
### Goal
A user can withdraw DAI from Vat via a dedicated panel in the liquidation flow

### Context
The liquidation incentive after a succssful vault liquidation ends up in the user's VAT balance. Since this might be unintuitive we should provide a user with the known withdrawal panel directly within the liquidation flow.
Open question remains where to introduce the panel. See two alternative proposals in asset section below

### Tasks
- [x] Decide on step where to introduce the withdrawal panel
- [x] Implement panel to liquidation flow
   - [x] Use existing panel `WithdrawDAIPanel`
   - [x] Adjust text as shown in mock below
- [x] On the transaction page: replace "This vault was liquidated N x ago in the transaction" with "This vault was _last liquidated_ N x ago in the transaction" because we've learned that the vault can be liquidated many times
- [x] On the transaction page: add panel below this message to withdraw DAI
- [x]  On the vault page: replace `This vault was liquidated N x ago in the transaction` with the table of transactions (date, hash, auction id with the link) that will display all transactions (like initially planned), not just one 

### Assets
**Alternative 1: Withdraw panel on liquidated page**

image

+ +Only at this step it will be possible to withdraw DAI 
- -Panel appears only after vault is liquidated and hence user doesn't have the "full picture" of the flow on the transaction page; liquidation page has two CTAs (proceed to auction, withdraw Dai from VAT)

**Alternative 2: Include as last panel on liquidationTransactionPage**

+ +Full flow is displayed on the transaction page
- -The panel will no longer be available once it is actually needed (ie. after successful liquidation)

aomafarag:
Closes #473.

Checklist:

### Assets

1. Panel
Screenshot from 2022-09-27 16-54-33

2. Vault transaction
Screenshot from 2022-09-27 16-55-51

3. Vault page
Screenshot from 2022-09-27 16-57-11


LukSteib (changes requested):
- [x] Can we concatenate the tx hash of the table on the transaction page. Although horizontal scroll of the side panel works, it still looks broken and I expect that we can show all three columns if we reduce size of the tx hash column (see s1 below)
- [x] We do not properly fetch DAI in VAT after liquidation. The introduced withdrawal panel is not refreshed with the correct DAI balance until opening the wallet management popup (see s2)
- [ ] Once the DAI in VAT amount is fetched correctly the auth state is no longer displayed correctly (see r1 below). One has to click the auth button again and without an actual tx it's in the correct state then.



S1
Screenshot 2022-09-28 at 15 16 19



S2:
image

r1
https://www.loom.com/share/5d74709337c9461c919590879f4566c8

frontend/components/vault/Vault.vue
```diff
@@ -157,16 +157,36 @@
                 </div>
             </Alert>
             <TextBlock class="mt-4">
-                This vault was liquidated <TimeTill :date="vaultTransaction.pastLiquidations[0].liquidationDate" /> in
-                the transaction <FormatAddress :value="vaultTransaction.pastLiquidations[0].transactionHash" />.
+                This vault was liquidated in the following transactions:
+                <table class="table-auto mt-4">
```
zoey-kaiser, line 161: How will this table look in dark mode? Have a look at #311 where I added a similar table and added styling for light and dark mode!

frontend/components/vault/Vault.vue
```diff
@@ -157,16 +157,36 @@
...
+                            <th>Liquidated since</th>
+                            <th>Auction link</th>
+                        </tr>
+                    </thead>
+                    <tbody>
+                        <tr
+                            v-for="liquidation in vaultTransaction.pastLiquidations"
+                            :key="liquidation.transactionHash"
+                        >
+                            <td><FormatAddress :value="liquidation.transactionHash" /></td>
```
zoey-kaiser, line 174: Agree with Lukas here to minimize the transactionhash!

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
+        <TextBlock v-if="isExplanationsShown">
+            After liquidating a vault, the DAI-based liquidation incentive will end up in your VAT account. One more
+            transaction is required to move that DAI to your wallet.
+        </TextBlock>
+        <div class="my-4">
+            <WalletAuthorizationCheckPanel
```
zoey-kaiser, line 9: Do we want to add a `WalletConnectionCheck` panel here too? Incase someone views it without having their wallet connected? (Super edge case, I know!)

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,116 @@
...
+                :is-loading="isAuthorizing"
+                @authorizeWallet="$emit('authorizeWallet')"
+            />
+        </div>
+        <div class="flex justify-end mt-2 gap-5">
+            <BaseButton
+                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw || isWithdrawing"
+                @click="$emit('manageVat')"
+            >
+                Withdraw DAI partially
```
zoey-kaiser, line 23: I would rephrase this to:

"Partially withdraw DAI", I think that sounds better.

Or do we take another description, as this opens the VAT modal right? Something in the direction of:

"Manage VAT"?

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,116 @@
...
+        TextBlock,
+        FormatCurrency,
+        WalletAuthorizationCheckPanel,
+    },
+    props: {
+        walletAddress: {
+            type: String,
+            default: undefined,
+        },
+        daiVatBalance: {
```
zoey-kaiser, line 64: This is something Valia and I discussed a while ago. This component does not work without `daiVatBalance` and a few other props. Therefore we should stop making these props optional and just add a `v-if` on the component in the parent to ensure it is hidden when in this case no wallet is connected. 

This is better for testing, as if it is ever not properly hidden or a value is not correctly passed, we get a error thrown during the tests. 

(If the prop is only optional we may never know we forgot to pass it somewhere)

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,116 @@
...
+            type: Boolean,
+            default: false,
+        },
+        isExplanationsShown: {
+            type: Boolean,
+            default: true,
+        },
+        vaultState: {
+            type: String as Vue.PropType<VaultTransactionState>,
+            default: 'liquidatable',
```
zoey-kaiser, line 86: This should also never be set as default! Lets make it required.

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,116 @@
...
+        },
+    },
+    computed: {
+        currentStateAndTitle(): PanelProps {
```
zoey-kaiser, line 90: You also have no state here what to display if no wallet is connected! 

This related to my other two points: Will we ever show this panel in a state if no wallet is connected?

YES: Keep the props optional and add a state here to handle it
NO: Make the props above required, to ensure there wont be an issue here!

frontend/containers/VaultsContainer.vue
```diff
@@ -37,7 +45,16 @@ export default Vue.extend({
...
             isWalletLoading: 'isLoading',
             walletAddress: 'getAddress',
+            walletBalances: 'walletBalances',
+            isWithdrawing: 'isDepositingOrWithdrawing',
         }),
+        ...mapGetters('authorizations', {
+            isAuthorizing: 'isWalletAuthorizationLoading',
+            isWalletAuthorized: 'isWalletAuthorizationDone',
+        }),
+        daiVatBalance(): BigNumber | undefined {
```
zoey-kaiser, line 55: Why do we have this extra? Can't you just pass `walletbalances?.walletVATDAI` directly in the component above?

zoey-kaiser (changes requested):

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,112 @@
...
+        BaseButton,
+        BasePanel,
+        TextBlock,
+        FormatCurrency,
+        WalletAuthorizationCheckPanel,
+    },
+    props: {
+        walletAddress: {
+            type: String,
+            default: undefined,
```
zoey-kaiser, line 61: These are all still set with a default value. My recommendation was to set these to required.

```suggestion
            required: true,
```

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,112 @@
...
+        },
+        daiVatBalance: {
+            type: Object as Vue.PropType<BigNumber>,
+            default: undefined,
```
zoey-kaiser, line 65: Same here

zoey-kaiser (changes requested):

zoey-kaiser (approved):

LukSteib (changes requested):
- [x] the two problems raised in prev review comment haven't been addressed (any update on those? 
- [ ] additionally: IMO there is no need for the blue info message on top of the transaction page (see screenshot below). The table covers this information already. So I would suggest to only keep this on the liquidation page 

image

LukSteib:
> That way, the user can easily navigate to that latest transaction if they were the one who had just executed it.

Fine for me to display it.

frontend/components/vault/VaultLiquidationTransactionFlow.vue
```diff
@@ -16,8 +16,23 @@
         <VaultLiquidationTransactionTable v-if="!wasLiquidated" class="my-4" :vault-transaction="vaultTransaction" />
         <TextBlock class="mt-4">
             <template v-if="wasLiquidated">
-                This vault was liquidated <TimeTill :date="vaultTransaction.pastLiquidations[0].liquidationDate" /> in
-                the transaction <FormatAddress :value="vaultTransaction.pastLiquidations[0].transactionHash" />.
+                This vault was last liquidated
+                <TimeTill :date="vaultTransaction.pastLiquidations[0].liquidationDate" /> in the transaction
+                <FormatAddress :value="vaultTransaction.pastLiquidations[0].transactionHash" />.
+                <WithdrawDaiFromVat
```
valiafetisov, line 22: Does this panel have rounded corners at the moment? Otherwise, let's put it into a separate empty div, so it will have them

frontend/components/vault/WithdrawDaiFromVat.vue
```diff
@@ -0,0 +1,112 @@
+<template>
+    <BasePanel :current-state="currentStateAndTitle.name">
+        <template #title>{{ currentStateAndTitle.title }}</template>
```
valiafetisov, line 3: Why was this component copied over? I can see the identical `WithdrawDAIPanel` that should've been reused instead

LukSteib (approved):

PR #485, opened 2022-09-29
Allow partial liquidations

### LINKED ISSUE #478: Improve limits panel to account for partial liquidation
### Goal

Limits panel do not prevent partial execution

### Context

Currently, the vault ID proposed as an example in the production `/vaults` page is presented with `VaultLiquidationLimitsCheckPanel` showing an error, while in fact it's partial execution is possible, so it should be presented as warning, not error (and not pass error state up in the `isCorrect` prop).

The best way to determine (on the frontend) if partial execution is possible is to be suggested by @KirillDogadin-std. Since it's a complex calculation involving checking if the part that are taken or the part that is left out is big enough, I suggest it will be derived from some vault values, rather than recomputed on the frontend.

Suggested changes in this PR:
- [ ] Highlight reached limits (-23364594.25 DAI in the screenshot below) in orange for partial liquidation (or red for no liquidation)
- [ ] Change color and content of the title accordingly (red or orange)

### Assets

Screenshot 2022-09-26 at 17 51 26

### Tasks

- [ ] Improve visibility on which limits are reached
- [ ] Improve signalling of partial execution via title and color
- [ ] Unblock partial execution

aomafarag:
Closes #478.

Screenshot from 2022-09-29 16-01-21

Checklist:

### DIFF
--- a/frontend/components/common/other/BasePanel.vue
+++ b/frontend/components/common/other/BasePanel.vue
@@ -4,7 +4,7 @@
             <div class="Icon" :class="iconClass">
                 <Icon v-if="currentState === 'inactive'" type="close" />
                 <Icon v-else-if="currentState === 'incorrect'" type="warning" theme="filled" />
-                <Icon v-else-if="currentState === 'partlyIncorrect'" type="issues-close" />
+                <Icon v-else-if="currentState === 'attention'" type="exclamation-circle" theme="filled" />
                 <Icon v-else-if="currentState === 'correct'" type="check" />
                 <Icon v-else-if="currentState === 'notice'" type="info-circle" theme="filled" />
             </div>
@@ -37,7 +37,7 @@ const STATES = [
         isExpanded: true,
     },
     {
-        name: 'partlyIncorrect',
+        name: 'attention',
         titleClass: 'text-orange-500',
         iconClass: 'text-orange-500',
         isExpanded: true,
--- a/frontend/components/common/other/BasePanel.vue
+++ b/frontend/components/common/other/BasePanel.vue
@@ -4,6 +4,7 @@
             <div class="Icon" :class="iconClass">
                 <Icon v-if="currentState === 'inactive'" type="close" />
                 <Icon v-else-if="currentState === 'incorrect'" type="warning" theme="filled" />
+                <Icon v-else-if="currentState === 'partlyIncorrect'" type="warning" theme="filled" />
                 <Icon v-else-if="currentState === 'correct'" type="check" />
                 <Icon v-else-if="currentState === 'notice'" type="info-circle" theme="filled" />
             </div>
@@ -33,6 +34,12 @@ const STATES = [
         name: 'incorrect',
         titleClass: 'text-red-500',
         iconClass: 'text-red-500',
+        isExpanded: true,
+    },
+    {
+        name: 'partlyIncorrect',
+        titleClass: 'text-orange-500',
+        iconClass: 'text-orange-500',
         isExpanded: true,
     },
     {
--- a/frontend/components/common/other/BasePanel.vue
+++ b/frontend/components/common/other/BasePanel.vue
@@ -4,7 +4,7 @@
             <div class="Icon" :class="iconClass">
                 <Icon v-if="currentState === 'inactive'" type="close" />
                 <Icon v-else-if="currentState === 'incorrect'" type="warning" theme="filled" />
-                <Icon v-else-if="currentState === 'partlyIncorrect'" type="warning" theme="filled" />
+                <Icon v-else-if="currentState === 'partlyIncorrect'" type="issues-close" />
                 <Icon v-else-if="currentState === 'correct'" type="check" />
                 <Icon v-else-if="currentState === 'notice'" type="info-circle" theme="filled" />
             </div>
--- a/frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
+++ b/frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
@@ -39,30 +39,28 @@ storiesOf('Panels/VaultLiquidationLimitsCheckPanel', module)
             };
         },
     }))
-    .add('Global Limits Reached', () => ({
+    .add('Partial Liquidation Possible', () => ({
         ...common,
         data() {
             return {
                 ...common.data(),
                 vaultTransaction: {
                     ...vaultTransaction,
-                    debtDai: new BigNumber(faker.datatype.float({ min: 1000 })),
-                    incentiveRelativeDai: new BigNumber(faker.finance.amount()),
-                    incentiveConstantDai: new BigNumber(faker.finance.amount()),
+                    state: 'liquidatable',
+                    debtDai: new BigNumber(faker.datatype.float({ min: 10000 })),
                 },
             };
         },
     }))
-    .add('Collateral Limits Reached', () => ({
+    .add('No Liquidation Possible', () => ({
         ...common,
         data() {
             return {
                 ...common.data(),
                 vaultTransaction: {
                     ...vaultTransaction,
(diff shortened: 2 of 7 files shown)


frontend/components/common/other/BasePanel.vue
```diff
@@ -4,6 +4,7 @@
             <div class="Icon" :class="iconClass">
                 <Icon v-if="currentState === 'inactive'" type="close" />
                 <Icon v-else-if="currentState === 'incorrect'" type="warning" theme="filled" />
+                <Icon v-else-if="currentState === 'partlyIncorrect'" type="warning" theme="filled" />
```
zoey-kaiser, line 7: I would recommend using the icon `<a-icon type="issues-close" />`. As I feel like it perfectly shows the state of it being partially correct.

Screenshot 2022-09-29 at 17 03 14

frontend/tailwind.config.js
```diff
@@ -21,6 +21,9 @@ module.exports = {
                     light: '#12302e',
                     dark: '#081615',
                 },
+                orange: {
+                    500: '#f97316',
```
zoey-kaiser, line 25: You can set this to `DEFAULT` which will allow us to use `text-orange` and still add future versions when we need it! (500 is also always the default)

zoey-kaiser (changes requested):

zoey-kaiser (approved):

frontend/components/common/other/BasePanel.vue
```diff
@@ -35,6 +36,12 @@ const STATES = [
         iconClass: 'text-red-500',
         isExpanded: true,
     },
+    {
+        name: 'partlyIncorrect',
```
valiafetisov, line 40: Let's make it `attention` so it can be used in other places
```suggestion
        name: 'attention',
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -124,41 +124,69 @@ export default Vue.extend({
...
                 return {
                     name: 'incorrect',
                     title: 'Current global liquidation limits are reached',
                 };
             }
             if (this.collateralDifference.isNegative()) {
+                if (this.vaultTransaction.state === 'liquidatable') {
+                    return {
+                        name: 'partlyIncorrect',
+                        title: `Current ${this.vaultTransaction.collateralType} liquidation limits are reached`,
```
valiafetisov, line 187:
```suggestion
                        name: 'partlyIncorrect',
                        title: `Current ${this.vaultTransaction.collateralType} liquidation limits are partially reached`,
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.vue
```diff
@@ -124,41 +124,69 @@ export default Vue.extend({
...
                     name: 'inactive',
                     title: 'Current liquidation limits are unknown',
                 };
             }
             if (this.globalDifference.isNegative()) {
+                if (this.vaultTransaction.state === 'liquidatable') {
+                    return {
+                        name: 'partlyIncorrect',
+                        title: 'Current global liquidation limits are reached',
+                    };
```
valiafetisov, line 176:
```suggestion
                        name: 'partlyIncorrect',
                        title: 'Current global liquidation limits are partially reached',
                    };
```

frontend/components/panels/VaultLiquidationLimitsCheckPanel.stories.js
```diff
@@ -39,30 +39,28 @@ storiesOf('Panels/VaultLiquidationLimitsCheckPanel', module)
             };
         },
     }))
-    .add('Global Limits Reached', () => ({
+    .add('Partial Liquidation Possible', () => ({
```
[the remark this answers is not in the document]
valiafetisov: Yes, please read the issue for more context. Please also tick the boxes in the issue that were implemented

valiafetisov (commented):
- [ ] Based on the discussion @aomafarag and @KirillDogadin-std had in the issue, I would also add a third row to the table with `Current global limit` and `Current XXX limit` called `Will be liquidated` (in bold) that will show the actual amount that will be liquidated – otherwise it's quite confusing what those numbers means

zoey-kaiser:
> [ ] Based on the discussion @aomafarag and @KirillDogadin-std had in the issue, I would also add a third row to the table with `Current global limit` and `Current XXX limit` called `Will be liquidated` (in bold) that will show the actual amount that will be liquidated – otherwise it's quite confusing what those numbers means

I tried to calculate a `maximumLiquidationAmount` however I ran into a specific question. 

We currently calculate the debtTimesPenaltyRatio like follows:
```ts
debtTimesPenaltyRatio(): BigNumber {
   return this.vaultTransaction.debtDai.times(this.vaultTransaction.liquidationPenaltyRatio);
},
```

Will this value stay the same even after changing how much DAI is being liquidated? I assume not. However, this would make the calculation of the actual DAI being liquidated a little bit harder, as I would always need to recalcualte the next value with the added liquidationPenaltyRatio. In the end if this is true we would need to implemente a simular calculation as with the step/cut, where we were able to determine when the step/cut parameters would decrease the price so much that a collateral auction became profitable. 

Is this correct or am I misunderstanding the system? I tried my best to read the convo in the issue, however I got a bit lost during it 😞 

My other clarification issue is the following: 
- Is it correctly that if either the global or collateral difference is negative, that a partial liquidation is always possible as long as the vault state is `liquidatable`? Do we change the state to `not-liquidatable` as soon as a partial liquidation is not possible? This is what I understood from reading the issue, however I wanted to double check as it confused me a bit and I think its important!

zoey-kaiser:
As discussed in the daily we removed the display of how much the current vault would impact the global and collateral limits.

The only indication of the vaults impact is if the end limit is displayed in orange (partial liquidation) or red (no liquidation).

There is also a new indication of the maximum possible liquidation amount based on the global and collateral limits.

LukSteib:
> There is also a new indication of the maximum possible liquidation amount based on the global and collateral limits.

I like this new indication of the maximum possible liquidation amount

> The only indication of the vaults impact is if the end limit is displayed in orange (partial liquidation) or red (no liquidation).

I am not able to test these cases of partial or no liquidation since there is no simulation in place that comprises such a situation. 
Only comment based on storybook: 
- [ ] In the "Partial liquidation possible" story the end limit is still displayed in black instead of the orange color coding.

LukSteib (dismissed):

valiafetisov (dismissed):

frontend/components/panels/VaultLiquidationPanel.vue
```diff
@@ -15,15 +15,15 @@
...
             >
                 Liquidate to another wallet
             </BaseButton>
             <BaseButton
                 type="primary"
-                :disabled="!isLiquidatable || disabled"
+                :disabled="disabled"
                 :is-loading="isLiquidating"
                 @click="$emit('liquidate', walletAddress)"
             >
```
valiafetisov, line 29: Why was it removed? If the vault is not in `liquidatable` state it can not be removed, right?

zoey-kaiser: This change was made by @aomafarag. I do not know either, maybe there was a reason we don't see?

valiafetisov: No, [specified line only checks for the limits](https://github.com/sidestream-tech/unified-auctions-ui/blob/1e885274400b0348d2f473d6f19910b96a2675d3/frontend/components/vault/VaultLiquidationTransactionFlow.vue#L50), but it's (obviously) not sufficient to liquidate a specific vault if the limits are not reached. If it would be true any current vault could've been liquidated right now

valiafetisov (approved):

PR #488, opened 2022-10-06
Display liquidatable amount

### LINKED ISSUE #486: Display partial liquidation amount
### Goal

Display how much of a vault will be liquidated in a partial liquidation 

### Context

For a partial liquidation of a `liquidatable` vault, we only display the difference between each liquidation limit and the vault's debt multiplied by the liquidation penalty ration in the `VaultLiquidationLimitsCheckPanel` (see [screenshot](#screenshot)).

We would like to additionally display the actual amount of the vault's debt that would be liquidated.

### Assets

1. <a name="screenshot">Screenshot</a>
image

2. Link to original issue: https://github.com/sidestream-tech/unified-auctions-ui/issues/478
3. Link to the proposal in the PR: https://github.com/sidestream-tech/unified-auctions-ui/pull/485#pullrequestreview-1129815860

### Tasks

- [ ] Add a new row in **bold** below the calculations of the aforementioned differences, with a `Will be liquidated` text on the left and the partial liquidation amount on the right.

aomafarag:
Closes #486.

Screenshot from 2022-10-06 17-28-20

Checklist:

### DIFF
--- a/frontend/components/panels/WithdrawDAIPanel.vue
+++ b/frontend/components/panels/WithdrawDAIPanel.vue
@@ -19,7 +19,7 @@
             <BaseButton :disabled="isWithdrawing" @click="$emit('manageVat')"> Manage DAI in VAT </BaseButton>
             <BaseButton
                 type="primary"
-                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw"
+                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw || isRefreshing"
                 :is-loading="isWithdrawing"
                 @click="$emit('withdrawAllDaiFromVat')"
             >
--- a/frontend/components/panels/WithdrawDAIPanel.vue
+++ b/frontend/components/panels/WithdrawDAIPanel.vue
@@ -19,7 +19,7 @@
             <BaseButton :disabled="isWithdrawing" @click="$emit('manageVat')"> Manage DAI in VAT </BaseButton>
             <BaseButton
                 type="primary"
-                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw || isRefreshing"
+                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw"
                 :is-loading="isWithdrawing"
                 @click="$emit('withdrawAllDaiFromVat')"
             >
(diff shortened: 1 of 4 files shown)


frontend/components/panels/WithdrawDAIPanel.vue
```diff
@@ -19,7 +19,7 @@
             <BaseButton :disabled="isWithdrawing" @click="$emit('manageVat')"> Manage DAI in VAT </BaseButton>
             <BaseButton
                 type="primary"
-                :disabled="!isWalletConnected || !isWalletAuthorized || !hasDaiToWithdraw || isRefreshing"
```
zoey-kaiser, line 22: Why did we remove the isRefreshing? If we are retrieving new values from the chain we don't want the user to execute any transactions as the data may be outdated. 

We also don't want the button to be loading while refreshing, as we use the loading indicator to show if the user is withdrawing.

zoey-kaiser (changes requested):

zoey-kaiser (dismissed):

valiafetisov:
Replication URL https://unified-auctions.makerdao.com/vaults?network=mainnet&vault=20775

LukSteib:
> Else, render Should be liquidated instead, with the vault debt as its value, and add an Explain text explaining that vaults cannot be partially or fully liquidated unless they are in a liquidatable state.

To me this sounds like confusing UX. The `VaultLiquidationLimitsCheckPanel` should only check if limits are reached or not without incorporating the vaults' liquidation state into the logic. So in the example case above I would expect that
* `VaultLiquidationLimitsCheckPanel` is shown in "Partial liquidation possible" state
* "Maximum liquidation amount" and "Will be liquidated" figures are displayed (in current state I would expect 5.000.000 to be displayed here)

KirillDogadin-std:
To me it seems like all the numbers you need are already exposed to the frontend.

Rephrasing your goal: you want to show if liquidation is possible / partially possible / impossible based on the limits only and neglect all the other factors that play into the determining the liquidatability state.

Hence:
  - you know the leftovers in the limits already,
  - you know the amount of debt in the vault
  - you know the minimal allowed leftower of the vault. (`minimalAuctionedDai`)

This means you're able to implement the following:

  - if the debt is less than the available limit - black coloring
  - if the debt is greater than the available limit - this is potentially partial liquidation.
    - if the leftover after partial liquidation is less than `minimalAuctionedDai` OR the auctioned amount is less than `minimalAuctionedDai` - red coloring
    - else orange coloring
  - if the limits are reached - red coloring

No such tool in the core. You can the function that returns 1 of 3 states based on the logic above and call it in the store.

valiafetisov:
The new mock

MakerDAO vaults@2x

LukSteib (changes requested):
Taking the same vault as reference (see below) - vaultID: 20775 (mainnet)

- [ ] I don't like the color coding - to me neither red nor orange color coding adds any value. I would propose to not use the number color coding

Other than that it's fine for me


image

LukSteib (approved):

PR #490, opened 2022-10-07
Reintegrate refresh logic

### LINKED ISSUE #489: Re-integrate refresh logic in `WithdrawDAIPanel`
### Goal

Re-integrate refresh logic in `WithdrawDAIPanel`

### Context

After we standardized the behavior of the secondary button in the `WithdrawDAIPanel` by replacing `Refresh` with `Manage DAI in VAT`, I proceeded to remove refresh-related logic from the panel, i.e. the `isRefreshing` prop. However, the [`WalletMKRBalanceCheckPanel`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/auction/surplus/SurplusAuctionTransactionFlow.vue#L33) in the `SurplusAuctionTransactionFlow`, which also includes the `WithdrawDAIPanel`, makes use of the refresh functionality. Hence, disabling the withdraw button in the `WithdrawDAIPanel` while the `WalletMKRBalanceCheckPanel` is refreshing the wallet is necessary to avoid concurrent withdrawal.

Credits to @zoey-kaiser for pointing that out.

### Assets

Screenshot from 2022-10-07 12-36-59

### Tasks

- [x] Re-integrate refresh logic in `WithdrawDAIPanel`

aomafarag:
Closes #489.

Screenshot from 2022-10-07 12-38-14

Checklist:

### DIFF
--- a/frontend/components/auction/surplus/SurplusAuctionTransactionFlow.stories.js
+++ b/frontend/components/auction/surplus/SurplusAuctionTransactionFlow.stories.js
@@ -17,6 +17,7 @@ const common = {
         isConnectingWallet: false,
         isRefreshingWallet: false,
         isSettingAllowance: false,
+        isWalletAuthorized: false,
         isAuthorizing: false,
 
         walletAddress: undefined,
--- a/frontend/components/panels/WithdrawDAIPanel.vue
+++ b/frontend/components/panels/WithdrawDAIPanel.vue
@@ -74,6 +74,10 @@ export default Vue.extend({
             type: Boolean,
             default: false,
         },
+        isRefreshing: {
+            type: Boolean,
+            default: false,
+        },
         isExplanationsShown: {
             type: Boolean,
             default: true,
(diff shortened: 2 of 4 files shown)


frontend/components/panels/WithdrawDAIPanel.vue
```diff
@@ -74,6 +74,10 @@ export default Vue.extend({
             type: Boolean,
             default: false,
         },
+        isRefreshing: {
```
zoey-kaiser, line 77: Do we want to rename this to just `disabled`? We currently only provide named actions when it actually changes something specific to this panel (Like a loading icon, texts etc). 

Changing it to disabled just makes it extendable if we ever need to disable it for another reason.

frontend/components/auction/surplus/SurplusAuctionTransactionFlow.stories.js
```diff
@@ -17,6 +17,7 @@ const common = {
         isConnectingWallet: false,
         isRefreshingWallet: false,
         isSettingAllowance: false,
+        isWalletAuthorized: false,
```
zoey-kaiser, line 20: Why do we add this here? Was it forgotten?

zoey-kaiser (changes requested):

zoey-kaiser (approved):

valiafetisov (approved):

PR #494, opened 2022-10-11
Crveth support

### LINKED ISSUE #484: Add support for CRVV1ETHSTETH collateral vaults.
### Goal

Have vault module in `core/` be able to operate with collaterals of type `CRVV1ETHSTETH` just as it does with all the other types (e.g. `ETH-A`). CRVV1ETHSTETH Vaults should be fetched, listed, mapped into `VaultTransaction` type. Simulations that involve interaction with collateral vaults should support `CRVV1ETHSTETH` as well.

### Context

`CRVV1ETHSTETH` is handled differently on the chain. Usually vaults have the id that is handed out by CDP-Manager contract. With this one it is not the case. Which means that there is a different way of interaction with CRVV1ETHSTETH vaults.

### Assets

#479 has added the simulations, #460 adds the support for vaults in the core

### Tasks

- [ ] Investigate how interactions (open , liquidate, etc.) are handled on the chain.
  - [ ] report findings in this issue
- [ ] Propose a patch
- [ ] Implement the patch

KirillDogadin-std:
Closes #484

Checklist:

### DIFF
--- a/core/src/abis/CDP_REGISTRY.json
+++ b/core/src/abis/CDP_REGISTRY.json
@@ -1,0 +1,137 @@
+[
+  {
+    "inputs": [
+      {
+        "internalType": "address",
+        "name": "cdpManager_",
+        "type": "address"
+      }
+    ],
+    "stateMutability": "nonpayable",
+    "type": "constructor"
+  },
+  {
+    "anonymous": false,
+    "inputs": [
+      {
+        "indexed": true,
+        "internalType": "address",
+        "name": "sender",
+        "type": "address"
+      },
+      {
+        "indexed": true,
+        "internalType": "address",
+        "name": "owner",
+        "type": "address"
+      },
+      {
+        "indexed": true,
+        "internalType": "uint256",
+        "name": "cdp",
+        "type": "uint256"
+      }
+    ],
+    "name": "NewCdpRegistered",
+    "type": "event"
+  },
+  {
+    "inputs": [],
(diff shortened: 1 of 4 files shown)


valiafetisov (approved):

KirillDogadin-std:
>  How do we handle vaults that do not have an integer as vaultID like for example:

there're no such vaults possible
the one that you've referenced has the id of 29393
"Wait what?" - ye, i mean i dont have an idea on how they scraped this info from the blockchain but this is obviously some kind of a hacky solution there. They list a few of the first hexes of the urn proxy address instead of an id. Although within the same transaction where this vault has been open the event with the vault id was emitted. https://etherscan.io/tx/0x165bd6973dc4c0ca64e14e68fca0bd4dbdd140160a31988e1202ec8353c2c53b#eventlog

SO i think we're approaching the point where we're shipping the trend-setter product 😎 

___

from the user perspective as of now - we don't care, it's up to them to know the vault id. If they see some hex in the side service - we're not responsible, it's the mess up of the side service.
from the user perspective for the future - user won't care about the id because all the liquidatable vaults will be scraped by us. We will make sure every vault in the database that we maintain has the id set (unlike this link you've sent).

LukSteib (approved):
> there're no such vaults possible
> the one that you've referenced has the id of 29393

I see, thx for clarification on this one. 
Hence LGTM and approved

PR #498, opened 2022-10-13
Add `MarketPriceSelection` component and stories

### LINKED ISSUE #496: Create `MarketPriceSelection` component
### Goal
Modify UI elements to support multiple callees

### Context
As part of our liquidity sources stream we decided to implement multi callee support. For background discussions and overall concept please read #374 

This issue tackles all of the required adjustments to the UI in order to finally enable the new feature

### Tasks
- [x] Modify `Auction` type to contain new keys (see E below for a proposal)
   - [x] note: `marketData` and `suggestedMarketId` are the new keys to be introduced
- [x] Modify fake function to create desired type (note that `unitPrice` can also be `new BigNumber(NaN)`)
- [x] Create `MarketPriceSelection` component as shown in the mock
   - [x] Use animation as in the `BasePanel`
   - [x] Expect `BigNumber.isNaN()` to be true for some markets (so they should display unknown and not be selectable)
   - [x] Use `synced` prop `marketId` to emit and receive currently selected `marketId`
   - [x] Expect that `suggestedMarketId` might change after the component render –> meaning that
      - [x]  in case no selection was made by the user, selected market might change by itself
      - [x] in case a selection was made by the user, it should stay fixed even if `suggestedMarketId` changes
- [x] Send `marketId` (`suggestedMarketId` if none selected) to the `bidWithCallee` action
 
### Assets

Mock:
MakerDAO collateral@2x

E: 

```ts
    marketUnitPrice: BigNumber
    suggestedMarketId: 'Uniswap v3'
    marketData: {
            "Uniswap v3": {
                 unitPrice: BigNumber,
                 route: string[],
            },
            "Uniswap v2": {
                 unitPrice: BigNumber,
                 route: string[],
            },
        }
    },
```

aomafarag:
Closes #496.

Screenshot from 2022-10-13 16-29-53

Checklist:


LukSteib (commented):
- Storybook lgtm ✔️ 
- Checks are failing

LukSteib (dismissed):
Component approved

valiafetisov:
Looking at the screenshot:
- [ ] Please make sure that prices shown in the table are ordered from the best (on top) to the worst (on the bottom) to the `Unknown` (on the very bottom) on the frontend
- [ ] (It might be just storybook, and might not need to be fixed) `Unknown` callee can never be selected (as it seems to be now, since `Market Difference` is Unknown, and both other prices have `Select` in front of them)
- [ ] (It might be just storybook, and might not need to be fixed) In case `Market Difference` is `Unknown`, profits and and `Estimated Profitability Time` will be unknown too

Why `issue checkboxes are all addressed` is not checked? Is there something outstanding?

frontend/components/auction/collateral/MarketPriceSelection.vue
```diff
@@ -82,6 +82,25 @@ export default Vue.extend({
...
+                    sorted.push([callee, this.auctionTransaction.marketData[callee].unitPrice]);
+                } else {
+                    unknown.push([callee, this.auctionTransaction.marketData[callee].unitPrice]);
+                }
+            }
+            sorted.sort((a, b) => a[1].minus(b[1]).toNumber());
+            const calleesArray = [...sorted, ...unknown];
+            const calleesObject = {};
+            calleesArray.forEach(item => (calleesObject[item[0]] = this.auctionTransaction.marketData[item[0]]));
+            return calleesObject;
```
valiafetisov, line 102: There is no reason to convert this back to the object. The problem is the controversy that JS does not guarantee object key order. Suggestion: please return `[...sorted, ...unknown]` from this function and update your template to match

core/src/types.ts
```diff
@@ -21,6 +21,8 @@ export declare interface AuctionInitialInfo {
     isFinished: boolean;
     isRestarting: boolean;
     marketUnitPrice?: BigNumber;
+    suggestedMarketId?: string;
+    marketData?: Record<string, any>;
```
valiafetisov, line 25: Please create a `MarketData` type instead of `any`

frontend/components/auction/collateral/MarketPriceSelection.vue
```diff
@@ -0,0 +1,118 @@
...
+        suggestionOrSelection(): string | undefined {
+            return this.marketId || this.auctionTransaction.suggestedMarketId;
+        },
+        marketUnitPrice(): BigNumber | undefined {
+            if (this.auctionTransaction.marketData && this.suggestionOrSelection) {
+                return this.auctionTransaction.marketData[this.suggestionOrSelection].unitPrice;
+            }
+            return undefined;
+        },
+        callees(): Object {
```
valiafetisov, line 85: Please use better typing, eg `Array<MarketData>`

frontend/components/auction/collateral/CollateralAuctionSwapTransactionTable.vue
```diff
@@ -21,16 +21,7 @@
...
-            <div>Price On Uniswap</div>
-            <div class="RightInfo">
-                <template v-if="auctionTransaction.isActive && auctionTransaction.marketUnitPrice">
-                    <FormatCurrency :value="auctionTransaction.marketUnitPrice" currency="DAI" /> per
-                    <span class="uppercase">{{ auctionTransaction.collateralSymbol }}</span>
-                </template>
-                <span v-else class="opacity-50">Unknown</span>
-            </div>
-        </div>
+        <MarketPriceSelection :auction-transaction="auctionTransaction" :market-id.sync="marketId" />
```
valiafetisov, line 24: As you raised in the issue and I answered: the values underneath this row is dependent on the selected `marketId`. They all now taken from the `marketData[marketId]` object: so that switching `marketId` displays new values

LukSteib (changes requested):
Maybe storybook related but to re-iterate on point @valiafetisov mentioned in prev comment

- [ ] when expanding component for the first time both selectable callees have "Select" and the uni v3 which is actually selected is not marked accordingly ("Selected")
- [ ] Why is the uni v3 callee the default selection (see screenshot below) if the 1inch callee has a better marketUnitPrice?


image

LukSteib:
Thanks for clarification! Makes sense and no need to write a dedicated function for storybook.

LukSteib (dismissed):

LukSteib:
> @LukSteib, regarding the problem you mentioned with the route. The collateral configuration of wstETH on Curve V3 has an empty route array, i.e. it gets swapped directly into DAI without passing by ETH.

I think the problem is not related to `WSTETH` since same problem occurs for a normal `ETH` auction. However, this is details that we might fix in a follow up issue. 

> Currently, I get this error in the browser console whenever I select a callee, which is self-explanatory and reflects the issue we are having:

Get the same error which prevents switching the callee via the UI. This problem however is blocking and I won't approve this PR as I don't want to merge and release a broken component.

frontend/store/auctions.ts
```diff
@@ -239,6 +239,7 @@ export const actions = {
             alternativeDestinationAddress,
         }: { id: string; marketId: string; alternativeDestinationAddress: string | undefined }
     ) {
+        console.info(marketId); // to be removed after core implementation
```
valiafetisov, line 242: Please remove. Also, in the future please prepend comments to yourself with `TODO:` (then it can be easily found)

frontend/components/auction/collateral/MarketPriceSelection.vue
```diff
@@ -0,0 +1,117 @@
...
+        marketUnitPrice(): BigNumber | undefined {
+            if (this.auctionTransaction.marketDataRecords && this.suggestionOrSelection) {
+                return this.auctionTransaction.marketDataRecords[this.suggestionOrSelection].marketUnitPrice;
+            }
+            return undefined;
+        },
+        callees() {
+            const sorted = [];
+            const unknown = [];
+            for (const callee in this.auctionTransaction.marketDataRecords) {
```
valiafetisov, line 90: How does this even pass typescript tests without the types? What exactly does this function returns? I can't figure it out from the names `callee`, `a`, `b`, `callee[0]`, `callee[1]`

Didn't you already have better (readable) logic implemented in the `getBestMarketId`?

LukSteib (changes requested):
- [ ] (note that it should not block if we cannot figure out easily): On initial load of the transaction page none of the callees is indicated as "Selected". Only after first selecting one of them the indicator holds (see s1 below)
- [ ] One more thing that makes me wonder: In the example below (`ETH-A` simulation) we preselect uniswapv2 as the market price we determine is lower (fine). However when comparing gross and net profit we indicate that uniswapv3 would yield overall better outcome for the user. Just trying to understand how this is possible. (see comparison in s2 below)

s1
image

s2
image

valiafetisov:
> This was answered [above](https://github.com/sidestream-tech/unified-auctions-ui/pull/498#issuecomment-1286601589)

Issue description doesn't imply that no price should be set as selected in the UI. I think we still need to indicate which one is currently "best" or "selected"

frontend/components/auction/collateral/MarketPriceSelection.vue
```diff
@@ -0,0 +1,118 @@
...
+                            <td class="pr-2 whitespace-nowrap">
+                                <span v-for="currency in callee[1].route" :key="currency"
+                                    >{{ currency }} &#8594;
+                                </span>
+                                DAI
+                            </td>
+                            <td class="w-full text-right whitespace-nowrap">
+                                <div v-if="callee[1].marketUnitPrice && !callee[1].marketUnitPrice.isNaN()">
+                                    <button type="button" @click="$emit('update:marketId', callee[0])">
+                                        <span v-if="marketId === callee[0]" class="opacity-50">Selected</span>
```
valiafetisov, line 30:
```suggestion
                                        <span v-if="suggestionOrSelection === callee[0]" class="opacity-50">Selected</span>
```
?

valiafetisov (approved):

Issue #501, opened 2022-10-18
Outline current state of the collateral simulations


valiafetisov:
### Goal

List of supported and unsupported collaterals, list of potential fixes

### Context

Since we're working on the collateral auctions simulations for quite some time already and faced different kinds of problems, we need to summarise somewhere its state and come up with a list of solutions to move forward and prioritise them. Therefore, let's summarise it here or even create/modify test script that will do it for us.

### Assets

Issues where simulations were previously developed and improved:
- Original issue with the approach to modify VAT balance and call drop https://github.com/sidestream-tech/unified-auctions-ui/issues/464
- Support for fetching of `CRVV1ETHSTETH-*` vaults https://github.com/sidestream-tech/unified-auctions-ui/issues/484
- Directly modify wallet balance instead of VAT balance https://github.com/sidestream-tech/unified-auctions-ui/issues/495
- Skip `YFI-A` testing https://github.com/sidestream-tech/unified-auctions-ui/pull/500

Previous state: https://github.com/sidestream-tech/unified-auctions-ui/pull/479#issuecomment-1270594143

### Tasks

- [x] Outline current state of the simulations (via a script so it can be run periodically or by hand)
  - [x] Post a single comment with a table `collateral type - state - reason`
- [x] Outline solutions / link existing issues that were created to overcome known limitations
- [ ] Create new issues, prioritise work
- [ ] Keep track of the progress in this issue until all collaterals are supported

valiafetisov:
| Collateral type   | State           | Reason |
|:------------------|:----------------|:-------|
| RETH-*            | ❌ not supported | Join contract does not have enough collateral. Will be addressed in https://github.com/sidestream-tech/unified-auctions-ui/issues/495 |
| CRVV1ETHSTETH-A   | ❌ not supported | ? |
| UNIV2DAIUSDC-A    | ❌ not supported | `Dog/liquidation-limit-hit` Liquidation limit too high |
| WSTETH-B          | ❌ not supported | `Dog/not-unsafe`: Successful vault creation, but stability fees did not change after time warp |
| YFI-A             | ⚠️ not testible  | `Vat/ceiling-exceeded`: Error during vault opening https://github.com/sidestream-tech/unified-auctions-ui/pull/500 |
| AAVE-A            | ⚠️ not testible  | `MCD_VAT.ilks().line == 0`: Max debt set to zero |
[...]

PR #510, opened 2022-10-25
Crveth simulation

### LINKED ISSUE #503: Feat: add vault opening simulation for CRVV1ETHSTETH
### Goal

Add functinality to enable simulation of CRVV1ETHSTETH vaults

### Context

Currently the simulations for the mentioned vault type are disabled because the mechanism for opening the vaults is different.

we can add support to `PROXY_ACTIONS_DSR` type contracts (which also is the [cropper](https://etherscan.io/address/0xa2f69f8b9b341cfe9bfbb3aab5fe116c89c95baf#code)):
  - add the corresponding key into the `COLLATERALS` config that specifies: `contractType: Standard | Proxied` where

```ts
interface Standard {
  type: 'standard'
}

interface Proxied {
  type: 'proxied';
  proxyName: string;
}
```

`Standard` logic executes the alrady existing code.
`Proxied` logic:
  1. calls `build` of [PROXY_FACTORY](https://etherscan.io/address/0xa26e15c895efc0616177b7c1e7270a4c7d51c997#writeContract) contract, creates the proxy and sets the `HARDHAT_PUBLIC_KEY` as owner.
  2. calls the function `execute(address _target, bytes _data)` of the created proxy. extracts the proxy from the chainlog - in this case [cropper](https://etherscan.io/address/0xa2f69f8b9b341cfe9bfbb3aab5fe116c89c95baf#code). Passes it as target. Passes `_data` where each point takes 32 bytes:
   - method id (`utils.id('transfer(address,uint256)').substring(0, 10)`). E.g. `openLockGemAndDraw(address,address,address,bytes32,uint256,uint256)` is converted into `0xff9a89c6` via https://emn178.github.io/online-tools/keccak_256.html
   - ...all the other args needed to call the method

the (2) from above is assembled via a helper function
the function to be called in the crveth case is `openLockGemAndDraw`. This immediately should result in debt drawn

example transaction: https://etherscan.io/tx/0x489d3cfc27f648dc2e79866b4a60bc8f57b836dbdd9ee5ce02dcb294d787cb6b

### Assets

Overview issue #501

### Tasks

- [ ] Discuss the proposed implementation
- [ ] Discover edgecases / questions and discuss them as well
- [ ] Implement vault opening mechanism
- [ ] remove the type from unsupported collaterals

KirillDogadin-std:
Closes #503 

Checklist:

### DIFF
--- a/core/helpers/hex.ts
+++ b/core/helpers/hex.ts
@@ -34,8 +34,3 @@ export const postpendZeros = (val: string, numberOf32ByteSegments: number) => {
     }
     return `${val}${new Array(dataLength - val.length).join('0')}`;
 };
-
-export const argsToData = (...args: [string, number][]) => {
-    const bytes = args.map(([arg, segments]) => postpendZeros(arg, segments).substring(2)).join('');
-    return `0x${bytes}`;
-};
--- a/core/helpers/hex.ts
+++ b/core/helpers/hex.ts
@@ -26,11 +26,3 @@ export const roundDownToFirstSignificantDecimal = (number: BigNumber) => {
 export const getMethodSignature = (methodString: string) => {
     return ethers.utils.id(methodString).substring(0, 10);
 };
-
-export const postpendZeros = (val: string, numberOf32ByteSegments: number) => {
-    const dataLength = numberOf32ByteSegments * 32 + 2;
-    if (val.length > dataLength) {
-        throw new Error(`String has more than ${dataLength} decimals: ${val}`);
-    }
-    return `${val}${new Array(dataLength - val.length).join('0')}`;
-};
--- a/core/helpers/hex.ts
+++ b/core/helpers/hex.ts
@@ -22,3 +22,20 @@ export const roundDownToFirstSignificantDecimal = (number: BigNumber) => {
     const power = Math.abs(number.e || 0);
     return new BigNumber(number.toFixed(power + 1, BigNumber.ROUND_DOWN));
 };
+
+export const getMethodSignature = (methodString: string) => {
+    return ethers.utils.id(methodString).substring(0, 10);
+};
+
+export const postpendZeros = (val: string, numberOf32ByteSegments: number) => {
+    const dataLength = numberOf32ByteSegments * 32 + 2;
+    if (val.length > dataLength) {
+        throw new Error(`String has more than ${dataLength} decimals: ${val}`);
+    }
+    return `${val}${new Array(dataLength - val.length).join('0')}`;
+};
+
+export const argsToData = (...args: [string, number][]) => {
+    const bytes = args.map(([arg, segments]) => postpendZeros(arg, segments).substring(2)).join('');
+    return `0x${bytes}`;
+};
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -30,7 +30,7 @@ import { getAllCollateralTypes } from '../../src/constants/COLLATERALS';
 import { createProxy } from '../../src/proxy';
 import { giveAllowanceToAddress } from '../../src/authorizations';
 import getProvider from '../../src/provider';
-import detectProxyTarget from 'evm-proxy-detection';
+import detectProxyTarget from '../../helpers/detectProxyTarget';
 
 const UNSUPPORTED_COLLATERAL_TYPES = [
     'UNIV2DAIUSDC-A', // Liquidation limit too high (fails with "Dog/liquidation-limit-hit")
@@ -217,10 +217,7 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
         TEST_NETWORK,
         getJoinNameByCollateralType(collateralType)
     );
-    const provider = await getProvider(TEST_NETWORK);
-    const proxyTarget = await detectProxyTarget(joinContractAddress, ({ method, params }) =>
-        provider.send(method, params)
-    );
+    const proxyTarget = await detectProxyTarget(TEST_NETWORK, joinContractAddress)
     if (!proxyTarget) {
         return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
     }
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -216,7 +216,7 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
         TEST_NETWORK,
         getJoinNameByCollateralType(collateralType)
     );
-    const proxyTarget = await detectProxyTarget(TEST_NETWORK, joinContractAddress)
+    const proxyTarget = await detectProxyTarget(TEST_NETWORK, joinContractAddress);
     if (!proxyTarget) {
         return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
     }
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -169,10 +169,10 @@ const giveJoinContractAllowance = async (collateralConfig: CollateralConfig, amo
 
 const createProxiedVaultWithCollateral = async (collateralType: CollateralType, collateralOwned: BigNumber) => {
     const { minUnitPrice, stabilityFeeRate } = await fetchVaultCollateralParameters(TEST_NETWORK, collateralType);
-    const collateralSymbol = getCollateralConfigByType(collateralType).symbol
+    const collateralSymbol = getCollateralConfigByType(collateralType).symbol;
     const drawnDebtExact = collateralOwned.multipliedBy(minUnitPrice).dividedBy(stabilityFeeRate);
     const drawnDebt = roundDownToFirstSignificantDecimal(drawnDebtExact);
-    const proxyAddress = await createProxy(TEST_NETWORK, HARDHAT_PUBLIC_KEY)
+    const proxyAddress = await createProxy(TEST_NETWORK, HARDHAT_PUBLIC_KEY);
     await giveAllowanceToAddress(TEST_NETWORK, collateralSymbol, proxyAddress, collateralOwned);
     await openVaultWithProxiedContractAndDrawDebt(
         TEST_NETWORK,
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -15,6 +15,7 @@ import {
     getContractAddressByName,
     getErc20Contract,
     getJoinNameByCollateralType,
+    getClipperNameByCollateralType,
 } from '../../src/contracts';
 import {
     depositCollateralToVat,
@@ -27,11 +28,16 @@ import { CollateralConfig, CollateralType } from '../../src/types';
 import { roundDownToFirstSignificantDecimal, roundUpToFirstSignificantDecimal } from '../../helpers/hex';
 import { determineBalanceSlot, setCollateralInWallet } from '../../helpers/hardhat/erc20';
 import { getAllCollateralTypes } from '../../src/constants/COLLATERALS';
+import { grantAdminPrivelegeForContract } from '../../helpers/hardhat/slotOverwrite';
+import {
+    allowAllActionsInClipperContract,
+    setCollateralDebtCeilingToGlobal,
+    setCollateralLiquidationLimitToGlobal,
+} from '../../helpers/hardhat/contractParametrization';
+import { overwriteStabilityFeeAccumulationRate } from '../../helpers/hardhat/overwrites';
 import detectProxyTarget from '../../helpers/detectProxyTarget';
 
 const UNSUPPORTED_COLLATERAL_TYPES = [
-    'UNIV2DAIUSDC-A', // Liquidation limit too high (fails with "Dog/liquidation-limit-hit")
-    'WSTETH-B', // Does not accumulate stability fee rate at all
     'RETH-A', // [temporary] this collateral is not yet deployed, tested via different flow
 ];
 
@@ -83,7 +89,7 @@ const ensureWalletBalance = async (collateralConfig: CollateralConfig, collatera
         HARDHAT_PUBLIC_KEY,
         collateralConfig.decimals
     );
-    if (!balance.eq(collateralOwned)) {
+    if (balance.minus(collateralOwned).shiftedBy(collateralConfig.decimals).abs().toFixed(0) !== '0') {
         throw new Error(
             `Unexpected wallet balance. Expected ${collateralOwned.toFixed()}, Actual ${balance.toFixed()}`
         );
@@ -210,6 +216,18 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
         return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
     }
     return await createProxiedVaultWithCollateral(collateralType, collateralOwned);
+};
+
+export const adjustLimitsAndRates = async (collateralType: CollateralType) => {
+    await grantAdminPrivelegeForContract('MCD_VAT');
+    await grantAdminPrivelegeForContract('MCD_DOG');
+    const clipper = getClipperNameByCollateralType(collateralType);
+    await grantAdminPrivelegeForContract(clipper);
+
+    await setCollateralLiquidationLimitToGlobal(collateralType);
+    await setCollateralDebtCeilingToGlobal(collateralType);
+    await allowAllActionsInClipperContract(collateralType);
+    await overwriteStabilityFeeAccumulationRate(collateralType, new BigNumber(1.0000000005));
 };
 
 export default createVaultWithCollateral;
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -222,7 +222,7 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
     try {
         return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
     } catch (e) {
-        if (e instanceof Error) {
+        if (e instanceof Error && e.message.includes('Error: Transaction reverted without a reason string')) {
             console.warn(
                 `Failed creating the vault with default settings. Falling back to creating via DS-PROXY: ${e.message}`
             );
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -219,12 +219,18 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
     }
     await ensureWalletBalance(collateralConfig, collateralOwned);
 
-    if (collateralConfig.joinContractType.type === 'default') {
-        return createDefaultVaultWithCollateral(collateralType, collateralOwned);
+    try {
+        return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
+    } catch (e) {
+        if (e instanceof Error) {
+            console.warn(
+                `Failed creating the vault with default settings. Falling back to creating via DS-PROXY: ${e.message}`
+            );
+        } else {
+            throw e;
+        }
     }
-    if (collateralConfig.joinContractType.type === 'proxied') {
-        return createProxiedVaultWithCollateral(collateralType, collateralOwned);
-    }
+    return await createProxiedVaultWithCollateral(collateralType, collateralOwned);
 };
 
 export default createVaultWithCollateral;
--- a/core/simulations/helpers/createVaultWithCollateral.ts
+++ b/core/simulations/helpers/createVaultWithCollateral.ts
@@ -27,8 +27,6 @@ import { CollateralConfig, CollateralType } from '../../src/types';
 import { roundDownToFirstSignificantDecimal, roundUpToFirstSignificantDecimal } from '../../helpers/hex';
 import { determineBalanceSlot, setCollateralInWallet } from '../../helpers/hardhat/erc20';
 import { getAllCollateralTypes } from '../../src/constants/COLLATERALS';
-import { createProxy } from '../../src/proxy';
-import { giveAllowanceToAddress } from '../../src/authorizations';
 import detectProxyTarget from '../../helpers/detectProxyTarget';
 
 const UNSUPPORTED_COLLATERAL_TYPES = [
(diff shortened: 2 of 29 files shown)


LukSteib (changes requested):
- [ ] NIT: why do we print the "balance..." line twice (see snippet from terminal below)
- [ ] If not much overhead: Can we print the `vaultID` after the first step of the simulation ("Create the vault")
- [ ] In general: Can we re-name this type of simulation from `Simulate liquidation Auctions` to `Simulate collateral auction per collateral type` 

Rest is looking good and working as expected (was able to see and participate in auction 👍 )


snippet from terminal: 

```sh
Minimum collateral amount to open vault: 27.027
Balance slot is 0x2, language format is vyper, contract address is 0x06325440D014e39736583c165C2963BA99fAf14E
Balance slot is 0x2, language format is vyper, contract address is 0x06325440D014e39736583c165C2963BA99fAf14E
```

LukSteib (dismissed):

core/src/constants/COLLATERALS.ts
```diff
@@ -325,6 +405,11 @@ const COLLATERALS: Record<string, CollateralConfig> = {
             route: ['ETH'],
         },
         oracle: CONFIG_WITH_NEXT_PRICE,
+        joinContractType: {
+            type: 'proxied',
+            proxyType: 'CROPPER',
+            joinContractName: 'CRVV1ETHSTETH_A',
+        },
```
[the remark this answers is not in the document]
KirillDogadin-std: - evidence:
image

> joinContractName

- now using function

>  Can't you just make cropper proxy a fallback?

ok, makes sense.

core/src/vaults.ts
```diff
@@ -381,6 +385,40 @@ const getLatestVault = async (network: string, walletAddress: string) => {
...
+    debtAmountDai: BigNumber
+) => {
+    const config = getCollateralConfigByType(collateralType);
+    if (config.joinContractType.type !== 'proxied') {
+        throw new Error(`Can't invoke this method for the collateral ${collateralType}`);
+    }
+    giveAllowanceToAddress(network, config.symbol, proxyAddress, collateralAmount);
+
+    const signer = await getSigner(network);
+    const proxyContract = new ethers.Contract(proxyAddress, DS_PROXY, signer);
```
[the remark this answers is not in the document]
KirillDogadin-std: I need to address the specific contract that has been freshly deployed and that only I can use. There is no name to the created contract, there is only the address. 

Do you suggest instead to create the `executeTransactionAtAddress`?

core/helpers/hex.ts
```diff
@@ -22,3 +22,7 @@ export const roundDownToFirstSignificantDecimal = (number: BigNumber) => {
     const power = Math.abs(number.e || 0);
     return new BigNumber(number.toFixed(power + 1, BigNumber.ROUND_DOWN));
 };
+
+export const getMethodSignature = (methodString: string) => {
+    return ethers.utils.id(methodString).substring(0, 10);
```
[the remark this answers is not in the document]
KirillDogadin-std: As function name states - it creates a method signature. If you have a better way of extracting this - please suggest.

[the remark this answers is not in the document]
KirillDogadin-std: issue header also contains instructions on how it's done manually

core/src/vaults.ts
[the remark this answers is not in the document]
valiafetisov: > freshly deployed

Missed that point, will re-review

core/helpers/hex.ts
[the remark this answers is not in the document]
valiafetisov: > it creates a method signature

I was asking about `.substring(0, 10)` part, not the whole thing. Why do you need to cut out first 8 hex numbers? What all other bytes do? If you don't know what's going on here, where did you copied it from?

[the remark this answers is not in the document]
KirillDogadin-std: see the example documentatino here https://docs.soliditylang.org/en/v0.5.3/abi-spec.html#examples - cmd+f for methodId

core/simulations/helpers/createVaultWithCollateral.ts
```diff
@@ -186,4 +207,30 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
...
+        // fallback to setting vat balance and withdrawing it
+        await setAndCheckCollateralInVat(collateralType, collateralOwned);
+        await checkAndWithdrawCollateralFromVat(collateralConfig, collateralOwned);
+    }
+    await ensureWalletBalance(collateralConfig, collateralOwned);
+
+    try {
+        return await createDefaultVaultWithCollateral(collateralType, collateralOwned);
+    } catch (e) {
+        if (e instanceof Error && e.message.includes('Error: Transaction reverted without a reason string')) {
```
valiafetisov, line 225: This error is very generic. I wonder if there is way to actually know that this collateral requires proxy eg `join.implementation` public value exist on the proxy-like contracts (since it's the address of the actual join contract it seems)? Then, you can just call `join.implementation` before trying to create default vault and not try/catch at all?

core/src/proxy.ts
```diff
@@ -0,0 +1,13 @@
+import getContract from './contracts';
+
+export const createProxy = async (network: string, proxyOwnerAddress: string) => {
+    const proxyFactoryContract = await getContract(network, 'PROXY_FACTORY', true);
+    await proxyFactoryContract['build(address)'](proxyOwnerAddress);
```
valiafetisov, line 5: Wouldn't this already return `proxy` as [per the contract code](https://etherscan.io/address/0xa26e15c895efc0616177b7c1e7270a4c7d51c997#code):
```
function build(address owner) public returns (DSProxy proxy) 
```

core/simulations/helpers/createVaultWithCollateral.ts
```diff
@@ -160,18 +167,32 @@ const giveJoinContractAllowance = async (collateralConfig: CollateralConfig, amo
...
+        TEST_NETWORK,
+        proxyAddress,
+        collateralType,
+        collateralOwned,
+        drawnDebt
+    );
+    const registry = await getContract(TEST_NETWORK, 'CDP_REGISTRY');
+    const filter = registry.filters.NewCdpRegistered(null, proxyAddress, null);
+    const vaultOpenEvents = await registry.queryFilter(filter);
+    const vaultId = vaultOpenEvents[vaultOpenEvents.length - 1].args?.cdp;
```
valiafetisov, line 187: Same here. Wouldn't `vaultId` (`cdpi` in maker terms) be returned by the open method from the `CDP_REGISTRY.open` call?

In either way, please move the code to determine opened vault into `openVaultWithProxiedContractAndDrawDebt` so it will return this number, just like `createDefaultVaultWithCollateral` does it

core/src/vaults.ts
```diff
@@ -381,6 +389,41 @@ const getLatestVault = async (network: string, walletAddress: string) => {
     return new BigNumber(lastHex._hex).toNumber();
 };
 
+export const openVaultWithProxiedContractAndDrawDebt = async (
+    network: string,
+    proxyAddress: string,
```
valiafetisov, line 394: Can you open a vault with proxied contract without creating a proxy upfront? Then should `createProxy` be moved into this function and not used outside? (so that we get clear API of this function)

core/simulations/helpers/createVaultWithCollateral.ts
valiafetisov: By the way, why don't we never use `implementation` address anywhere? Isn't it the desired contract already with which we can open a vault?

core/src/proxy.ts
KirillDogadin-std: The return value of the function in solidity is only available on chain. Exceptions are view functions / constants. even if `build` returns the proxy, it is not a constant or a view = we have to extract it differently.

core/simulations/helpers/createVaultWithCollateral.ts
KirillDogadin-std: > Same here

as above.

core/src/vaults.ts
KirillDogadin-std: No and no.

This would force creation of the new proxy each time when the vault is opened by our functionality.
This is undesired because single DSProxy instance is sufficient (reusable), can be used for variety of purposes and not only for the current action we're performing.

core/simulations/helpers/createVaultWithCollateral.ts
KirillDogadin-std: Using the library now.

> Isn't it the desired contract already with which we can open a vault?

no, here as far as i understand we're talking about the join contract. Join contract is responsible for holding funds and letting VAT know that the user has deposited the funds.

On top of that, neither of our ABIs contains the declaration of the `implementation` const (aka claims that no such const is present). The abi that we store for `join` of the crveth under the name `MCD_JOIN_CRVV1ETHSTETH_A` is in fact the abi of this contract's proxy. This allows us to know which join function we have access to in the implementation through the proxy, but does not inform us about the proxy's functions.

valiafetisov: > On top of that, neither of our ABIs contains the declaration of the implementation const

This is a limitation of our codebase, not a technical problem. You can either pass a different ABI in the `getContractInterfaceByName` for this contract (isolating away the detection of the contract type) to somewhere else; or build a simple human-readable ABI for this purpose yourself (eg `["function implementation() view returns (address)"]`). But anyway, you already use a library for this (to get the `proxyTarget` address)

> as far as i understand we're talking about the join contract

As opposite to what contract? I will repeat my question. When you find `proxyTarget` can't you just use it instead of creating a new a new proxy? Can you actually try it?

core/src/proxy.ts
valiafetisov: Well, you're right, but according to https://ethereum.stackexchange.com/a/88122 there are better ways to do it anyway: Instead of building a new filter and fetching the last event, use executed transaction hash to fetch transaction receipt + events emitted with it. Maybe you can even make a helper from it? eg `getTransactionEvents`?

core/simulations/helpers/createVaultWithCollateral.ts
valiafetisov: See my suggestion above

core/src/vaults.ts
valiafetisov: > This would force creation of the new proxy each time when the vault is opened by our functionality.

It wouldn't if it will be optional, eg `proxyAddress?: string`. Anyway, please see my other comment about proxy creation https://github.com/sidestream-tech/unified-auctions-ui/pull/510#discussion_r1006737780

core/simulations/helpers/createVaultWithCollateral.ts
KirillDogadin-std: > When you find proxyTarget can't you just use it instead of creating a new a new proxy?

the found proxy target is the JOIN contract.

In order to create the vault you require DS-PROXY contract. 

These 2 are used for different purpose.

see https://github.com/sidestream-tech/unified-auctions-ui/issues/501#issuecomment-1293491292 with overview of what-proxy-for-what.

core/src/vaults.ts
KirillDogadin-std: i don't see the point. As this function is the part of `src` (not specifically dedicated to simulations), i would have to instead provide the proxyOwner address - then there're 2 mutually exclusive arguments. 

Should i then move this func into simulation suite for the specific use exclusively inside it?

core/simulations/helpers/createVaultWithCollateral.ts
```diff
@@ -186,4 +201,30 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
...
 
+const createVaultWithCollateral = async (collateralType: CollateralType, collateralOwned: BigNumber) => {
+    const collateralConfig = getCollateralConfigByType(collateralType);
+    const [balanceSlot, languageFormat] = await determineBalanceSlot(collateralType);
+    if (balanceSlot && languageFormat) {
+        await setCollateralInWallet(collateralConfig.symbol, collateralOwned);
+    } else {
+        // fallback to setting vat balance and withdrawing it
+        await setAndCheckCollateralInVat(collateralType, collateralOwned);
+        await checkAndWithdrawCollateralFromVat(collateralConfig, collateralOwned);
```
valiafetisov, line 212: Do we still need the fallback for RENBTC?

core/simulations/helpers/createVaultWithCollateral.ts
```diff
@@ -186,4 +201,30 @@ const createVaultWithCollateral = async (collateralType: CollateralType, collate
...
+    await ensureWalletBalance(collateralConfig, collateralOwned);
+
+    const joinContractAddress = await getContractAddressByName(
+        TEST_NETWORK,
+        getJoinNameByCollateralType(collateralType)
+    );
+    const provider = await getProvider(TEST_NETWORK);
+    const proxyTarget = await detectProxyTarget(joinContractAddress, ({ method, params }) =>
+        provider.send(method, params)
+    );
```
valiafetisov, line 223: Can you please make a reusable helper for it? eg

```
detectProxyTarget(network, contractAddress)

File: core/src/authorizations.ts [hunk 10]
```diff
@@ -180,3 +181,15 @@ export const getDebtAuctionAuthorizationStatus = async (network: string, walletA
...
+export const giveAllowanceToAddress = async (
+    network: string,
+    collateralSymbol: string,
+    spenderAddress: string,
+    collateralAmount: BigNumber
+) => {
+    const config = getCollateralConfigBySymbol(collateralSymbol);
+    const tokenAddress = await getContractAddressByName(network, collateralSymbol);
+    const tokenContract = await getErc20Contract(network, tokenAddress, true);
+    await tokenContract.approve(spenderAddress, collateralAmount.shiftedBy(config.decimals).toFixed());
```
valiafetisov, line 194: Before sending amounts to a chain, you always need to have integer-like string without floating points. In case `collateralAmount` was calculated (multiplied/divided) it might have more precision than required, so it's safer to:
```suggestion
    await tokenContract.approve(spenderAddress, collateralAmount.shiftedBy(config.decimals).toFixed(0));
```

core/src/helpers/extractEventFromTransaction.ts
```diff
@@ -0,0 +1,11 @@
+import getProvider from '../provider';
+import { ethers } from 'ethers';
+
+const extractEventFromTransaction = async (network: string, transactionHash: string, event: string) => {
+    const provider = await getProvider(network);
+    const receipt = await provider.getTransactionReceipt(transactionHash);
```
valiafetisov, line 6: This will fetch logs unparsed, instead, you can wait for the transaction to get the events parsed according to ABI via [`transaction.wait`](https://docs.ethers.io/v5/api/providers/types/#providers-TransactionReceipt). Or, which will be more cumbersome: `tx.logs.map((log) => contract.interface.parseLog(log))`.

(parsing will not only help with automatically parsed event names, but also with automatically parsed values, so you don't have to use `ethers.utils.defaultAbiCoder.decode` in the parent)

Therefore I suggest to pass transaction here instead of the hash, wait for it and return list of events

core/simulations/helpers/createVaultWithCollateral.ts
KirillDogadin-std: Yes, this is different issue that makes us disable RENBTC

core/src/authorizations.ts
KirillDogadin-std: having precision incorrectly handled on the previous steps results in the error (incorrect argument) here if i remember correctly. And from my perspective it's desired to have the error raised to fix it asap. At least from my experience: having a couple of toFixed(0) where not actually needed already resulted into some time spent debugging

valiafetisov: > And from my perspective it's desired to have the error raised to fix it asap

I'm just saying that calling this function with specific _appropriate_ parameters will fail. You usually don't want your code to be more fragile to help you with debugging. Also, if this code will be called outside tests, no one would knew your intention to make it fragile. Therefore, please handle this edge case properly and prefer `toFixed(0)` over `.toFixed()` when sending requests to blockchain in the future.

core/src/helpers/extractEventFromTransaction.ts
valiafetisov: Well, you're right that my suggestion wouldn't improve some particular edge-case, but I believe it would still return you an object with all events, even the ones that wasn't parsed. Also, if you wish to improve this particular edge case as well, you can add optional `contractInterfaces` argument to supply other abis and parse it manually via `tx.logs.map((log) => contract.interface.parseLog(log))` provided above. I just suggest to make it nice and reusable, not just hack/hardcode it every time you need to parse a transaction event.

core/src/vaults.ts
```diff
@@ -381,6 +390,51 @@ const getLatestVault = async (network: string, walletAddress: string) => {
...
+    const transaction = await proxyContract['execute(address,bytes)'](target, transactionData);
+    const events = await extractEventFromTransaction(
+        network,
+        transaction.hash,
+        'NewCdpRegistered(address,address,uint256)'
+    );
+    if (events.length !== 1) {
+        throw new Error('Unexpected number of NewCdpRegistered events');
+    }
+    const vaultId = new BigNumber(events[0].topics[3]).toNumber();
```
valiafetisov, line 434: This is again a fragile code. I am not sure that order of the events or topics is guaranteed by the blockchain and/or by the ethers, please follow the suggestion here https://github.com/sidestream-tech/unified-auctions-ui/pull/510#discussion_r1009469950

core/src/proxy.ts
```diff
@@ -0,0 +1,18 @@
...
+    const transaction = await proxyFactoryContract['build(address)'](proxyOwnerAddress);
+    const events = await extractEventFromTransaction(
+        network,
+        transaction.hash,
+        'Created(address,address,address,address)'
+    );
+    if (events.length !== 1) {
+        throw new Error("Undexpected number of 'Created' events");
+    }
+    const proxyAddress = ethers.utils.defaultAbiCoder.decode(['address', 'address'], events[0].data)[0];
```
valiafetisov, line 16: Same here! You've already ABI of the event (`'Created(address,address,address,address)'`) why don't you use it inside `extractEventFromTransaction` to parse returned values, but have to hardcode it again?

core/src/authorizations.ts
KirillDogadin-std: - Well, my point is derived from the axiom where the parameters that you would like to support with `(0)` are inappropriate:
  - it implicitly changes the provided value if it is incorrect
  - it is an approval contract: we're trying to approve an amount that does not match the input.
- As far as i am concerned this actually "declines" incorrect inputs. Code should fail on incorrect inputs / be fragile enough to crash & burn on incorrect inputs.

If the function is called outside the tests with inappropriate arguments in `()` case - it will fail because the input is not correct. In the `(0)` noone will know that it changes the allowance amount without explicit intent of the function user

Introduced the change though.

valiafetisov: > As far as i am concerned this actually "declines" incorrect inputs

See my initial comment

> In case collateralAmount was calculated (multiplied/divided) it might have more precision than required

Let's say for some reason, you need to get the approval for the 1 ETH divided by 3. There is better edge-case that actually already happen to me, the precision of DAI stored in your VAT is [RAD (ie 45 decimal points)](https://github.com/makerdao/dss/blob/fa4f6630afb0624d04a003e920b0d71a00331d98/src/vat.sol#L59) while the precision of DAI token itself is only 18. So it you need to authorize some number taken from VAT, you will end up with the precision grater than 18 and the transaction will fail. Is this enough to convince you?

core/src/vaults.ts
```diff
@@ -381,6 +391,51 @@ const getLatestVault = async (network: string, walletAddress: string) => {
...
+    const typesArray = ['address', 'address', 'address', 'bytes32', 'uint256', 'uint256'];
+    const encodedArgs = ethers.utils.defaultAbiCoder.encode(typesArray, args);
+    const transactionData = method + encodedArgs.substring(2);
+    const target = (await getContract(network, `PROXY_ACTIONS_${proxyType}`)).address;
+    const transaction = await proxyContract['execute(address,bytes)'](target, transactionData);
+    const events = await extractEventFromTransaction(
+        transaction,
+        'NewCdpRegistered(address,address,uint256)',
+        new ethers.utils.Interface(CDP_REGISTRY)
+    );
```
valiafetisov, line 431: You basically provide ABI 3 times: 
1. It comes with `transaction`
2. The `NewCdpRegistered(address,address,uint256)` is the ABI of the event you need to get
3. The interface itself `new ethers.utils.Interface(CDP_REGISTRY)`

Initially I suggested to use ABI that is provided with transaction, but it turns out it's not feasible due to the key event being emitted by the other contract. So instead you can use `'NewCdpRegistered(address,address,uint256)'` to create interface and pass it to decode specific event that you need.

You might even replace second parameter with `NewCdpRegistered(address sender,address owner,uint256 cdp)` to retrieve values under the named key `args.cdp`. In other words:

```ts
const events = await extractEventFromTransaction(
    transaction,
    'NewCdpRegistered(address sender, address owner, uint256 cdp)'
);
```
then, inside `extractEventFromTransaction`
```
const contractInterface = new ethers.utils.Interface('NewCdpRegistered(address sender, address owner, uint256 cdp)')
```
While keeping the rest of the code the same


Nit: the function should be called `extractEventsFromTransaction` as it returns an array and might make `eventABI` parameter optional

core/src/authorizations.ts
valiafetisov: This comment was not addressed

core/src/vaults.ts
KirillDogadin-std: This suggestion gets the same problem as https://github.com/ethers-io/ethers.js/issues/1233 where incomplete interface specification results in problems with parsing the transaction logs. Hence i resided with providing signature and complete abi separately. Do you have the better suggestion than `try-catching`?

core/src/helpers/extractEventFromTransaction.ts
```diff
@@ -0,0 +1,18 @@
...
+const extractEventFromTransaction = async (
+    transaction: ethers.ContractTransaction,
+    eventSignature: string,
+    contractInterface: ethers.utils.Interface
+) => {
+    const receipt = await transaction.wait();
+    const eventLogs = receipt.logs.filter(log => log.topics[0] === ethers.utils.id(eventSignature));
+    const filteredLogs = eventLogs.map(log => ({ topics: log.topics, data: log.data }));
+    const events = filteredLogs.map(log => contractInterface.parseLog(log));
+    if (events.length !== 1) {
```
valiafetisov, line 12: I don't think it's anyhow restricted by the EVM. In fact, I would expect multiply `Transfer` events in one big (eg swap) transaction.
```suggestion
    if (events.length < 1 ) {
```

core/src/vaults.ts
valiafetisov: > receives "what to look for" and "how to parse"

No, you provide "what to look for + types" and "the whole interface of the contract", while `what to look for + types` in theory is enough to parse what you need. I don't see a problem with using try/catch to get only desired events.

Anyway, I think you can keep it as is if you want to add the interface of the whole contract when you need just a single event, but by suggestion is to simplify it. And it's a nit.

core/src/authorizations.ts
valiafetisov: I've created the issue myself https://github.com/sidestream-tech/unified-auctions-ui/issues/517

core/src/vaults.ts
KirillDogadin-std: 1. i've tried to implement the try-catch as in the linked issue. It won't "just work" - the interface's `parseLog` and `decodeEventLog` do not result in any return (assuming that i did not mess up something obvious & primitive)
2. given that the parsing by the incomplete interface results in the errors instead of array of logs - i also get the impression that the intention there is to use the parsing logic with the complete abi.

Due to the 2 above points + the intent of resulting code to be "simple" - i would side with the current solution still.

KirillDogadin-std:
TUSD-A is messing up the ci

it is both:

 - a proxy token - we fallback to overwriting vat balance
 - the join contract does not have sufficinet funds - we cannot withdraf from vat.

i will unsupport it for now

LukSteib:
> i will unsupport it for now

That is completely fine. As it is an offboarded collateral type no need to invest further for now.

valiafetisov (approved):

PR #511, opened 2022-10-26
Refactor `core/` for multi-callee support

### LINKED ISSUE #497: Refactor core for multi callee support
### Goal
Adjust core in a way to allow for multi callee support

### Context
As part of our liquidity sources stream we decided to implement multi callee support. For background discussions and overall concept please read #374 

This issue tackles all of the required adjustments to the core. 

### Tasks
- [x] Update collateral config `exchange` key to `exchanges` 
    from
    ```ts
    exchange: {
        callee: 'UniswapV3Callee',
        route: ['ETH'],
    },
    ```
    to
    ```ts
    exchanges: {
        "Uniswap v3": {
            callee: 'UniswapV3Callee',
            route: ['ETH'],
        },
        "Uniswap v2": {
            callee: 'UniswapV2CalleeDai',
            route: ['ETH'],
        },
        // "Uniswap v3 automatic route": {
        //    callee: 'UniswapV3Callee',
        //    route: undefined,
        // },
    },
    ```

    So basically prepending it with unique id, which will also be displayed on the frontend as title
- [x] Create function `getMarketData` that use `exchanges` configs and returns 

    ```ts
    {
        "Uniswap v3": {
             unitPrice: BigNumber,
             route: string[], // passed from the config in this PR, generated in case of automatic router in another PR
        },
        //...
    }
    ```

Notes: 
        - In case it's impossible to fetch specific market, the object should still be there, but `unitPrice` should be set to `new BigNumber(NaN)`
        - The main complexity here is that returned unit prices should all include exchange fees in order to be comparable between each other, see discussion above. For this, existing bid transaction fee logic should be reworked: `swapTransactionFee` should be split into variable part (that depends on the callee) and on constant part (basically the should be much more close to the `bidTransactionFee`); the variable part will then be hidden in the `unitPrice`. Technically that means, that while computing unit price (eg [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/d5325f33edbf3fe588b2670d8b5eb0d1ae82f0b0/core/src/calleeFunctions/UniswapV3Callee.ts#L37)) we should first remove exchange fees (can be hardcoded per callee per route length)
- [x] Create `getBestMarketData` function that sort output of the `getMarketData` by `unitPrice` and return the lowest (together with it's id)

    ```ts
    {
         marketId: "Uniswap v3"
         unitPrice: BigNumber,
    }
    ```
- [x] Modify `_getMarketPrice` to use `getBestMarketData` and just output `unitPrice`
- [x] During collateral auction enrichment
   - [x] Keep `marketUnitPrice` as is (but put the best market price there)
   - [x] Add result of the `getMarketData` under new `marketData` key along existing `marketUnitPrice`
   - [x] Add `suggestedMarketId` key to the auction type along existing `marketUnitPrice`
- [x] Extend `bidWithCallee` to accept `marketId` required argument
   - [x] Generate callee data based on the `marketId`
- [x] Modify store to send `suggestedMarketId` to the `bidWithCallee`
- [x] Modify bot to send `suggestedMarketId` to the `bidWithCallee`

aomafarag:
Closes #497 .

Checklist:


core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
```diff
@@ -12,10 +12,10 @@ const getCalleeData = async function (
     collateral: CollateralConfig,
     profitAddress: string
 ): Promise<string> {
-    if (collateral.exchange.callee !== 'CurveLpTokenUniv3Callee') {
+    if (!collateral.exchanges.hasOwnProperty('Curve Token V3')) {
```
valiafetisov, line 15: If you properly extend the type, you don't need to change this line

core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
```diff
@@ -12,10 +12,10 @@ const getCalleeData = async function (
...
         throw new Error(`Can not encode route for the "${collateral.ilk}"`);
     }
-    const route = await encodeRoute(network, collateral.exchange.route);
+    const route = await encodeRoute(network, (collateral.exchanges['Curve Token V3'] as RegularCalleeConfig).route);
```
valiafetisov, line 18: And will avoid typecasting (which is always bad)

core/src/calleeFunctions/UniswapV2LpTokenCalleeDai.ts
```diff
@@ -25,8 +25,14 @@ const getCalleeData = async function (
         profitAddress,
         joinAdapterAddress,
         minProfit,
-        await getUniswapRouteAddressesBySymbol(network, collateral.exchange.token0),
-        await getUniswapRouteAddressesBySymbol(network, collateral.exchange.token1),
+        await getUniswapRouteAddressesBySymbol(
+            network,
+            (collateral.exchanges['Uniswap Token V2'] as UniswapV2LpTokenCalleeConfig).token0
```
valiafetisov, line 30: Same here and everywhere, please remove those typecasting everywhere by applying the type changes

core/src/calleeFunctions/index.ts
```diff
@@ -20,30 +27,95 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
-    if (!calleeFuctions) {
+    const calleeFuctions = {} as Record<string, CalleeFunctions>;
+    Object.entries(collateral.exchanges).forEach(([key, value]) => {
+        calleeFuctions[key] = allCalleeFunctions[value.callee];
+    });
+    if (!Object.keys(calleeFuctions).length) {
         throw new Error(`Unsupported collateral type "${collateralType}"`);
     }
-    return await calleeFuctions.getCalleeData(network, collateral, profitAddress);
+    const calleesData = {} as Record<string, string>;
```
valiafetisov, line 43: Please always prefer to use correct type instead of record

core/src/calleeFunctions/index.ts
```diff
@@ -20,30 +27,95 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
     amount: BigNumber = new BigNumber('1')
-): Promise<BigNumber> {
+): Promise<Record<string, MarketData>> {
     const collateral = getCollateralConfigBySymbol(collateralSymbol);
-    const calleeFuctions = allCalleeFunctions[collateral.exchange.callee];
-    if (!calleeFuctions) {
+    const calleeIdAndFuctions = {} as Record<string, CalleeFunctions>;
+    Object.keys(collateral.exchanges).forEach(id => {
+        calleeIdAndFuctions[id] = allCalleeFunctions[collateral.exchanges[id].callee];
+    });
```
valiafetisov, line 59: In general `{} as Record` and `calleeIdAndFuctions[id] = xxx` is not a preferred way of doing it in typescript. Every typecasting is bad and basically cancels data validation. Instead, you would want to create a loop over the keys that creates all necessary values and then returns them altogether as a complete object – this way you wouldn't need to cast anything.

It also goes along with the idea that modifying (eg reassigning values to the object keys in javascript is very tricky and you should avoid it) Why? Because objects are stored as references, so the references are passed between functions, not copies of the same object. If in some other function you will change a key, it will also be changed everywhere – including functions running in parallel. For this case, it's always better to avoid using `object[key] =` at all not validate every time what's going on

core/src/calleeFunctions/index.ts
```diff
@@ -20,30 +27,95 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        marketData[id] = data;
+    }
+    return marketData;
+};
+
+export const getBestMarketData = async function (
+    network: string,
+    collateralSymbol: string,
+    amount: BigNumber = new BigNumber('1')
+): Promise<Record<string, string | BigNumber>> {
```
valiafetisov, line 92: This function should accept already calculated `auction.marketData: Record<string, MarketData>` it should NOT refetch it again (which is a lot of requests + it might change between the block)

core/src/calleeFunctions/index.ts
```diff
@@ -20,30 +27,95 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        marketUnitPrice: marketData[0][1].marketUnitPrice,
+    };
+};
+
+const _getMarketPrice = async function (
+    network: string,
+    collateralSymbol: string,
+    amount: BigNumber = new BigNumber('1')
+): Promise<BigNumber> {
+    return (await getBestMarketData(network, collateralSymbol, amount)).marketUnitPrice as BigNumber;
```
valiafetisov, line 118: No typecasting please

core/src/types.ts
```diff
@@ -6,6 +6,18 @@ export declare interface GasParameters {
     gasPrice?: string;
 }
 
+export declare interface MarketData {
+    marketUnitPrice: BigNumber;
+    exchangeFee: BigNumber;
+    route?: string[];
+    token0?: string;
+    token1?: string;
```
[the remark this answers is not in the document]
valiafetisov: please get an idea from the types that I've provided, you might need to create two extension types, then combine them with |

core/src/types.ts
```diff
@@ -6,6 +6,18 @@ export declare interface GasParameters {
     gasPrice?: string;
 }
 
+export declare interface MarketData {
+    marketUnitPrice: BigNumber;
+    exchangeFee: BigNumber;
```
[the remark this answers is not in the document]
valiafetisov: yes, we add it to market unit price, basen on the amount

core/src/calleeFunctions/index.ts
```diff
@@ -18,30 +25,95 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+    Object.keys(collateral.exchanges).forEach(id => {
+        calleeIdAndFuctions[id] = allCalleeFunctions[collateral.exchanges[id].callee];
+    });
+    if (!Object.keys(calleeIdAndFuctions).length) {
         throw new Error(`Unsupported collateral symbol "${collateralSymbol}"`);
     }
-    return await calleeFuctions.getMarketPrice(network, collateral, amount);
+    const marketData = {} as Record<string, MarketData>;
+    const gasPrice = await getGasPriceForUI(network);
+    const exchangeFee = gasPrice.multipliedBy(722651 - 145438);
```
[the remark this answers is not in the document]
valiafetisov: Yes, but
- this will calculate `exchangeFee` in ETH, you will need to convert it into DAI
- please move this calculation in to `fees` file
- please put numbers into variables eg `SWAP_TRANSACTION_GAS_LIMIT - BID_TRANSACTION_GAS_LIMIT` that are reused in other places in the same file too

core/src/calleeFunctions/index.ts
valiafetisov: The modern syntax is to use spread operator to recreate object `{ ...previousObject, newKey: newValue }` instead of using `Object.assign(Object.assign({}, previousObject), { newKey: newValue })` (which would do the same thing)

core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
valiafetisov: Yes, you're right, but I meant the usage of `hasOwnProperty` which depends on the arbitrary key name instead of the exact callee name

core/src/auctions.ts
```diff
@@ -83,7 +83,11 @@ const enrichAuctionWithMarketValues = async function (auction: Auction, network:
             marketUnitPrice,
             marketUnitPriceToUnitPriceRatio,
         };
-        const transactionGrossProfit = calculateTransactionGrossProfit(auctionWithMarketValues);
+        const transactionGrossProfit = calculateTransactionGrossProfit(
+            marketUnitPrice,
+            collateralToCoverDebt,
+            auction.approximateUnitPrice
+        );
```
valiafetisov, line 90: You already calculate `transactionGrossProfit` and `transactionGrossProfitDate` for each market inside `transactionGrossProfitDate`, why do you need to calculate it here again?

core/src/auctions.ts
```diff
@@ -127,7 +135,42 @@ export const enrichAuctionWithPriceDrop = async function (auction: Auction): Pro
...
 };
 
-export const enrichAuctionWithPriceDropAndMarketValue = async function (
+export const enrichAuctionWithMarketData = async function (auction: Auction, network: string): Promise<Auction> {
+    if (!auction.isActive || auction.isFinished || !auction.collateralToCoverDebt || !auction.approximateUnitPrice) {
+        return auction;
+    }
+    try {
+        const marketData = await getMarketData(network, auction.collateralSymbol, auction.collateralToCoverDebt);
+        const suggestedMarketId = (await getBestMarketData(marketData)).marketId;
```
valiafetisov, line 144: You might want to use all other keys from here to populate other values, instead of calculating it one more time inside `enrichAuctionWithMarketValues` – which not only unnecessarily, but can return different values, which will be very confusing

core/src/auctions.ts
```diff
@@ -127,7 +135,42 @@ export const enrichAuctionWithPriceDrop = async function (auction: Auction): Pro
...
+                auction.collateralToCoverDebt,
+                auction.approximateUnitPrice
+            );
+
+            market.transactionGrossProfitDate = calculateTransactionGrossProfitDate(
+                auction,
+                market.marketUnitPrice,
+                await getNetworkDate(auction.network)
+            );
+        }
```
valiafetisov, line 161: You're again, modifying `market` object, without copying it which will have side effects

core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
```diff
@@ -12,10 +12,10 @@ const getCalleeData = async function (
     collateral: CollateralConfig,
     profitAddress: string
 ): Promise<string> {
-    if (collateral.exchange.callee !== 'CurveLpTokenUniv3Callee') {
+    if (!('route' in collateral.exchanges['Curve Token V3'])) {
```
valiafetisov, line 15: Again, please prefer to validate `callee` name, not the key. The key is only used to be displayed on the frontend, is it not to identify the callee name. For example, **the same callee will be used multiple times** (with static vs auto routing) in the future! So you need to provide `marketId` to the `getCallee` data and validate it's callee name to be valid

core/src/types.ts
```diff
@@ -76,22 +81,42 @@ export declare interface AuctionTransaction extends Auction, TransactionFees {
...
+    transactionGrossProfit?: BigNumber;
+    transactionGrossProfitDate?: Date;
+    transactionNetProfit?: BigNumber;
+}
+
+declare interface MarketDataRouted extends MarketDataBase, Routed {}
+
+declare interface MarketDataTokenized extends MarketDataBase, Tokenized {}
+
+export type MarketData = MarketDataRouted | MarketDataTokenized;
```
valiafetisov, line 118: Why did you have to create new `Routed` type and extend it instead of `RegularCalleeConfig`? That is the problem you're experiencing here: https://github.com/sidestream-tech/unified-auctions-ui/pull/511#discussion_r1009555381

When you check callee name of the `RegularCalleeConfig`-type object, typescript can already say if `route` or `token*` keys should be present. But when you do the same for `MarketData` there is no (typed) connection between the callee name and the keys.

```suggestion
declare interface MarketDataRegular extends MarketDataBase, RegularCalleeConfig {}

declare interface MarketDataUniswapV2Lp extends MarketDataBase, UniswapV2LpTokenCalleeConfig {}

export type MarketData = MarketDataRegular | MarketDataUniswapV2Lp;
```

core/src/fees.ts
```diff
@@ -49,6 +60,15 @@ export const enrichAuctionWithTransactionFees = async function (
     let combinedSwapFeesETH = fees.swapTransactionFeeETH;
     let combinedBidFeesETH = fees.bidTransactionFeeETH;
 
+    if (auction.marketData && fees.swapTransactionVariableFeeDAI) {
+        const swapTransactionVariableFeePerUnitDAI = fees.swapTransactionVariableFeeDAI.dividedBy(
+            auction.collateralToCoverDebt
+        );
+        Object.values(auction.marketData).forEach(market => {
+            market.marketUnitPrice = market.marketUnitPrice.plus(swapTransactionVariableFeePerUnitDAI);
+        });
```
valiafetisov, line 69: As suggested, please don't modify values inside functions, this will create side effects in javascript. I would say you don't need to modify this function at all, as it is not suitable for modifying values, but for `enriching` objects with extra key-value pairs

core/src/fees.ts
```diff
@@ -15,16 +21,21 @@ export const getApproximateTransactionFees = async function (network: string): P
...
+    const authTransactionFeeETH = gasPrice.multipliedBy(AUTH_TRANSACTION_GAS_LIMIT);
+    const restartTransactionFeeETH = gasPrice.multipliedBy(RESTART_TRANSACTION_GAS_LIMIT);
 
     return {
         bidTransactionFeeETH,
         bidTransactionFeeDAI: await convertETHtoDAI(network, bidTransactionFeeETH),
         swapTransactionFeeETH,
         swapTransactionFeeDAI: await convertETHtoDAI(network, swapTransactionFeeETH),
+        swapTransactionVariableFeeETH,
+        swapTransactionVariableFeeDAI: await convertETHtoDAI(network, swapTransactionVariableFeeETH),
```
valiafetisov, line 38: I would also not include this new key into the auction object, since it's just for the internal purpose. Once again: the transaction fee consist of 

- A. exchanging one currency into another
- B. executing the auction

`B.` is approximately equals to `bidTransactionFeeETH`. Previously we included `A.` into `B.` and always displayed them together, but now we want to split those two and have a `A.` specified per callee/market. This means, you can add those keys to the `MarketDataBase` and create a small function, eg `getDefaultMarketFee` that is called in the `enrichAuctionWithMarketData` and  just returns those fees under `exchangeFeeEth` and `exchangeFeeDai`

core/src/types.ts
valiafetisov: Why exactly it did work before when it was in the `exchange` object directly `exchange.callee`, but doesn't work when it's one level deeper, in the `exchanges` record `exchanges[exchange].callee`?

core/src/auctions.ts
valiafetisov: Sorry, I meant `enrichAuctionWithMarketData`, not `transactionGrossProfitDate`. You have `marketData[marketId].transactionGrossProfit` which you can just take from there (if you would call `enrichAuctionWithMarketData` from within `enrichAuctionWithMarketValues`), instead of recalculating. As said above: recalculating might produce a different number and will be confusing

core/src/types.ts
valiafetisov: > since we know for certain that the `callee` name narrows down the type

No, it's not optimal. The linked issue doesn't talk about nesting at all. As I said, `exchange.callee` check did work before (so it's not a typescript problem), but currently types are implemented differently and with the current implementation I don't even expect it to work. Have you tried [the suggestion](https://github.com/sidestream-tech/unified-auctions-ui/pull/511#discussion_r1009568739)? Where can I see a commit with it?

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,97 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+                ...marketData,
+                [key]: {
+                    marketUnitPrice,
+                    token0: value.token0,
+                    token1: value.token1,
+                },
+            };
+        }
+    }
+    return marketData;
```
valiafetisov, line 80: - Please avoid using generic variable names, eg `key`, `value`, etc. (nit: If you need to distinguish between singular `marketData` and object with `marketData`s, let's rename `marketData` key to `marketDataRecords`?)

- [The problem related to the types](https://github.com/sidestream-tech/unified-auctions-ui/pull/511#discussion_r1009568739) is that here you have to return proper `MarketDataRegular` OR `MarketDataUniswapV2LpToken` type. It might help to assign the object keys based on the callee type, not `'route' in value`. Otherwise how typescript will know the connection?
   - The problem might also It might be related to the `Object.entries` typescript issue that you mentioned before – so let's avoid using it and instead use `for (const marketId in collateral.exchanges)`?

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,97 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+    let isCollateralSupported = false;
+    for (const market of Object.values(collateral.exchanges)) {
+        if (market.callee && allCalleeFunctions[market.callee]) {
+            isCollateralSupported = true;
+            break;
+        }
+    }
+    if (!isCollateralSupported) {
         throw new Error(`Unsupported collateral symbol "${collateralSymbol}"`);
     }
```
valiafetisov, line 52: Nit: most of the time the usage of let can be avoided by better structured code
```suggestion
    const isCollateralSupported = Object.values(collateral.exchanges).some(market => market.callee && allCalleeFunctions[market.callee])
    if (!isCollateralSupported) {
        throw new Error(`Unsupported collateral symbol "${collateralSymbol}"`);
    }
```

core/src/calleeFunctions/index.ts
valiafetisov: Do you actually provide callee name to the market data? Why not do this instead?

```ts
{
    marketUnitPrice,
    ...value,
},
```

If none of the above things work, you can build typed data object directly

```ts
 const marketData: MarketDataUniswapV2Lp = {
      marketUnitPrice,
      token0: value.token0,
      token1: value.token1,
 }
```
And then assign it under arbitrary market id

valiafetisov: The proof of concept [is compiled and works as expected](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEoQOYFcA2coDCcOOEEBA9iDMBsgN4BQyLyCxpEAXM63wD7IA5AFUQwAM4B3OAAcAagCYiJMgBE4wIbz4tBQgOoSwAUQAqACQJYoANwhjgtgMwrO23a33W7EADKyZhQA1hAgji5uZB6eyPqO0nLyrhzROrr6UOZWNvYRKaoQQgDc6VAUWJBcyMZQoBgA2gC6pQC+jIygkLCIKAkyCooBQaEgUeRUNHRMfOyF1aLiiYPDIWHjGlqlfGBrIAAM1bX126y7owCMR2B1IBhtHV3Q8EjIlCRw3cSU1LQM6WBgGBSNdbvd0hAAB4IAAWcDuEAk1XQCAoUAAJgAeY53AA0aEwuHw4x+UziyH6SSGgT2JMmtAAfA9GKiQMY3gB5Px+ACCZlQPL8AGVkRBURjsTd6vj3ngvjhSYzkABef58IQEVDyeQXbJCszZAC0PKE1RmsUBwO4wh89hqpksQlx6T4UNh8IwiLNLtiGtyKBGYWQyVNati4bmnAWtv8NNG+XGTp94ZY5Uq1oaQmyQiazpTrFaedihfSJfVIgAcgBJJRqHlVo0m73hy0g4SVmuKOsNyzIPwABST4bdcIRSLD+cWkgGyEDIGDilD5vzs1S1qnyyUq1GG00Q5XOz2h2E3f3B5Y5zCV2E2aL+bLxbvyBL7UYMDRyAAFKz2aiPvKzAAT1kFAKBgZAOQAIwAKzFMAADpbGILBEU-AguV5flBSFABKHCJ2Qd8oC-H8wGQEcPVA8CoNghAEKQnAUIkb8KH-aBiCAkD4Iosc8IIvhgHAz8eM9eDIzIFVlVVP1fG3MIEzXIR8OXFMf1YiB4JwCgME-IQ00gJ1yOhUdRP0iAcOTZ9LMEr8RI08SUCk6TKRWON1jXTYlP42I1NITTtN0y8DkMuz4KC-YLPDdo+HaVogA)

Screenshot 2022-11-02 at 15 29 59

core/src/auctions.ts
```diff
@@ -67,26 +67,50 @@ const calculateCollateralToCoverDebt = async function (network: string, auction:
...
         const collateralToCoverDebt = await calculateCollateralToCoverDebt(network, auction);
-        const marketUnitPrice = await getMarketPrice(network, auction.collateralSymbol, collateralToCoverDebt);
-        const marketUnitPriceToUnitPriceRatio = auction.approximateUnitPrice
-            .minus(marketUnitPrice)
-            .dividedBy(marketUnitPrice);
-        const auctionWithMarketValues = {
+        let marketDataRecords = await getMarketData(network, auction.collateralSymbol, collateralToCoverDebt);
+        const exchangeFees = await getDefaultMarketFee(network);
+        const exchangeFeePerUnitDAI = exchangeFees.exchangeFeeDAI.dividedBy(collateralToCoverDebt);
+        for (let marketData of Object.values(marketDataRecords)) {
```
valiafetisov, line 79: Currently, you put the whole loop into the try/catch here. So in case any market fails (which is expected, see comment below `catch`) the auction will be returned without any market data. Please make so that each market data is checked separately

Nit: you _might_ want to move the whole loop into a separate function to simplify try/catching it.

core/src/auctions.ts
```diff
@@ -173,12 +195,12 @@ export const enrichAuction = async function (
...
 
-    // enrich them with market values
-    const auctionWithMarketValue = await enrichAuctionWithMarketValues(auctionWithPriceDrop, network);
+    // enrich them with market data
+    const auctionWithMarketData = await enrichAuctionWithMarketData(auctionWithPriceDrop, network);
 
     // enrich with profit and fee calculation
     const fees = await getApproximateTransactionFees(network);
-    const auctionWithFees = await enrichAuctionWithTransactionFees(auctionWithMarketValue, fees, network);
+    const auctionWithFees = await enrichAuctionWithTransactionFees(auctionWithMarketData, fees, network);
```
valiafetisov, line 203: How gross profit can be calculated inside `enrichAuctionWithMarketData` if `auctionWithFees` is only calculated here? I imagine that the order needs to be changed?

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,101 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        try {
+            marketUnitPrice = await allCalleeFunctions[marketData.callee].getMarketPrice(
+                network,
+                collateral,
+                marketId,
+                amount
+            );
+        } catch {
+            marketUnitPrice = new BigNumber(NaN);
+        }
```
valiafetisov, line 62: Why do you need to try/catch it here? (if your function can still `throw`)

In other words, if you want to return market data records in _any_ case, you need to move `Unsupported collateral symbol "${collateralSymbol}"` to be thrown on the market level?

I also suggest to add new optional key `marketData.errorMessage` which will contain those errors (because previously it was cached on the frontend level and we were able to display this error)

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,101 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        if (marketData.callee === 'UniswapV2LpTokenCalleeDai') {
+            marketDataRecords = {
+                ...marketDataRecords,
+                [marketId]: {
+                    marketUnitPrice,
+                    token0: marketData.token0,
+                    token1: marketData.token1,
+                },
```
valiafetisov, line 70: Don't you need to add other fields from `marketData` here as well?
```suggestion
                [marketId]: {
                    ...marketData,
                    marketUnitPrice,
                    token0: marketData.token0,
                    token1: marketData.token1,
                },
```

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,101 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        }
+        if (b[1].marketUnitPrice.isNaN()) {
+            return -1;
+        }
+        return a[1].marketUnitPrice.minus(b[1].marketUnitPrice).toNumber();
+    });
+    return {
+        marketId: marketDataRecordsSorted[0][0],
+        marketUnitPrice: marketDataRecordsSorted[0][1].marketUnitPrice,
+    };
```
valiafetisov, line 110: Nit: Maybe it might actually make sense to return just the `marketId` from this function (renaming it to `getBestMarketId`) or the whole market data. Otherwise the name is misleading

core/src/auctions.ts
valiafetisov: you're right

valiafetisov: But the same applies to `netProfit`: we suppose to calculate it for every market, right?

core/src/auctions.ts
valiafetisov: So the question then boils down to: why do you need try catch above the whole loop then if you catch market problems on the level deeper? As it will return incorrect state, where `marketDataRecords === undefined`.

The proposed solution is to wrap the parts into try/catch at the place where you need to replace them with other "error"-like objects

core/src/calleeFunctions/index.ts
valiafetisov: I still don't see any `errorMessage` inside `marketData`

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,104 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        return {
+            ...marketData,
+            marketUnitPrice,
+            errorMessage,
+        };
+    }
+    return {
+        ...marketData,
+        marketUnitPrice: marketUnitPrice ? marketUnitPrice : new BigNumber(NaN),
+    };
```
valiafetisov, line 67: Why do you need to try catch it here if you already catching it in the parent function `getMarketData` (within the loop)?

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,104 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+    for (const marketId in collateral.exchanges) {
+        let marketData: MarketData;
+        try {
+            marketData = await getMarketDataById(network, collateral, marketId, amount);
+        } catch (error: any) {
+            marketData = {
+                errorMessage: error,
+                marketUnitPrice: new BigNumber(NaN),
+                route: [],
+            };
```
valiafetisov, line 86: Why does this suddenly returns a `route`?

core/src/auctions.ts
valiafetisov: The whole thing is still in the try/catch, right?

core/src/auctions.ts
```diff
@@ -67,26 +67,50 @@ const calculateCollateralToCoverDebt = async function (network: string, auction:
...
+                    marketData.marketUnitPrice,
+                    collateralToCoverDebt,
+                    auction.approximateUnitPrice
+                ),
+                transactionGrossProfitDate: calculateTransactionGrossProfitDate(
+                    auction,
+                    marketData.marketUnitPrice,
+                    currentDate
+                ),
+            };
```
valiafetisov, line 98: The idea of the last suggestion was to move this whole block with calculations into the `getMarketDataById`. The reason is: it's extremely confusing that everything is a `marketData` and there is no types to enforce proper keys. You call `makertData` `collateral.exchanges[marketId]` in one place, `{ errorMessage: error, marketUnitPrice: new BigNumber(NaN), route: [], }` in another, then here you call `marketData` something else, but those are all different thing and we have typescript to enforce them

core/src/auctions.ts
valiafetisov: This wasn't addressed?

LukSteib (changes requested):
Tested on hardhat fork

Running the **bot** resulted in the following errors: 

- [ ] When testing on current block: Not able to participate in any auction (similar to the problem seen when using the UI below) -> needs to be addressed

```
ℹ collateral keeper: auction "ETH-A:836" is not tradable                                                                  19:00:25
``` 



---



Testing the **frontend**:

- [ ] Currently unable to swap any collateral into profit as the button to access transaction page is disabled (see below)



image

LukSteib (changes requested):
Tested via hardhat fork, simulation `ETH-A` auction

- [ ] The bot throws an error when trying to execute the swap transaction (see terminal output below). While authorisation steps work as expected there seems to be a problem with callee contract determination
- [ ] Similar problem on frontend (see console output below)




terminal output for bot execution:

```sh 
ℹ collateral keeper: auction "ETH-A:836" net profit is 6154 DAI after transaction fees, moving on to the execution        12:16:04
ℹ keeper: wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" has not been authorized yet. Attempting authorization now...
ℹ Transaction is preparing to be mined...                                                                                 12:16:04
ℹ Transaction is pending to be mined...                                                                                   12:16:05
ℹ Transaction was mined into block "15939548"                                                                             12:16:05
ℹ keeper: wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" successfully authorized via "0x76f9eed11c31cd9fcfe3226af1dd21613f9f2fbed4f724b211cdfe55bc3ee9c3" transaction
ℹ collateral keeper: auction "ETH-A:836" net profit is 6154 DAI after transaction fees, moving on to the execution        12:16:05
ℹ keeper: wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" has already been authorized                                 12:16:05
ℹ keeper: collateral "ETH-A" has not been authorized on wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" yet. Attempting authorization now...
ℹ Transaction is preparing to be mined...                                                                                 12:16:05
ℹ Transaction is pending to be mined...                                                                                   12:16:05
ℹ Transaction was mined into block "15939549"                                                                             12:16:05
ℹ keeper: collateral "ETH-A" successfully authorized on wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" via "0x6c4ee203e3f0d4810f93a9ea76c37d6af77f0b254f5cc3133bd64c3932ed9706" transaction
ℹ collateral keeper: auction "ETH-A:836" net profit is 6154 DAI after transaction fees, moving on to the execution        12:16:05
ℹ keeper: wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8" has already been authorized                                 12:16:05
ℹ keeper: collateral "ETH-A" has already been authorized on wallet "0x70997970C51812dc3A010C7d01b50e0d17dc79C8"           12:16:05
ℹ collateral keeper: auction "ETH-A:836": attempting swap execution                                                       12:16:05

 ERROR  collateral keeper: unexpected error: No callee address found for the "ETH-A" collateral on "custom" network
```

console output with stack trace:
```
Bidding error Error: No callee address found for the "ETH-A" collateral on "custom" network
    at getCalleeAddressByCollateralType (CALLEES.ts?96e2:45:1)
    at _callee15$ (auctions.ts?450c:334:1)
    at tryCatch (router-sdk.cjs.development.js?ce8d:423:1)
    at Generator.invoke [as _invoke] (router-sdk.cjs.development.js?ce8d:654:1)
    at Generator.eval [as next] (router-sdk.cjs.development.js?ce8d:479:1)
    at asyncGeneratorStep (asyncToGenerator.js?1da1:3:1)
    at _next (asyncToGenerator.js?1da1:25:1)
    at eval (asyncToGenerator.js?1da1:32:1)
    at new Promise (<anonymous>)
    at eval (asyncToGenerator.js?1da1:21:1)
```

LukSteib:
Next round of functional review (again based on hardhat fork with `ETH-A` simulation)

* was able to execute swap transaction via the bot

- [ ] frontend execution is not working. The console is flooded with warnings (see s1 below). And when trying to execute the swap transaction I receive same error as reported above

s1:
image


console error on swap execution:
```
Bidding error Error: No callee address found for the "ETH-A" collateral on "custom" network
    at getCalleeAddressByCollateralType (CALLEES.ts?96e2:45:1)
    at _callee15$ (auctions.ts?450c:341:1)
    at tryCatch (router-sdk.cjs.development.js?ce8d:423:1)
    at Generator.invoke [as _invoke] (router-sdk.cjs.development.js?ce8d:654:1)
    at Generator.eval [as next] (router-sdk.cjs.development.js?ce8d:479:1)
    at asyncGeneratorStep (asyncToGenerator.js?1da1:3:1)
    at _next (asyncToGenerator.js?1da1:25:1)
    at eval (asyncToGenerator.js?1da1:32:1)
    at new Promise (<anonymous>)
    at eval (asyncToGenerator.js?1da1:21:1)
```

valiafetisov:
> The above bidding error, however, will resolve after MarketPriceSelection is merged

The main/staging should never be in a broken state. So if we want to merge a refactoring PR, current functionality should stay intact. You can achieve that by temporary passing the suggest market id in the modified store action.

LukSteib (commented):
Was able to execute swap transaction via frontend ✔️

core/src/calleeFunctions/index.ts
```diff
@@ -23,27 +23,104 @@ const allCalleeFunctions: Record<CalleeNames, CalleeFunctions> = {
...
+        let marketData: MarketData;
+        try {
+            marketData = await getMarketDataById(network, collateral, marketId, amount);
+        } catch (error: any) {
+            marketData = {
+                errorMessage: error?.message,
+                marketUnitPrice: new BigNumber(NaN),
+                route: [], // dummy value: MarketData requires either a route or tokens
+            };
+        }
```
valiafetisov, line 87: If you remember, we want to show the route on the frontend. In case there is no "marketData", the route will be empty and there will be nothing to display.

If you move try/catch down the `getMarketDataById`, it can return correct `route` or `token` based on the collateral config, not just a dummy empty value.

Please do so in the next frontend PR

valiafetisov (approved):

LukSteib (approved):

core/src/calleeFunctions/index.ts
valiafetisov: Good question. I guess linked throw condition shouldn't even be possible (eg via static typing), or should be validated during tests as it's strictly depends on the static config, not on the external conditions.

PR #514, opened 2022-10-28
Support file method simulations

KirillDogadin-std:
Closes #505

Checklist:
- [x] issue number linked above after pound (`#`)
    - replace "Closes " with "Contributes to" or other if this PR does not close the issue
- [x] issue checkboxes are all addressed
- [x] manually checked my feature / not applicable
- [x] wrote tests / not applicable
- [x] attached screenshots / not applicable

PR #516, opened 2022-11-02
Adjust time warp

### LINKED ISSUE #515: Adjust skip time step for collateral auction simulation
### Goal
We ensure that the auction is in a profitable state but not expired yet after executing the "Skip Time" step of the current simulation

### Context
We already have a simulation in place that allows the creation + liquidation of vaults for all supported collateral types. This ultimately enables us to test auction participation.
However, in order to participate in an auction via swap feature it needs to be profitable (ie. enough time has be passed in order for the auction price to be below market price).

For some collateral auctions (explicitly `USDC-A` and `PAXUSD-A`) the auction is not yet profitable after the "Skip Time" step in our simulation.
Since we know when the auction will expire let's use this info and skip the time to a point in time that is closely before the auction expiration.

### Tasks

- [ ] Adjust the "Skip Time" step of the collateral auction simulation to increase likelihood of auction being profitable (and hence swapable)

KirillDogadin-std:
Closes #515

Checklist:

### DIFF
--- a/core/simulations/configs/vaultLiquidation.ts
+++ b/core/simulations/configs/vaultLiquidation.ts
@@ -8,6 +8,7 @@ import createVaultWithCollateral, {
     getLiquidatableCollateralTypes,
 } from '../helpers/createVaultWithCollateral';
 import promptToSelectOneOption from '../helpers/promptToSelectOneOption';
+import getContract, { getClipperNameByCollateralType } from '../../src/contracts';
 
 const simulation: Simulation = {
     title: 'Simulate liquidation Auctions',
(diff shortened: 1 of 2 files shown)


LukSteib (dismissed):
Checked for `ETH-A` and looking good 
Would like to test it on coll types like `USDC-A` via #514 once this PR is merged

core/simulations/configs/vaultLiquidation.ts
```diff
@@ -60,12 +61,23 @@ const simulation: Simulation = {
...
         },
         {
             title: 'Skip time',
             entry: async context => {
-                await warpTime(60, 60);
+                const contract = await getContract(
+                    TEST_NETWORK,
+                    getClipperNameByCollateralType(context.collateralType)
+                );
+                const auctionLifetime: number = (await contract.tail()).toNumber();
```
valiafetisov, line 74: Don't we have this number as a date already stored within collateral `auction.endDate` as per proposal in the issue? At least reuse the existing function https://github.com/sidestream-tech/unified-auctions-ui/blob/4693a9852b65ee87719d0b559aded9949386223f/core/src/fetch.ts#L11-L18 if you don't want to deal with dates

KirillDogadin-std: > Don't we have this number as a date already stored within collateral auction.endDate

this is harder because we would have to fetch vault in the simulation step and add it to the context (or something like that)

valiafetisov: > Do you remember that this already exists or do you check every time?

I did a quick search for `.tail()` in the codebase

valiafetisov (approved):

PR #531, opened 2022-11-10
feat: use automatic routing

### LINKED ISSUE #527: Extend multi-callee with auto router
### Goal

Automatic router is used for tokens exchangeable via `UniswapV3Callee` as a separate exchange config.

### Context

After
- Initial conceptualisation via https://github.com/sidestream-tech/unified-auctions-ui/issues/375
- Adding auto routes to the dashboard https://github.com/sidestream-tech/unified-auctions-ui/pull/467
- Multi-callee core refactoring https://github.com/sidestream-tech/unified-auctions-ui/pull/511
- Multi-callee frontend support https://github.com/sidestream-tech/unified-auctions-ui/pull/498

We're finally ready to add full support of the uniswap automatic router. The idea is to add them as a new separate exchange config, eg:
```diff
  exchanges: {
+     'Uniswap V3 auto router': {
+         callee: 'UniswapV3Callee',
+         automaticRouter: true,
+     },
      'Uniswap V3': {
          callee: 'UniswapV3Callee',
          route: ['ETH'],
      },
      'Uniswap V2': {
          callee: 'UniswapV2CalleeDai',
          route: ['ETH'],
      },
  },
```
And then get quote and the route in the `getMarketPrice` (or some higher level function). Then, route will become part of the `marketData` object inside `marketDataRecords['Uniswap V3 auto router']` during the auction enrichment and will be picked up _from there_ by the `getCalleeData` during auction execution.

Note: we currently store routes as just a list of intermediate `collateralSymbols` through which the exchange should be made eg:
```ts
const route = [ 'ETH' ]
```
While the callees accept array of token addresses + pools, eg:
```ts
const path = [
  '0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9', // AAVE address
  3000, // AAVE+WETH uniswap pool
  '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', // WETH address
  3000, // WETH+DAI uniswap pool
  '0x6B175474E89094C44Da98b954EedeAC495271d0F' // DAI address
]
```
While we might want to store this data in this format:
```ts
const pools = [
  {
    names: ['AAVE', 'WETH'],
    addresses: ['0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9', '0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2'],
    fee: 3000,
  },
  {
    names: ['WETH', 'DAI'],
    addresses: ['0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2', '0x6B175474E89094C44Da98b954EedeAC495271d0F'],
    fee: 3000,
  }
]
```

Therefore, we should:
- Keep `exchange[marketId].route` config as now
- Transform `route` format into `pools` format during enrichment and store it under `marketDataRecords[marketId].pools` instead of `route`
- We should expect that intermediate tokens might be tokens that are not part of the collateral config
  - So that we need to store their address in the `marketDataRecords[marketId].pools`
  - So that we might need to fetch the symbol name from the chain (if it's not provided with the route)

### Assets

- `getUniswapAutoRoute` https://github.com/sidestream-tech/unified-auctions-ui/blob/cbf572753d9dde45d03ff63be8a62cbe03f70548/core/src/calleeFunctions/helpers/uniswapAutoRouter.ts#L16-L49

### Implementation proposal

- [ ] Refactor `route->pools`:
  - [ ] Generate a new `pools` value during enrichment _instead of_ `route`
  - [ ] Use `pools` on the frontend
  - [ ] Use `pools` value during execution (ie create `encodePools` instead of using `encodeRoute`)
  - [ ] Check that old routes still working
  - [ ] (optional) Merge `enrichAuctionWithMarketDataRecords` and `enrichMarketDataRecordsWithValues` into one function, but without a loop `getMarketDataById` which returns both fetched values, calculated values and pools
- [ ] Add new `automaticRouter` markets to the collateral config
- [ ] Handle new `automaticRouter` type when the callee is `UniswapV3Callee` via calling `getUniswapAutoRoute`
  - [ ] Transform output of the `getUniswapAutoRoute` into the `pools` format
  - [ ] Include estimated exchange fees into marketUnitPrice via using [`route.quoteGasAdjusted `](https://docs.uniswap.org/sdk/guides/auto-router)
  - [ ] Do not reapply hardcoded fees when `automaticRouter` is present

KirillDogadin-std:
Closes #527

Checklist:


KirillDogadin-std:
I've added some codechanges that already indicate what code adjustment i'm intending to implement. Please go through them and point out some decisions that are incorrect/in fact harmful.

E.g. https://github.com/sidestream-tech/unified-auctions-ui/pull/531/files#diff-c6b4bdf1b82a84a00b56df1404f49c064a303b5919e55e1335c7afecaab1a7edR12 seems questionable.

core/src/calleeFunctions/index.ts
```diff
@@ -24,15 +27,44 @@ export const getCalleeData = async function (
...
+    amount: BigNumber = new BigNumber('1')
+): Promise<string[] | undefined> => {
+    const calleeConfig = collateral.exchanges[marketId];
+    if (!('automaticRouter' in calleeConfig)) {
+        return undefined;
+    }
+    return (await fetchAutoRouteInformation(network, collateral.symbol, amount.toFixed())).route || [];
+}
+
+const getMarketDataWithoutPrice = (calleeConfig: CalleeConfig, route: string[] | undefined, pools: Pool[] | undefined) => {
```
valiafetisov, line 53: - Not sure what this function is and what it is doing in `calleeFunction`
- You don't need to pass `route: string[] | undefined, pools: Pool[] | undefined` you only need to pass one thing. Route should not be used beyond the enrichment step. Making one or the other available and spreading `'automaticRouter' in calleeConfig` beyond enrichment is not desirable

core/src/calleeFunctions/index.ts
```diff
@@ -44,6 +76,10 @@ export const getMarketDataById = async function (
     if (!allCalleeFunctions[calleeConfig?.callee]) {
         throw new Error(`Unsupported callee "${calleeConfig?.callee}" for collateral symbol "${collateral.symbol}"`);
     }
+
+    const route = 'route' in calleeConfig ? calleeConfig.route : await getCalleeAutoRoute(network, collateral, marketId, amount);
+    const pools = route ? await routeToPool(network, route, UNISWAP_FEE) : undefined;
```
valiafetisov, line 81: Same here. `pools` should replace `route` at this point

core/src/types.ts
```diff
@@ -88,6 +89,11 @@ export declare interface RegularCalleeConfig {
     route: string[];
 }
 
+export declare interface UniswapV3AutoRouterCalleeConfig {
+    callee: RegularCalleeConfig['callee'];
```
valiafetisov, line 93: Since you're extending `RegularCalleeConfig['callee']` it's not only `UniswapV3`
```suggestion
export declare interface AutoRouterCalleeConfig {
    callee: RegularCalleeConfig['callee'];
```

core/src/types.ts
```diff
@@ -108,23 +114,35 @@ declare interface MarketDataBase extends Partial<ExchangeFees> {
...
+    pools: Pool[]
+}
+
+export declare interface MarketDataUniswapV2LpToken extends MarketDataBase, Omit<UniswapV2LpTokenCalleeConfig, 'callee'> {}
 
-declare interface MarketDataUniswapV2LpToken extends MarketDataBase, Omit<UniswapV2LpTokenCalleeConfig, 'callee'> {}
+export declare interface MarketDataUniswapV3Automatic extends MarketDataBase, Omit<UniswapV3AutoRouterCalleeConfig, 'callee'> {
+    route: string[];
+    pools: Pool[];
+}
```
valiafetisov, line 130: Pools should now be part of every market data, otherwise you risk to spread the logic of determining between automatic and hardcoded everywhere, which is a lot of edits and a lot of ifs

core/src/calleeFunctions/helpers/uniswapV2.ts
```diff
@@ -19,7 +19,10 @@ const EXCHANGE_RATE_CACHE = 20 * 1000;
 
 const getCalleeConfig = function (collateral: CollateralConfig, marketId: string): RegularCalleeConfig {
     const marketData = collateral.exchanges[marketId];
-    if (marketData?.callee === 'UniswapV2CalleeDai' || marketData?.callee === 'UniswapV3Callee') {
+    const isUniswapTokenNonAutoRouted =
+        (marketData?.callee === 'UniswapV2CalleeDai' || marketData?.callee === 'UniswapV3Callee') &&
```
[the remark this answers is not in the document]
KirillDogadin-std: collateral config contains exchanges that in its turn can be of different types such as `RegularCalleeConfig` or `UniswapV3AutoRouterCalleeConfig`. The calles of these types are the same.

as discussed in the issue , i have implemented
```
interface UniswapV3AutoRouterCalleeConfig {
    callee: RegularCalleeConfig['callee'];
    automaticRouter: true;
}
```

this means that checking for the callee name is no more sufficient to make sure that type `RegularCalleeConfig` is returned, hence i also check the automaticRouter parameter

core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
```diff
@@ -11,10 +11,12 @@ const getCalleeData = async function (
...
     collateral: CollateralConfig,
     marketId: string,
-    profitAddress: string
+    profitAddress: string,
+    _pools?: Pool[]
 ): Promise<string> {
     const marketData = collateral.exchanges[marketId];
-    if (marketData?.callee !== 'CurveLpTokenUniv3Callee') {
+    const isAutorouted = 'automaticRouter' in marketData;
+    if (marketData?.callee !== 'CurveLpTokenUniv3Callee' || isAutorouted) {
```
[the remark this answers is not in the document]
KirillDogadin-std: same response as above:

  - there's a callee config that contains the callee name and possibly a route.
  - we can't make sure (via ts) that there's a route by just using the name anymore.
  - i check for the presense of the autorouter flag to make sure that the arrived type is correct

KirillDogadin-std:
Encountering error when `route` is called

```
TypeError: Cannot read properties of undefined (reading 'sortsBefore')
    at V3PoolProvider.getPoolAddress (pool-provider.js?276b:86:1)
    at CachingV3PoolProvider.getPoolAddress (caching-pool-provider.js?9cf4:67:1)
    at CachingV3PoolProvider.getPools (caching-pool-provider.js?9cf4:30:1)
    at getHighestLiquidityV3NativePool (gas-factory-helpers.js?fa4f:32:1)
    at V3HeuristicGasModelFactory.buildGasModel (v3-heuristic-gas-model.js?385b:110:1)
    at async Promise.all (:3000/index 0)
    at async AlphaRouter.route (alpha-router.js?d7e0:329:1)
```

https://github.com/sidestream-tech/unified-auctions-ui/blob/81859b0fa1f79d6c542d92922d64a339c829850f/core/src/calleeFunctions/helpers/uniswapAutoRouter.ts#L32

KirillDogadin-std:
The call of `fetchAutoRouteInformation` on production env works, for local setup it does not:

Possible reason: chain id is actually hardcoded :/ . SO when the router is created on main, the chain id corresponds to the actual one. On local setup it does not. Even on main branch the error appears (in the dashboard). So i would guess that setting the actual chain id should work (via fetching it with the existing method).
But in this case the error is thrown:
```
Error: No address for Uniswap Multicall Contract on chain id: 1337
```

@valiafetisov do you have the instant idea on how to solve this? If not, i then go investigate

valiafetisov:
> The call of fetchAutoRouteInformation on production env works, for local setup it does not:

If by "local setup" you mean "hardhat fork", then yes. It seems to be hardcoded:

https://github.com/sidestream-tech/unified-auctions-ui/blob/695d43daaa69d27a900cc625820ef65452cfd3ad/core/src/calleeFunctions/helpers/uniswapAutoRouter.ts#L24

Maybe making it dynamic will already solve the error?

KirillDogadin-std:
> So i would guess that setting the actual chain id should work (via fetching it with the existing method).
But in this case the error is thrown:

This paragraph was intended to let know that I've already tried your suggestion above. See the error is that is thrown in the same comment.

valiafetisov:
> This paragraph was intended

Sorry, I thought that reinstalling the dependencies solved the original error https://github.com/sidestream-tech/unified-auctions-ui/pull/531#issuecomment-1313612602

> `No address for Uniswap Multicall Contract on chain id: 1337`

This error is expected, since 1337 is a chain id that exist only for us. Uniswap SDK doesn't have an idea what is this chain id, as they have to unfortunately still hardcode some staff per chain id in the SDK (eg contract addresses). So for the hardhat fork we would still need to provide chain id that the hardhat is forking, not the one it is calling itself. This will solve the above problem and will bring us back to the initial:

> `TypeError: Cannot read properties of undefined (reading 'sortsBefore')`

Which according to a quick search originates from https://stackoverflow.com/questions/64674034/uniswap-typeerror-cannot-read-property-sortsbefore-of-undefined invalid `Token` instance creation. We create it via `getUniswapTokenBySymbol` where `decimalChainId` is passed derived from `network` instead of using the same as passed to the `AlphaRouter`.

KirillDogadin-std:
Currently some tests started to fail after the 'eth-a' collateral was extended with the automatic routing config.
One of the tests is `Fetches not liquidatable vault` and it goes into the process of extracting the route automatically. it hangs during `AlphaRouter.route()` call in `getUniswapAutoRoute` function.

See commit [87dfced](https://github.com/sidestream-tech/unified-auctions-ui/pull/531/commits/87dfced12870508835986621fbaac1f31005d21a) which pinpoints the place.
To reproduce you can run `npm run test -- ./test/vault-test.ts` (the very first test fails / times out for me)

Adjusting the mocha test timeout to 100 seconds instead of 40s did not solve the problem.

valiafetisov:
It's certainly the timeout. Dashboard on the hardhat first time took a very long time for me to load

LukSteib (commented):
Current state of functional review:
* I am stuck in the loading state of the general auction table (see screenshot below)
* The console throws a lot of errors several times (see below)

stuck loading state:
image

error message 1:
```
Error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0x1F98415757620B543A52E61c46B32eB19261F984","data":"0x1749e1e30000000000000000000000000000000000000000000000000000000000000020000000000000000000000000000000000000000000000000000000000000001500000000000000000000000000000000000000000000000000000000000002a0000000000000000000000000000000000000000000000000000000000000034000000000000000000000000000000000000000000000000000000000000003e00000000000000000000000000000000000000000000000000000000000000480000000000000000000000000000000000000000000000000000000000000052000000000000000000000000000000000000000000000000000000000000005c00000000000000000000000000000000000000000000000000000000000000660000000000000000000000000000000000000000000000000000000000000070000000000000000000000000000000000000000000000000000000000000007a0000000000000000000000000000000000000000000000000000000000000084000000000000000000000000000000000000000000000000000000000000008e000000000000000000000000000000000000000000000000000000000000009800000000000000000000000000000000000000000000000000000000000000a200000000000000000000000000000000000000000000000000000000000000ac00000000000000000000000000000000000000000000000000000000000000b600000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000ca00000000000000000000000000000000000000000000000000000000000000d400000000000000000000000000000000000000000000000000000000000000de00000000000000000000000000000000000000000000000000000000000000e800000000000000000000000000000000000000000000000000000000000000f20000000000000000000000000b4e16d0168e52d35cacd2c6185b44281ec28c9dc000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac000000000000000000000000000000000000000000000000000000000000000000000000000000000d4a11d5eeaac28ec3f61d100daf4d40471f1852000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000a478c2975ab1ea89e8196811f51a7b7ade33eb11000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000bb2b8038a1640196fbe3e38816f3e67cba72d940000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac0000000000000000000000000000000000000000000000000000000000000000000000000000000094b0a3d511b6ecdb17ebf877278ab030acb0a878000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000ae461ca67b15dc8dc81ce7615e0320da1a9ab8d5000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000b20bd5d04be54f870d5c0d3ca85d82b34b836405000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000231b7589426ffe1b75405526fc32ac09d44364c4000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac00000000000000000000000000000000000000000000000000000000000000000000000000000000765be0ee8e1857a8d7e906afa07cf0d2cf36ff59000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac0000000000000000000000000000000000000000000000000000000000000000000000000000000021b8065d10f73ee2e260e5b47d3344d3ced7596e000000000000000000000000000000000000000000000000000000000005b8d8000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000040902f1ac000000000000000000000000000000000000000000000000000000000000000000000000000000009c4fe5ffd9a9fc5678cfbd93aa2d4fd684b67c4c000000000000000000000    at Logger.makeError (index.js?ef0f:219:1)
    at checkError (json-rpc-provider.js?9d4a:64:1)
    at StaticJsonRpcProvider.eval (json-rpc-provider.js?9d4a:533:1)
```

error message 2:
```
Wallet authorization status error: missing revert data in call exception; Transaction reverted without a reason string [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (data="0x", transaction={"to":"0x35D1b3F3D7966A1DFe207aa4514C12a259A0492B","data":"0x4538c4eb000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb922660000000000000000000000009759a6ac90977b93b58547b4a71c78317f391a28","accessList":null}, error={"reason":"missing response","code":"SERVER_ERROR","requestBody":"{\"method\":\"eth_call\",\"params\":[{\"to\":\"0x35d1b3f3d7966a1dfe207aa4514c12a259a0492b\",\"data\":\"0x4538c4eb000000000000000000000000f39fd6e51aad88f6f4ce6ab8827279cfffb922660000000000000000000000009759a6ac90977b93b58547b4a71c78317f391a28\"},\"latest\"],\"id\":137,\"jsonrpc\":\"2.0\"}","requestMethod":"POST","serverError":{},"url":"http://127.0.0.1:8545"}, code=CALL_EXCEPTION, version=providers/5.6.8)
```

info message: 
`try catch will not finish`

LukSteib:
> Was it tested against hardhat fork? 

Yes 

> If so, can you please 1) restart hardhat 2) go to [the dashboard using custom network](http://localhost:3000/dashboard?network=custom) 3) wait till the dashboard is loaded (or refresh page every now and then until hardhat cached all prices) 4) retest collateral page again?

After waiting for several minutes (and refreshing the dashboard page every now and then) was not able to load a single token's information. Only info provided via console is `try catch will not finish    uniswapAutoRouter.ts?1c1d:38`


image

KirillDogadin-std:
> After waiting for several minutes (and refreshing the dashboard page every now and then) was not able to load a single token's information

That's indeed so bad/slow.
I recall leaving it in the tab, coming back from lunch and seeing some routes loaded.

perhaps reverting [87dfced](https://github.com/sidestream-tech/unified-auctions-ui/pull/531/commits/87dfced12870508835986621fbaac1f31005d21a) (it is a demo commit without much useful changes) will make it better because rn there's literally a useless function call. This "doubles" the time needed to load the dashboard record.

On top of that
Currently one more problem that i did not investigate yet is known. If you run tests, the oracle price fetching test will fail one of the test cases (because ETH-A has the autorouter in the config now). I don't know exactly how, but this autorouting backfired into the oracle functionality.

valiafetisov:
Latest state: I was able to load created ETH-A auction, but its auto-routed price displayed as unknown

Screenshot 2022-11-17 at 11 35 28

valiafetisov:
> If you run tests, the oracle price fetching test will fail one of the test cases (because ETH-A has the autorouter in the config now).

Yes, I think this is related to the problem discovered in https://github.com/sidestream-tech/unified-auctions-ui/pull/512#discussion_r1007066637. Somehow other tests influence oracle prices. I assume it can happen when some transactions are executed in-between tests. More specifically, it can be related to the use of `callStatic` trick in [our codebase](https://github.com/sidestream-tech/unified-auctions-ui/blob/24a1932c9eb9e12c69f052a0b1788574cdda5833/core/src/calleeFunctions/helpers/uniswapV3.ts#L62), which for some reason influence latest block number/internal date.

valiafetisov:
Latest state was: 
- Auto-routing works on goerli for the dashboard, but still doesn't work for auctions (shows undefined)
- Types are not optimal, see below

The suggestion was to isolate the change of the config on the level of the collateral config and then transform both auto-routed and non-auto-routed collaterals into the same type that always have `pools`. This way, the change will be isolated to enrichment. Passing it `autoRouter` more deeply down will not be useful anyway. You can no longer refetch routing information at the time you need it (like it was possible before to get it from the collateral config) because refetching might return different values (eg a different route) than displayed to the user.

```diff
declare interface MarketDataBase extends Partial<ExchangeFees> {
    marketUnitPrice: BigNumber;
    marketUnitPriceToUnitPriceRatio?: BigNumber;
    transactionGrossProfit?: BigNumber;
    transactionGrossProfitDate?: Date;
    transactionNetProfit?: BigNumber;
    errorMessage?: any;
-   pools?: Pool[];
}

export declare interface Pool {
    addresses: string[];
    fee: number;
+   routes: string[];
}
export declare interface MarketDataRegular extends MarketDataBase, Omit<RegularCalleeConfig, 'callee'> {
+   pools: Pool[];
}

export declare interface MarketDataUniswapV2LpToken
    extends MarketDataBase,
        Omit<UniswapV2LpTokenCalleeConfig, 'callee'> {}

- export type MarketData = MarketDataRegular | MarketDataUniswapV2LpToken | MarketDataUniswapV3Automatic;
+ export type MarketData = MarketDataRegular | MarketDataUniswapV2LpToken;
```

valiafetisov:
Checked the type. Everything seems fine, but I think `CalleeFunctions.getCalleeData` still need to accept `pools`: we can't generate (or regenerate) `pools` during execution (ie when user clicks "swap") – since regenerated pools (in case of the auto-router) might return a completely different pools array which will be different from that the user seen while pressing the button. Therefore, the proposal is generate pools _at the time of the auction enrichment_ and don't regenerate it in any later stage, but pass it via an argument to where it's needed (both frontend and core).

KirillDogadin-std:
ok, added the pools. Let's please establish if this solution layout is close enough to the desired one and i guess i will go try to debug the case where it's not loaded into the auction for some reason.

KirillDogadin-std:
I am now switching to finguring out what takes so long to load, so far still seeing a lot of requests pending in the console, but still loading just the list of auctions takes long for yet not clear to me reason.
Does not seem to be a repetitive auction-store action.

valiafetisov (commented):
As discussed orally, you might want to start debugging from making sure that `getPools` is not called [every second](https://github.com/sidestream-tech/unified-auctions-ui/blob/7aaec050700f464a13e38158f8ff25576f80b1c8/frontend/store/auctions.ts#L232).

KirillDogadin-std:
> As discussed orally, you might want to start debugging from making sure that getPools is not called every second.

not the case

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -9,13 +9,14 @@ const getCalleeData = async function (
     network: string,
     collateral: CollateralConfig,
     marketId: string,
-    profitAddress: string
+    profitAddress: string,
+    preloadedPools?: Pool[]
```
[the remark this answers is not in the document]
KirillDogadin-std: That's the enforced (by ts) signature of the function that is supposed to be generic - same signature for multiple callees.

`preloadedPools` is not required by multiple functions - therefore is set to be unrequired

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
```diff
@@ -57,17 +57,25 @@ export const fetchAutoRouteInformation = async function (
     try {
         const token = await getUniswapTokenBySymbol(network, collateralSymbol);
         const autoRouteData = await getUniswapAutoRoute(network, collateralSymbol, inputAmount, walletAddress);
-        const routes = autoRouteData.route[0].tokenPath.map(p => p.symbol);
+        const route = autoRouteData.route[0].tokenPath.map(p => {
+            if (!p.symbol) {
+                throw new Error(`Could not get symbol for token "${p.address}".`);
```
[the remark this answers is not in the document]
KirillDogadin-std: since the uniswap should always return it - let's keep it as is then?

core/src/calleeFunctions/helpers/uniswapV3.ts
```diff
@@ -76,7 +76,7 @@ export const convertCollateralToDaiUsingPool = async function (
     if (!route) {
         throw new Error(`No route found for ${collateralSymbol} to DAI`);
     }
-    const quote =  UNISWAP_FEE
+    const quote = quoteGasAdjusted ? Math.ceil(quoteGasAdjusted.toNumber()) : UNISWAP_FEE;
```
KirillDogadin-std, line 79: so if i change this to hardcoded `UNISWAP_FEE`, then the ui displays the value. if not, then autorouting extraction fails and in the ui the `Unknown` is displayed.

The error is not very informative:
```
Error: call revert exception [ See: https://links.ethers.org/v5-errors-CALL_EXCEPTION ] (method="quoteExactInput(bytes,uint256)", data="0x", errorArgs=null, errorName=null, errorSignature=null, reason=null, code=CALL_EXCEPTION, version=abi/5.6.4)
```

this one appears in multitude of the cases. Including (as of my xp) the cases where the provided encoded route includes a duplicate at the end (e.g. `[... 'DAI', 'DAI]`)

valiafetisov: Those are two completely different things: fee for the transaction paid as gas and fee for the exchange paid to the protocol. The second thing is sometimes called `pool` for the Uniswap v3 since there are different liquidation pools that are identified by the fee that pool takes. TLDR: you need to find another kind fee inside `getRouteAndGasQuote`

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
```diff
@@ -55,19 +61,34 @@ export const fetchAutoRouteInformation = async function (
...
-
+        const bestRoute = autoRouteData.route[0];
+        const route = bestRoute.tokenPath.map(p => {
+            if (!p.symbol) {
+                throw new Error(`Could not get symbol for token "${p.address}".`);
+            }
+            return p.symbol;
+        });
+        if (bestRoute.route.protocol !== Protocol.V3) {
+            throw new Error('Only V3 routes are supported.');
```
KirillDogadin-std, line 73: apparently not always the router returns fees. So i check the protocol version to ensure that the fees are there.

KirillDogadin-std:
@LukSteib could you summarize once again: which process exactly seems to work very slow for you?

LukSteib:
> which process exactly seems to work very slow for you?

Referring to an older review (see [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/531#pullrequestreview-1182954409)): The process to fetch all active auctions on `/collateral` was taking very long. Tested on a hardhat fork with a ETH-A auction creation simulation running.

valiafetisov:
Can we still investigate what's taking so long? For example, dashboard loading time doesn't seem to be that long, although it does the same kind of auto-routing request.

LukSteib:
Reviewed once more

on `goerli`:
* fetching of active auctions took ~1min
* fetching dashboard values took ~10sec per token
  * all resulted in error expect for `ETH-*` and `USDC-*`
* executing an `ETH-A` was successful (https://goerli.etherscan.io/tx/0x7c2dae2cead8c795089f75e6675ba32b612b4772d7e15faf1092063f347ddebc)

on hardhat fork (with active `ETH-A` auction):
* fetching of active auctions: waited for 6min but stuck with loading 
* fetching dashboard values: waited for 6min but not a single token was loaded

KirillDogadin-std:
Additional info points:

- `convertSymbolToDai` takes long. The subcall to `uniswapV3quoterContract.callStatic.quoteExactInputSingle` is something that takes too much time.
- there're side effects for tests:
  1. the oracle price extraction returns different price
  2. the oracle price extraction tests on ci/cd timeout

valiafetisov:
> however the part that was not changed in the pr

Loading auctions on the latest production build takes ~4 seconds https://main.auction-ui.k8s.sidestream.tech/collateral?network=goerli. So maybe `quoteExactInputSingle` is not the reason?

How long does auto router takes per collateral during auction enrichment? How long fetching every auction takes?

valiafetisov:
> the oracle price extraction returns different price

Can you please provide more details? Have you looked into this issue? When I was searching for something similar in the hardhat repo last time, I saw that there were problems with `callStatic`. Should we maybe try to update hardhat to the latest version?

KirillDogadin-std:
as agreed, here's the timings:

_method_: place `console.time()` and `console.timeLog()` at the start and the end of `enrichAuction`

_results_:
  - enriching auction on goerli that has autorouter enabled (ETH-A) requires approx 12 seconds
  - enriching multiple auctions due to this being async takes also round about the same time: 14 seconds for calling `fetchAllAuctions` in the store

valiafetisov:
@KirillDogadin-std, we discussed the situation with @LukSteib and think that based on the numbers, we should not introduce auto router as a default behaviour as it makes the participation in auctions less stable and unprepared for mass liquidation. In other words, added instability overweights benefits of having automatic router. Proposed workarounds:

a) Do not execute automatic router on the page load
   - Implement a button that enables auto-router for the selected auction
     - The button is visible inside expandable market selector
     - Once clicked it refetches the auction with auto-router information and enable further updates
   - The switch is not global (should be turned on for every auction separately)
   - The switch is not preserved on page refresh
   - (I imagine this can be done via a separate store variable holding auction ids and additional argument to the `enrichAuctionWithPriceDropAndMarketDataRecords`)

b) Research caching
   - via `CachingV3PoolProvider`
   - or manually / via external package, by supplying custom provider [to the `AlphaRouter`](https://github.com/sidestream-tech/unified-auctions-ui/blob/6cdba5def95207b3e1f4178878c1678124b18d63/core/src/calleeFunctions/helpers/uniswapAutoRouter.ts#L24)

<details><summary>Mock for the option a)</summary>

MakerDAO new collateral@2x (10)

</details>


My proposal would be to go with a) since we're short on time, and do it in the same PR for the same reason. Would you agree?

KirillDogadin-std:
> My proposal would be to go with a) since we're short on time, and do it in the same PR for the same reason. Would you agree?

sure, a) also seems like more preferable to me.

KirillDogadin-std:
> Do you have the same experience?

I have only experienced waiting for metamask prompt. I don't have any immediate ideas on how to improve this except for altering the market data refresh frequency (currently updated every second if i recall correctly).

KirillDogadin-std:
Alternatively, we might want to change the ui/ux even further: load the prices via autorouter just once on the button press. Currently we enable/disable the every-second-update logic (and the update lasts ~12 seconds)

valiafetisov:
> Currently we enable/disable the every-second-update logic

I think I pointed that out several times in the past. It should never be the case that market prices are refetched every second. We should only recalculate values every second, but not do any requests. Doesn't matter with or without auto router prices doesn't need to be refetched every second in any case.

valiafetisov:
> Should i then just enable caching for getMarketDataRecords?

I guess so. One way is caching (if it works). The other is to just split the logic into fetching data and updating calculations (as we suppose to do already [via two different intervals](https://github.com/sidestream-tech/unified-auctions-ui/blob/6cdba5def95207b3e1f4178878c1678124b18d63/frontend/store/auctions.ts#L231-L232))

Did [e921cfe](https://github.com/sidestream-tech/unified-auctions-ui/pull/531/commits/e921cfe2511847ece7a082f48784fab27752d949) solved the problem outlined by @LukSteib?

LukSteib:
Just did another round of review. Compared to the [last review](https://github.com/sidestream-tech/unified-auctions-ui/pull/531#issuecomment-1332192268) experience significantly improved even after enabling the auto router for an auction. 

Do you still want to look into splitting the logic into fetch and update or leave it as is for now?

valiafetisov:
For me it's fine as long as caching works

LukSteib (approved):
Approved from my side then

core/src/types.ts
```diff
@@ -88,6 +88,11 @@ export declare interface RegularCalleeConfig {
     route: string[];
 }
 
+export declare interface AutoRouterCalleeConfig {
+    callee: RegularCalleeConfig['callee'];
+    automaticRouter: true;
+}
```
valiafetisov, line 94: If you modify this to be 
```suggestion
export declare interface AutoRouterCalleeConfig {
    callee: 'UniswapV3Callee';
    automaticRouter: true;
}
```

core/src/calleeFunctions/CurveLpTokenUniv3Callee.ts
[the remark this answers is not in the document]
valiafetisov: > we can't make sure (via ts) that there's a route by just using the name anymore

You wouldn't need to do this if you modify `AutoRouterCalleeConfig` type

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -34,7 +40,7 @@ const getMarketPrice = async function (
     collateralAmount: BigNumber
 ): Promise<BigNumber> {
     // convert collateral into DAI
-    const daiAmount = await convertCollateralToDaiUsingRoute(network, collateral.symbol, marketId, collateralAmount);
+    const daiAmount = await convertCollateralToDaiUsingPool(network, collateral.symbol, marketId, collateralAmount);
```
valiafetisov, line 43: `convertCollateralToDaiUsingPool` means you always generate a new pool there (as it's not provided as an optional argument)?

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
valiafetisov: Are you talking about exchange fees? Is it a type problem or did you encounter such situations? (given line 58 above)

core/src/calleeFunctions/helpers/uniswapV2.ts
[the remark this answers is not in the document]
valiafetisov: Let's change the type then to the suggested in this review, so those changes are no longer necessary!

core/src/calleeFunctions/helpers/uniswapV3.ts
```diff
@@ -15,38 +17,69 @@ const getUniswapV3quoterContract = async function (network: string): Promise<eth
...
     const collateral = getCollateralConfigBySymbol(collateralSymbol);
-    const marketData = collateral.exchanges[marketId];
-    if (marketData?.callee !== 'UniswapV3Callee') {
+    const calleeConfig = collateral.exchanges[marketId];
+    if (calleeConfig?.callee !== 'UniswapV3Callee') {
         throw new Error(`getCalleeData called with invalid collateral type "${collateral.ilk}"`);
     }
     const collateralIntegerAmount = collateralAmount.shiftedBy(collateral.decimals).toFixed(0);
-    const route = encodeRoute(network, [collateral.symbol, ...marketData.route]);
+    const { route, fees } = await getRouteAndGasQuote(network, collateralSymbol, collateralAmount, marketId);
```
valiafetisov, line 72: As suggested above, this function should not generate pools. Moreover, `convertCollateralToDaiUsingPool` inside `uniswapV3` file should _never_ use auto router

core/src/calleeFunctions/index.ts
```diff
@@ -95,6 +167,12 @@ export const getMarketDataRecords = async function (
     return marketDataRecords;
 };
 
+export const getMarketDataRecords = memoizee(_getMarketDataRecords, {
+    promise: true,
+    maxAge: 30 * 1000,
+    length: 4,
+});
```
valiafetisov, line 174: Why was this change necessary? Anyway, `amount` here is an object, so it might not get cached properly (will be cached forever or not at all depending on how the amount is created)

core/src/constants/COLLATERALS.ts
```diff
@@ -90,6 +90,10 @@ const COLLATERALS: Record<string, CollateralConfig> = {
         symbol: 'ETH',
         decimals: 18,
         exchanges: {
+            'Uniswap V3 Autorouter': {
+                callee: 'UniswapV3Callee',
+                automaticRouter: true,
+            },
```
valiafetisov, line 96: Can you please add the same market to all other collaterals with `UniswapV3Callee`?

frontend/components/auction/collateral/MarketPriceSelection.vue
```diff
@@ -24,6 +24,32 @@
                                 </span>
                                 DAI
                             </td>
+                            <td class="pr-2 text-right whitespace-nowrap">
+                                <div v-if="id === 'Uniswap V3 Autorouter'">
```
valiafetisov, line 28: Let's not use title, but the `autoRouter` key

frontend/containers/AuctionsContainer.vue
```diff
@@ -164,6 +167,11 @@ export default Vue.extend({
             this.$store.dispatch('wallet/setup');
             this.$store.dispatch('authorizations/setup');
         },
+        toggleAutoRouterLoad(id: string): void {
+            const previousValue = this.$store.getters['auctions/getAuctionAutoRouterStates'][id] || false;
```
valiafetisov, line 171: Nit: move `|| false` to the getter

valiafetisov (commented):
Why did the package-lock changed?

KirillDogadin-std:
>  Why did the package-lock changed?

you've requested to update hardhat

core/package.json
```diff
@@ -7,7 +7,7 @@
     "dev": "tsc-watch --noClear",
     "build": "tsc --build tsconfig.json",
     "start": "node .",
-    "test": "npx hardhat test --network testnetwork --config local.hardhat.config.ts",
+    "test": "TEST_ENV=1 npx hardhat test --network testnetwork --config local.hardhat.config.ts",
```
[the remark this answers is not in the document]
KirillDogadin-std: apparently not, so setting it manually

core/src/calleeFunctions/UniswapV3Callee.ts
KirillDogadin-std: yes, also in the function body the route is always converted to pool

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
KirillDogadin-std: yes, the fees for exchange are not always provided. It is a type problem. I have to ensure that the protocol version is correct - otherwise i cannot be sure that the fees are in the return object.

core/src/calleeFunctions/helpers/uniswapV2.ts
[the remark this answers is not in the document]
KirillDogadin-std: ts is still complaining - the callee can be: version 2, version 3 that is not auto-routed. version 3 that is autorouted:

```
Type 'RegularCalleeConfig | AutoRouterCalleeConfig' is not assignable to type 'RegularCalleeConfig'.
     Property 'route' is missing in type 'AutoRouterCalleeConfig' but required in type 'RegularCalleeConfig'.
```

frontend/containers/AuctionsContainer.vue
KirillDogadin-std: getter gets the whole dict out. but i guess we can just remove `false` here since the fallback would happen if the value is falsy anyway 🤷🏻 .

Or should i refactor the getter to return the specific value?

valiafetisov: Technically yes, you can make functional getter. But it's alright, it's just a nit

valiafetisov:
- [ ] Frontend build fails
- [ ] Can you please test that bot still works (I see that `useAutoRouter` have default value, but maybe we use some other core internals there?)

KirillDogadin-std:
> Frontend build fails

sorry about this, i think we could benefit from a `.sh` that runs lint and build commands in every dir. It would save some iterations. Not that it matters for the near future though, does it? :D

> Can you please test that bot still works (I see that useAutoRouter have default value, but maybe we use some other core internals there?)

seems to not fail. But prints

```
 ERROR  collateral keeper: unexpected error: Cannot read properties of undefined (reading 'addresses')
```
for one of the auctions

which points towards pools not being generated. Should i investigate further here?

core/src/calleeFunctions/index.ts
valiafetisov: So do you want to remove it?

core/src/calleeFunctions/index.ts
```diff
@@ -123,10 +201,4 @@ const _getMarketPrice = async function (
     return marketDataRecords[bestMarketId].marketUnitPrice;
 };
 
-export const getMarketPrice = memoizee(_getMarketPrice, {
-    maxAge: MARKET_PRICE_CACHE_MS,
-    promise: true,
-    length: 3,
-});
```
valiafetisov, line 130: Why was it removed?

core/src/constants/COLLATERALS.ts
valiafetisov: This wasn't addressed so far

valiafetisov:
> Should i investigate further here?

Yes

core/src/calleeFunctions/index.ts
```diff
@@ -67,22 +128,31 @@ export const getMarketDataById = async function (
...
+    useAutoRouter = false
 ): Promise<Record<string, MarketData>> {
     const collateral = getCollateralConfigBySymbol(collateralSymbol);
     let marketDataRecords = {};
     for (const marketId in collateral.exchanges) {
         let marketData: MarketData;
         try {
-            marketData = await getMarketDataById(network, collateral, marketId, amount);
+            // turn off autorouter during tests because it takes too long
+            if ((process.env.NODE_ENV === 'test' || !useAutoRouter) && marketId === 'Uniswap V3 Autorouter') {
```
valiafetisov, line 143: Please don't use `marketId` to determine if it's auto-routed collateral or not, you have specific key `autoRouter` for it!

core/src/calleeFunctions/index.ts
KirillDogadin-std: yes, i confused the names 😢

core/src/constants/COLLATERALS.ts
KirillDogadin-std: i am extremely confused because i pasted the config and it's not there now. I need a christmas vacation apparently

core/src/calleeFunctions/index.ts
KirillDogadin-std: because i am illiterate and can't read function names.  Fixed

KirillDogadin-std:
> Yes

tests also fail for crveth with `Dog/not-unsafe`

will update this comment with the investigation later when it's ready

valiafetisov:
> tests also fail for crveth with `Dog/not-unsafe`

Can you please `cherry-pick` https://github.com/sidestream-tech/unified-auctions-ui/pull/546/commits/925b3d90d8725bd5b9842e2e76012dfd6bc15662 to temporary fix the tests?

valiafetisov (approved):
Tests fixed, code approved

LukSteib (commented):
* UI Test on `goerli` looks good ✅ 
* Tested keeper on `goerli` and executed ETH auction via swap (including auth txs) ✅ 
* Tested keeper on `hardhat fork` with ETH-A simulation running
  * authorisation steps successfully executed ✅ 
  * for swap tx I get the following error in the console (see below) 

```
ℹ collateral keeper: auction "ETH-A:840": attempting swap execution                                                       
 ERROR  collateral keeper: unexpected error: Cannot read properties of undefined (reading 'addresses')
```

Raising this as I am not able to reproduce it on `main` branch.

KirillDogadin-std:
Curent bugs seem to be related to:
 
1. https://github.com/sidestream-tech/unified-auctions-ui/pull/531#issuecomment-1324819185 not being fulfilled properly. At some cases the pools are not pregenerated apparently.
2. `routeToPool` logic does not cover the edgecase where the `route` is empty array. Therefore for this case empty pool will be generated

LukSteib:
I need more context / info in order to properly decide how to proceed here:
* Is the problem of not properly pregenerated pools only related to hardhat fork?
* Do we know how to tackle the problems described above and only need more time to address?

LukSteib:
> (1) requires me to put more thinking in to figure out how to wire up things properly, @valiafetisov might come up with the fix faster.

Ok, then I'd suggest to have @valiafetisov look into it in order to get this PR over the last mile and proceed with release.

valiafetisov:
Yes, it seems like `routeToPool` will produce invalid pool based on `exchanges[marketId].route`. It should include starting token into the array. Eg if `ETH-A['Uniswap V3'].route` is `[]`, the pool should include `ETH, DAI`, but currently (as far as I understood) only is `DAI`. Overall, the `getRouteSteps` + `routeToPool` seems to be overly complicated

KirillDogadin-std:
I've fixed (2), the side effect of this seems to be advance for (1) - the error is changed

now with

1. Hardhat fork
2. ETH-A auction simulation
3. choosing autorouter

The error is `"Error: VM Exception while processing transaction: reverted with reason string 'Too little received'"` from JSON-RPC

___

I was able to use the old Uniswap V3 router to execute swap

core/src/calleeFunctions/index.ts
```diff
@@ -24,14 +36,65 @@ export const getCalleeData = async function (
...
+    marketId: string,
+    amount: BigNumber = new BigNumber(1)
+) => {
+    const calleeConfig = collateral.exchanges[marketId];
+    if (calleeConfig.callee === 'UniswapV2LpTokenCalleeDai') {
+        return undefined;
+    }
+    const route =
+        'route' in calleeConfig ? calleeConfig.route : await getCalleeAutoRoute(network, collateral, marketId, amount);
+    return await routeToPool(network, route, collateral.symbol);
```
valiafetisov, line 80: As pointed out in the issue several times: we CAN NOT assume that the route provided from the auto router will ONLY contain symbols that are known to the maker. Therefore we CAN NOT convert automatic route to pool. Instead, the information provided by the AlphaRouter should contain enough information to build a pool object: addresses, symbols and fees.

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -34,7 +42,18 @@ const getMarketPrice = async function (
     collateralAmount: BigNumber
 ): Promise<BigNumber> {
     // convert collateral into DAI
-    const daiAmount = await convertCollateralToDaiUsingRoute(network, collateral.symbol, marketId, collateralAmount);
+    const { route, fees } = await getRouteAndGasQuote(network, collateral.symbol, collateralAmount, marketId);
+    if (!route) {
+        throw new Error(`No route found for ${collateral.symbol} to DAI`);
+    }
+    const pools = await routeToPool(network, route, collateral.symbol, fees);
```
valiafetisov, line 49: Here again, the pool should not be generated at the time when market price is requested (or at least should be optional in case of simple conversions like displaying transaction fees in DAI)

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -34,7 +42,18 @@ const getMarketPrice = async function (
...
+    const daiAmount = await convertCollateralToDaiUsingPool(
+        network,
+        collateral.symbol,
+        marketId,
+        collateralAmount,
+        pools
+    );
```
valiafetisov, line 56: Maybe a better explanation would've been: previously we expected that pools (/routes) are static. Now we should receive and store pools together with prices (because re-requesting price via auto-router can change the pool) – so if the pool is generated, it have to be stored in the auction together with the price (unless the price is not an auction price, but just a reference price)

valiafetisov:
Was able to execute transaction using auto router on a fork (at least that's how it looks)
Screenshot 2022-12-05 at 19 42 22

But two problems outlined above still there:
- We're regenerating pools in a different moment than when we fetch prices (so the price displayed and the route displayed are different and requires two requests)
- There will be error during pool generation in case auto router returns symbol that is not present in the chainlog (not sure how it will be handled)

LukSteib:
Next round of tests:

* UI Test on `goerli` looks good ✅ 
  * was able to execute a tx see: https://goerli.etherscan.io/tx/0xe479025dea52a8f5c87c427de00bbd645631e3380836ed6bcd892d81362ae24a -> net profit indicator was also very accurate
* Tested UI on hardhat fork and was able to execute ETH-A Auction ✅ 
* Tested keeper on `hardhat fork` with ETH-A simulation running
  * authorisation steps successfully executed ✅ 
  * swap tx ✅

LukSteib (approved):

LukSteib:
> But two problems outlined above still there:

> We're regenerating pools in a different moment than when we fetch prices (so the price displayed and the route displayed are different and requires two requests)
> There will be error during pool generation in case auto router returns symbol that is not present in the chainlog (not sure how it will be handled, hard to simulate)

We will tackle these problems in a dedicated follow up issue and proceed with the merge of this PR for now.

PR #532, opened 2022-11-10
Support `WalletConnect`

### LINKED ISSUE #353: Investigate WalletConnect support
### Goal
Outline on how to implement WalletConnect support

### Context
Current we support metamask only. Although this is the wallet with highest adoption so far, chances are that users would like to interact with our UI using a different wallet.
In the early days we arleady decided that WalletConnect would be a good way to go since this service constitutes a wallet aggregator on its own. So implementing support for WalletConnect would allow us to support a range of other wallets as well.

### Tasks
- [ ] Understand the current wallet implementation
- [ ] Understand WalletConnect preferred way of interaction
- [ ] Come up with a plan on how to support walletConnect and make a proposal in this issue

### Assets
WalletConnect docs: 
https://docs.walletconnect.com/

aomafarag:
Closes #353.

Screenshot from 2022-11-10 17-07-40

Checklist:

---

### !!! IMPORTANT !!!

~In order to run this PR, you have to provide our Infura ID (just the ID without the URL) as an environment variable under the name `INFURA_PROJECT_ID`~

**Testing**:

- Install a WalletConnect-supported wallet on a smartphone (I opted for MetaMask)
- This will be used for scanning the QR-code when prompted

### DIFF
--- a/core/src/rpc.ts
+++ b/core/src/rpc.ts
@@ -9,16 +9,20 @@ import {
 } from './network';
 import { formatToHexWithoutPad } from '../helpers/format';
 
-const getChainIdFromRpcUrl = async function (rpcUrl: string): Promise<string> {
+export const getDecimalChainIdFromRpcUrl = async function (rpcUrl: string): Promise<number> {
     const provider = new ethers.providers.StaticJsonRpcProvider({ url: rpcUrl });
     const networkInfo = await provider.getNetwork();
     if (!networkInfo || !networkInfo.chainId) {
         throw new Error(`Can not verify RPC url`);
     }
-    return formatToHexWithoutPad(networkInfo.chainId);
+    return networkInfo.chainId;
 };
 
-const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
+const getChainIdFromRpcUrl = async function (rpcUrl: string): Promise<string> {
+    return formatToHexWithoutPad(await getDecimalChainIdFromRpcUrl(rpcUrl));
+};
+
+export const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
     const result = /infura\.io\/v3\/(\S+)/gm.exec(rpcUrl);
     return result?.[1] ?? undefined;
 };
--- a/core/src/rpc.ts
+++ b/core/src/rpc.ts
@@ -9,20 +9,16 @@ import {
 } from './network';
 import { formatToHexWithoutPad } from '../helpers/format';
 
-export const getDecimalChainIdFromRpcUrl = async function (rpcUrl: string): Promise<number> {
+const getChainIdFromRpcUrl = async function (rpcUrl: string): Promise<string> {
     const provider = new ethers.providers.StaticJsonRpcProvider({ url: rpcUrl });
     const networkInfo = await provider.getNetwork();
     if (!networkInfo || !networkInfo.chainId) {
         throw new Error(`Can not verify RPC url`);
     }
-    return networkInfo.chainId;
+    return formatToHexWithoutPad(networkInfo.chainId);
 };
 
-const getChainIdFromRpcUrl = async function (rpcUrl: string): Promise<string> {
-    return formatToHexWithoutPad(await getDecimalChainIdFromRpcUrl(rpcUrl));
-};
-
-export const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
+const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
     const result = /infura\.io\/v3\/(\S+)/gm.exec(rpcUrl);
     return result?.[1] ?? undefined;
 };
--- a/frontend/lib/wallets/WalletConnect.ts
+++ b/frontend/lib/wallets/WalletConnect.ts
@@ -27,7 +27,6 @@ export default class WalletConnect extends AbstractWallet {
                 (accumulatorObj, network) => ({ ...accumulatorObj, [parseInt(network.chainId, 16)]: network.url }),
                 {}
             );
-            console.warn(rpcMap);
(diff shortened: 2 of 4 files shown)


frontend/lib/wallets/WalletConnect.ts
```diff
@@ -1,32 +1,88 @@
...
+        if (Array.isArray(this.addresses) && this.addresses[0]) {
+            return this.addresses[0].toLowerCase();
+        }
+    }
+
+    async getProvider(): Promise<ethers.providers.JsonRpcProvider> {
+        const infuraId = process.env.INFURA_PROJECT_ID;
+        if (!infuraId) {
+            throw new Error(`No INFURA_PROJECT_ID env variable was provided`);
+        }
```
valiafetisov, line 27: Initially [you claimed](https://github.com/sidestream-tech/unified-auctions-ui/issues/353#issuecomment-1309041473) that it can work with regular RPC urls if the chainId is provided. Has this changed?

Otherwise the requirement of the issue stands to _not_ introduce any new env variables. We already have `RPC_URL` and that should be enough to provide it here either as a single url or as multiply urls derived from it (see `getDefaultNetworkConfigs`).

To be clear, this feature should also _not_ introduce new requirement to have `PRC_URL` provider to be infura. It should work with any

frontend/package.json
```diff
@@ -48,7 +49,8 @@
     "uuid": "^8.3.2",
     "vue-cryptoicon": "^0.18.0-a",
     "vue-plausible": "^1.2.1",
-    "vuex-persist": "^3.1.3"
+    "vuex-persist": "^3.1.3",
+    "web3": "^1.8.1"
```
valiafetisov, line 53: Why web3 is required? Can we use `ethers.js` instead?

core/src/rpc.ts
```diff
@@ -9,16 +9,20 @@ import {
...
-    return formatToHexWithoutPad(networkInfo.chainId);
+    return networkInfo.chainId;
+};
+
+const getChainIdFromRpcUrl = async function (rpcUrl: string): Promise<string> {
+    return formatToHexWithoutPad(await getDecimalChainIdFromRpcUrl(rpcUrl));
 };
 
-const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
+export const parseInfuraProjectIdFromRpcUrl = function (rpcUrl: string): string | undefined {
```
valiafetisov, line 25: Is any of the changes to this file still necessary?

frontend/lib/wallets/WalletConnect.ts
```diff
@@ -1,32 +1,97 @@
...
+    public get address() {
+        if (Array.isArray(this.addresses) && this.addresses[0]) {
+            return this.addresses[0].toLowerCase();
+        }
+    }
+
+    async getProvider(): Promise<ethers.providers.JsonRpcProvider> {
+        if (!WalletConnect.provider) {
+            const rpcUrl = process.env.RPC_URL;
+            const { networks } = await setupRpcUrlAndGetNetworks(rpcUrl);
```
valiafetisov, line 27: If you only need networks, you can just call `getNetworks` here, since setup will also do other things that you don't need.
```suggestion
            const networks = awat getNetworks();
```

frontend/lib/wallets/WalletConnect.ts
```diff
@@ -1,32 +1,97 @@
...
+            let rpcMap = {};
+            for (const network of networks) {
+                const chainIdDecimal = parseInt(network.chainId, 16);
+                rpcMap = {
+                    ...rpcMap,
+                    [chainIdDecimal]: network.url,
+                };
+            }
```
valiafetisov, line 35: Nit: you can also use reduce instead of for loop:
```ts
const obj = yourArray.reduce((o, key) => ({ ...o, [key]: whatever}), {})
```

frontend/lib/wallets/WalletConnect.ts
```diff
@@ -1,32 +1,97 @@
...
-        throw new Error(`${WalletConnect.title} network switching is not yet implemented`);
+    public async networkChangedHandler() {
+        if (!WalletConnect.provider) {
+            return;
+        }
+        const signer = await this.getSigner();
+        const chainId = formatToHexWithoutPad(await signer.getChainId());
+        const networkType = getNetworkTypeByChainId(chainId);
+        if (networkType) {
+            setSigner(networkType, signer as any);
```
valiafetisov, line 69: Why typecasting?

frontend/lib/wallets/WalletConnect.ts
```diff
@@ -1,32 +1,97 @@
...
+        WalletConnect.provider.on('accountsChanged', this.accountsChangedHandler.bind(this));
+        WalletConnect.provider.on('chainChanged', this.networkChangedHandler.bind(this));
+    }
 
-    teardown() {}
+    public async teardown() {
+        if (!WalletConnect.provider) {
+            return;
+        }
+        window.$nuxt.$store.dispatch('network/setWalletChainId', undefined);
```
valiafetisov, line 91: When is this called?

frontend/lib/wallets/WalletConnect.ts
valiafetisov: If the problem is only with the promise, you can always use `Promise.resolve()`, eg
```suggestion
            setSigner(networkType, Promise.resolve(signer));
```

valiafetisov:
- [ ] Were you able to execute any transactions with the new wallet, eg on the goerli?
- [ ] Where can I find the description of the problem related to simulations or was it resolved?

valiafetisov:
> they should be added first with a `wallet_addEthereumChain` request

Have you tried that?

valiafetisov:
I'm fine with merging it without network switch or support of the localhost (this one is not even available in production) if the flow in general works for @LukSteib

LukSteib (approved):

PR #535, opened 2022-11-14
Improve format date component

### LINKED ISSUE #523: Improve date indicator for next price drop
### Goal
Don't show "In invalid date" any longer for the next price drop on the vault page

### Context
For some collateral types (experienced during tests for `USDC-A`) we show "In invalid date" as an indicator for the next price dop.
We might approve that by indicating that the date is not available for these cases instead of cryptic invalid date that might confuse the user

### Assets
`USDC-A` vault creation simulation
image

### Tasks
- [ ] Adjust the information on next price drop with a more speaking information

zoey-kaiser:
Closes #523 

If the date is not a javascript date object the following is displayed:

Screenshot 2022-11-14 at 17 31 10

I added a new story to check this behavior -> `Vault/Vault/Invalid Date` 

Checklist:

### DIFF
--- a/frontend/components/common/formatters/TimeTill.stories.js
+++ b/frontend/components/common/formatters/TimeTill.stories.js
@@ -23,4 +23,8 @@ storiesOf('Common/Formatters/TimeTill', module)
             date: new Date(Date.now() + 5000).toUTCString(),
         }),
         template: `<TimeTill :date="date" />`,
+    }))
+    .add('Invalid date', () => ({
+        components: { TimeTill },
+        template: `<TimeTill date="Invalid Date" />`,
     }));
--- a/frontend/components/vault/Vault.vue
+++ b/frontend/components/vault/Vault.vue
@@ -30,14 +30,20 @@
                         <tr>
                             <td>Next price update</td>
                             <td>
-                                <div v-if="vaultTransaction.nextPriceChange" class="flex items-center space-x-1">
+                                <div class="flex items-center space-x-1">
                                     <AnimatedArrow
                                         :direction="getIsPriceGoingUpOrDown(vaultTransaction)"
                                         class="h-4 opacity-50"
                                     />
-                                    <span>in <TimeTill :date="vaultTransaction.nextPriceChange" /></span>
+                                    <span
+                                        v-if="
+                                            vaultTransaction.nextPriceChange &&
+                                            vaultTransaction.nextPriceChange instanceof Date
+                                        "
+                                        >in <TimeTill :date="vaultTransaction.nextPriceChange"
+                                    /></span>
+                                    <span v-else class="opacity-50">Unknown</span>
                                 </div>
-                                <span v-else class="opacity-50">Unknown</span>
                             </td>
                         </tr>
                         <tr>
--- a/frontend/components/vault/Vault.stories.js
+++ b/frontend/components/vault/Vault.stories.js
@@ -46,4 +46,17 @@ storiesOf('Vault/Vault', module)
     .add('Not found', () => ({
         ...common,
         template: `<Vault :vaultId='vaultId' />`,
+    }))
+    .add('Invalid Date', () => ({
+        ...common,
+        data() {
+            return {
+                vault: {
+                    ...fakeVaultNotLiquidatedTransaction,
+                    nextPriceChange: 'Invalid Date',
+                },
+                vaultId: fakeVaultLiquidatedTransaction.id.toString(),
+            };
+        },
+        template: `<Vault :vaultTransaction="vault" :vaultId='vaultId' @liquidate="liquidate" />`,
     }));


zoey-kaiser:
> I like that you made changes to the `TimeTill` itslef!
> 
> But why do you decide to keep arrow pointing upwards even if the date is unknown?

I began by doing it (therefore also the changes). But I noticed TimeTill also supports strings as a date fallback. If I introduce the same checks as in the `Vault` component we would lose that support. In addition to this, we use different `unknown` designs in parts of our UI, making it more realistic to check it when entering it, as we may need to design it differently based on if it is in a table, popover hove etc.

LukSteib:
> But why do you decide to keep arrow pointing upwards even if the date is unknown?

I cannot derive an answer to that from your last comment @zoey-kaiser. My suggestion on that particular one: Let's remove the arrow in case the date is `Unknown`. Otherwise confusing from a user perspective.

>  But I noticed TimeTill also supports strings as a date fallback. If I introduce the same checks as in the Vault component we would lose that support.

When it comes to this refacroting related discussion I cannot add too much. What is the advantage of having strings as fallback which we would loose in case of your initially intended refactoring (if I understand correctly)?

zoey-kaiser:
> I cannot derive an answer to that from your last comment @zoey-kaiser. My suggestion on that particular one: Let's remove the arrow in case the date is `Unknown`. Otherwise confusing from a user perspective.

I really don't know how I misinterpreted @valiafetisov question like that 😓 I must have been a bit off that day. I will remove the arrow when we don't have the date estimate!

frontend/components/vault/Vault.vue
```diff
@@ -30,7 +30,13 @@
                         <tr>
                             <td>Next price update</td>
                             <td>
-                                <div v-if="vaultTransaction.nextPriceChange" class="flex items-center space-x-1">
+                                <div
+                                    v-if="
+                                        vaultTransaction.nextPriceChange &&
+                                        vaultTransaction.nextPriceChange instanceof Date
```
valiafetisov, line 36: What else can it be besides `instanceof Date`?

According to the type it is only allowed to be `Date` anyway: https://github.com/sidestream-tech/unified-auctions-ui/blob/24a1932c9eb9e12c69f052a0b1788574cdda5833/core/src/types.ts#L375

Maybe you wanted to check it against `NaN` instead?

frontend/components/common/formatters/TimeTill.stories.js
```diff
@@ -23,4 +23,8 @@ storiesOf('Common/Formatters/TimeTill', module)
             date: new Date(Date.now() + 5000).toUTCString(),
         }),
         template: `<TimeTill :date="date" />`,
+    }))
+    .add('Invalid date', () => ({
+        components: { TimeTill },
+        template: `<TimeTill date="Invalid Date" />`,
```
valiafetisov, line 29: Nice that you've added a new story, but is it actually handled now? Does it show `Unknown` in case date is `undefined` or `NaN`? Otherwise I would suggest to do it in the `TimeTill.vue` – then you don't need to check it every time you use `TimeTill` – it will be handled automatically (besides the cases like with the arrow)

frontend/components/vault/Vault.vue
zoey-kaiser: Hmm, Honestly I do not know the type that was passed when the error occurred (as I could not reproduce it and it was not given in the issue). 

Therefore I went out on a limb and hoped that checking it against Date would be enough. In the end it is only a theoretical decision (as it should always be date anyway). If we had a way to reproduce the bug this would be helpful. Otherwise we could also go ahead and check it against date.

frontend/components/common/formatters/TimeTill.stories.js
zoey-kaiser: I was unsure of this. Officially the component does support other types such as Strings. I don't think we currently use this in the UI at all, however it would be nice to verify this. 

Currently our fallback for the component if it can't be turned into a Date is to show the String. I think this should be fine for now, as we almost always check the value we pass to the component in the other place it is used. 

We also have different designs for how we display the "unknown" text depending on where it is. Therefore I believe it makes little sense to incorporate the error logic into the component itself.

frontend/components/vault/Vault.vue
valiafetisov: > as I could not reproduce it and it was not given in the issue

Oh wow! Next time please raise this in the issue instead of submitting a PR which doesn't fix anything!

The `Invalid Date` is shown whenever a stablecoin vault is fetched (as indicated in the issue). This is where the invalid date is produced:
https://github.com/sidestream-tech/unified-auctions-ui/blob/c3c3cc4ed3fe7d7cf408ceca9fd44a4f9aec143b/core/src/oracles.ts#L116

Therefore, please check for NaN instead of the instanceof

frontend/components/common/formatters/TimeTill.stories.js
valiafetisov: - [ ] Please follow the advice and set `Unknown` to be displayed as a fallback instead of `Invalid Date` or the string
  - [ ] The original input can be set as the `title` html attribute for "debugging" purposes

> We also have different designs for how we display the "unknown" text depending on where it is

If we need to display another text, there can be an additional check in the parent

valiafetisov (dismissed):

LukSteib (changes requested):
- [ ] while we've remove the price update arrow indicator on the vault page itself we still keep showing the arrow on the vault liquidation page (see screenshot below). In case of `Unknown` I'd suggest to remove the indicator there as well

image

LukSteib:
Problem reported above still apparent

LukSteib:
>  Can you guide me on how to start it?

1. inside `/core` run `npm run simulate`
2. select simulation `Simulate collateral auction per collateral type`
3. select `USDC-A` 
4. Execute first step `Create the vault`

At this point in time the simulation provides you with a vault ID that you can then go an check via UI

LukSteib (approved):

valiafetisov (approved):

PR #537, opened 2022-11-14
Quick fix: Switch estimated profitability time

### LINKED ISSUE #533: Discuss time till profitability
### Goal
Callee selection affects time till profitability

### Context
We have introduced multi callee support that allows a user to select between different exchange callees. These come with different market price. 
So ultimately one can expect that the "Time till profitability" indicator would need to change based on the callee selection. 
Currently unsure whether we take this into account. 
If the market prices of different callees only slightly differ chances are high that auction price will drop below all with the same price drop step. 
However if prices of different callees differ quite a lot the time till profitability might also be different since the thresholds won't be reached with the same price drop

Let's use this issue to discuss whether the assumption above is correct and if so how to address

### Tasks

- [x] Discuss effect of callee selection on "Time till profitability"
- [x] (if needed)Outline approach for adjustment
- [x] (if needed) Implement adjustment

aomafarag:
Closes #533.

Checklist:

### DIFF
--- a/frontend/components/auction/collateral/CollateralAuctionSwapTransactionTable.vue
+++ b/frontend/components/auction/collateral/CollateralAuctionSwapTransactionTable.vue
@@ -34,7 +34,7 @@
         <div class="flex w-full justify-between">
             <div>Estimated Profitability Time</div>
             <div class="RightInfo">
-                <time-till-profitable :auction="auctionTransaction" />
+                <TimeTillProfitable :auction="auctionTransaction" :market-id="currentMarketId" />
             </div>
         </div>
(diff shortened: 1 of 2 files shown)


LukSteib (commented):
Hard for me to test functionally since `goerli` auctions currently cannot be restarted. For the simulation based auctions I don't see a change in time when switching callees. However, the testing case here in general is not really realistic when it comes to the times. (see below)

Hence fine for me if @valiafetisov accepts from technical perspective.

https://www.loom.com/share/a622c1636d884ca8be4bda747ff8bbef

valiafetisov (approved):

PR #546, opened 2022-11-21
Support Coinbase Wallet

### LINKED ISSUE #545: Add Coinbase Wallet
### Goal

Add Coinbase Wallet to our UI.

### Context

Currently, we support both MetaMask and WalletConnect wallets. Additionally, we would like to support Coinbase Wallet.

### Assets

Coinbase Wallet SDK docs: https://docs.cloud.coinbase.com/wallet-sdk/docs

### Tasks

- [x] Read the docs
- [x] Understand how to implement the wallet
- [x] Integrate the wallet in our codebase

aomafarag:
Closes #545.

Screenshot from 2022-11-21 16-16-18

Checklist:

---

### Remarks

- To test the wallet, you can use either the mobile app or the browser extension (Google Chrome only)
- The wallet [does not support switching accounts](https://github.com/coinbase/coinbase-wallet-sdk/issues/23#issuecomment-593692744) (account is fixed throughout the connection)
- Although switching to our `custom` network works, carrying out transactions on it does not.
- Even when one has `GoerliETH` on the wallet, the balance appears as `$0.00`. I guess this has to do with the fact that one has to link their Coinbase account to their wallet in order to deal in cryptos. Nevertheless, I managed to transfer funds to the wallet on `goerli`, and successfully carried out a transaction.


florianstoecker (changes requested):
- [x] In this popup, the text collides with the red x symbol
  image

- [x] When the wallet is set to main net, but UI is set to Goerli & then connecting the wallet, clicking on `main` within this popup, will result in the auction table showing some finished auctions which I cannot make sense of. What is this happening here?
  image

LukSteib:
> However, we are using the Alert component from Ant Design directly to display this, and I can't seem to find anything in their API that could fix this.

As this is unrelated to the PR at hand I will create a dedicated issue for that.

> I could not replicate this. Navigating to mainnet shows No active auctions.

I am able to replicate this problem of fetching / displaying wrong auctions (see loom below). So in the case at hand: `Goerli` auctions are displayed when switching the UI to `main`. Might be caching problem?! Let's spend some time to investigate as it might be related to the wallet integration.

https://www.loom.com/share/e4f33d1311c94427a603496ef0725481

LukSteib (approved):

frontend/assets/icons/wallets/metamask.svg
```diff
@@ -1 +1,44 @@
...
+<svg version="1.1" id="Layer_1" xmlns:ev="http://www.w3.org/2001/xml-events"
```
valiafetisov, line 1: Why was the change of this svg necessary? If you need to keep the new version, please remove `xmlns:xlink`, `xmlns:ev`, `xml:space` and other non-html compatible or [deprecated](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/xml:space) html arguments

svgo is a useful tool for this 
 - https://marketplace.visualstudio.com/items?itemName=1000ch.svgo
 - https://jakearchibald.github.io/svgomg/

frontend/assets/icons/wallets/coinbase.svg
```diff
@@ -0,0 +1 @@
+<svg viewBox="0 0 1024 1024" fill="none" xmlns="http://www.w3.org/2000/svg"><rect width="1024" height="1024" fill="#0052FF"/><path fill-rule="evenodd" clip-rule="evenodd" d="M152 512C152 710.823 313.177 872 512 872C710.823 872 872 710.823 872 512C872 313.177 710.823 152 512 152C313.177 152 152 313.177 152 512ZM420 396C406.745 396 396 406.745 396 420V604C396 617.255 406.745 628 420 628H604C617.255 628 628 617.255 628 604V420C628 406.745 617.255 396 604 396H420Z" fill="white"/></svg>
```
valiafetisov, line 1: Please add new line in the end of the file

frontend/lib/wallets/Coinbase.ts
```diff
@@ -0,0 +1,81 @@
...
+        }
+    }
+
+    getProvider(): ethers.providers.JsonRpcProvider {
+        const rpcUrl = process.env.RPC_URL;
+        if (!rpcUrl) {
+            throw new Error(`No RPC_URL env variable was provided`);
+        }
+        if (!Coinbase.provider) {
+            const walletSdk = new CoinbaseWalletSDK({ appName: 'Unified Auctions UI' });
```
valiafetisov, line 30: Maybe you can provide [SITE_TITLE from the nuxt config](https://github.com/sidestream-tech/unified-auctions-ui/blob/6cdba5def95207b3e1f4178878c1678124b18d63/frontend/nuxt.config.js#L4) via env vars if you don't want to hardcode it here

frontend/lib/wallets/Coinbase.ts
```diff
@@ -0,0 +1,81 @@
...
+    private addresses?: string[];
+
+    public get address() {
+        if (Array.isArray(this.addresses) && this.addresses[0]) {
+            return this.addresses[0].toLowerCase();
+        }
+    }
+
+    getProvider(): ethers.providers.JsonRpcProvider {
+        const rpcUrl = process.env.RPC_URL;
```
valiafetisov, line 25: If you take RPC_URL directly from the env var, how does the network switch suppose to work? `RPC_URL` is different per network and as I several times pointed out, you should instead use
```ts
const rpcUrl = getNetworks()[netoworkName].url
```

frontend/assets/icons/wallets/metamask.svg
```diff
@@ -1 +1,44 @@
...
+<g>
+	<polygon fill="#763D16" stroke="#763D16" stroke-linecap="round" stroke-linejoin="round" points="278.3,114.2 286.8,73.4 274.1,35.5 177.9,106.9 214.9,138.2 267.2,153.5 278.8,140 273.8,136.4 281.8,129.1 275.6,124.3 283.6,118.2" />
+	<polygon fill="#763D16" stroke="#763D16" stroke-linecap="round" stroke-linejoin="round" points="31.8,73.4 40.3,114.2 34.9,118.2 42.9,124.3 36.8,129.1 44.8,136.4 39.8,140 51.3,153.5 103.6,138.2 140.6,106.9 44.4,35.5" />
+</g>
+<polygon fill="#F6851B" stroke="#F6851B" stroke-linecap="round" stroke-linejoin="round" points="267.2,153.5 214.9,138.2 230.8,162.1 207.1,208.1 238.3,207.7 284.8,207.7" />
+<polygon fill="#F6851B" stroke="#F6851B" stroke-linecap="round" stroke-linejoin="round" points="103.6,138.2 51.3,153.5 33.9,207.7 80.3,207.7 111.4,208.1 87.8,162.1" />
+<polygon fill="#F6851B" stroke="#F6851B" stroke-linecap="round" stroke-linejoin="round" points="174.6,164.6 177.9,106.9 193.1,65.8 125.6,65.8 140.6,106.9 144.1,164.6 145.3,182.8 145.4,227.6 173.1,227.6 173.3,182.8" />
+</svg>
```
valiafetisov, line 44: New line is missing

frontend/assets/icons/wallets/metamask.svg
valiafetisov: I'm not sure what you mean, production always showed icons for me:

Screenshot 2022-11-30 at 19 56 52

Which browser you're using?

frontend/lib/wallets/Coinbase.ts
```diff
@@ -1,8 +1,9 @@
 import CoinbaseWalletSDK, { CoinbaseWalletProvider } from '@coinbase/wallet-sdk';
 import { ethers } from 'ethers';
-import { getChainIdByNetworkType, getNetworkTypeByChainId } from 'auctions-core/src/network';
+import { getChainIdByNetworkType, getNetworkConfigByType, getNetworkTypeByChainId } from 'auctions-core/src/network';
 import { setSigner } from 'auctions-core/src/signer';
 import { formatToHexWithoutPad } from 'auctions-core/helpers/format';
+import { SITE_TITLE } from '~/nuxt.config';
```
valiafetisov, line 6: Let's not import `nuxt.config` into our files, since it shouldn't be available after build and/or can introduce circular dependencies. Instead, as suggested, pass `SITE_TITLE` to the `env` object and then use `process.env.SITE_TITLE` here

frontend/lib/wallets/Coinbase.ts
```diff
@@ -21,13 +22,19 @@ export default class Coinbase extends AbstractWallet {
...
+        let rpcUrl: string | undefined;
+        try {
+            rpcUrl = getNetworkConfigByType(network).url;
+        } catch {
+            rpcUrl = process.env.RPC_URL;
+        }
         if (!rpcUrl) {
             throw new Error(`No RPC_URL env variable was provided`);
         }
+        console.warn(`RPC_URL: ${rpcUrl}`);
```
valiafetisov, line 35: Please remove unused console logs. Also, it's recommended to use `console.log` or debugging mode for debugging

frontend/lib/wallets/Coinbase.ts
valiafetisov: Yes

frontend/lib/wallets/Coinbase.ts
```diff
@@ -0,0 +1,86 @@
...
+        }
+    }
+
+    getProvider(network?: string): ethers.providers.JsonRpcProvider {
+        let rpcUrl: string | undefined;
+        try {
+            rpcUrl = getNetworkConfigByType(network).url;
+        } catch {
+            rpcUrl = process.env.RPC_URL;
+        }
```
valiafetisov, line 30: Why did you still decided to fallback into it pure `process.env.RPC_URL`? in case of the network change it can lead to issues

valiafetisov (approved):

frontend/lib/wallets/Coinbase.ts
valiafetisov: I think that's not responsibility of the wallet to initialise `env` variables (otherwise every wallet will have to do so). Instead, 
1. Wallets doesn't need to be initialised at all before user action
2. In case the wallet was previously connected (and we remember that), wallets doesn't need to be set up until network is set up (`wallet/connect` should be called only when network is there)

2nd is the problem outside of the scope of this PR. Or does 1st also applies here?

valiafetisov: > we are just initializing (there's no switch), so the network parameter is undefined

Can you please provide network provider during the initialization (so not taken from the store, but provided with the argument to the `connect()`)? Your solution would work too, but hacky!

LukSteib (commented):
- [x] When trying to execute a tx on `goerli` with coinbase wallet connected I am getting the gas related error known from the past (with metamask wallet connected everything works as expected)

`Auction redo error: err: max fee per gas less than block base fee: address 0x0E1FbEa3b3a96311b93dAd0Da488B3C2dcEa6463, maxFeePerGas: 1500688016 baseFee: 11262064220 (supplied gas 0)`

LukSteib (commented):
- [ ] Cannot reproduce the problem reported above anymore. However, now I am getting another error `underlying network changed` although both UI and wallet apparently are on the same network (ie. `goerli`). See recording and console error below

https://www.loom.com/share/b441e4de6de14a87859174328f460ca0

console error: 
```
auctions.ts?b334:324 Auction redo error: underlying network changed (event="changed", network={"name":"homestead","chainId":1,"ensAddress":"0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e"}, detectedNetwork={"name":"goerli","chainId":5,"ensAddress":"0x00000000000C2E074eC69A0dFb2997BA6C7d2e1e"}, code=NETWORK_ERROR, version=providers/5.4.5)
```

frontend/store/wallet.ts
```diff
@@ -154,12 +154,13 @@ export const actions = {
...
         }
     },
-    async connect({ commit, dispatch }: ActionContext<State, State>, walletType: string): Promise<void> {
+    async connect({ commit, dispatch, rootGetters }: ActionContext<State, State>, walletType: string): Promise<void> {
         dispatch('authorizations/reset', undefined, { root: true });
         commit('setIsConnecting', true);
         try {
             const wallet = getWallet(walletType);
-            await wallet.connect();
+            const network = rootGetters['network/getMakerNetwork'];
```
valiafetisov, line 162: Why did you choose to use `getMakerNetwork` over `getPageNetwork` that [you've initially planned to use](https://github.com/sidestream-tech/unified-auctions-ui/pull/546#discussion_r1037359342)? 

`getMakerNetwork` is the value that "combines" both wallet and page networks, and it might be tricky to use here

frontend/lib/wallets/Coinbase.ts
```diff
@@ -0,0 +1,84 @@
...
+        if (Array.isArray(this.addresses) && this.addresses[0]) {
+            return this.addresses[0].toLowerCase();
+        }
+    }
+
+    getProvider(network: string): ethers.providers.JsonRpcProvider {
+        const rpcUrl = getNetworkConfigByType(network).url;
+        if (!Coinbase.provider) {
+            const walletSdk = new CoinbaseWalletSDK({ appName: process.env.SITE_TITLE || '' });
+            Coinbase.provider = walletSdk.makeWeb3Provider(rpcUrl);
```
valiafetisov, line 33: Btw, what RPC url _network_ it expects here? Any, on which the website is? Or the url of the network on which the wallet is? Then, how do we get the network which network the wallet is currently on?

On a bigger picture, with metamask for example, we first connect the wallet on whatever network the metamask is and only then we suggest network change (in our UI) in case there is mismatch

LukSteib (approved):

frontend/lib/wallets/Coinbase.ts
valiafetisov: Would you please still answer the question? 🙂

valiafetisov (approved):

frontend/lib/wallets/Coinbase.ts
valiafetisov: > because we cannot obtain the wallet's network unless we connect it

Well, I think this only depends on the API of the wallet, so it can be possible

Anyway, I would open an issue regarding this for the `WalletConnect`: to provide correct rpc url during connection and after the network switch

Issue #245, opened 2022-05-05
Setup deployment for the new `notification-service`


zoey-kaiser:
### Goal
In #240 we added a new service to our project called the `notification-service`. In that PR we added a very "barebones" implementation of the CI to check the linting of the project. This issue is meant to address the deployment of this service. 

### Context
In addition to setting up the pipeline we also need to configure the following secrets: 

- [ ] INFURA_PROJECT_ID - same as on bot
- [ ] ETHERSCAN_API_KEY - can be created for free [here](https://docs.etherscan.io/getting-started/viewing-api-usage-statistics) 
- [ ] SMTP Settings - We need to have some sort of SMTP server with which we can send emails
  - [ ] SMTP_HOST
  - [ ] SMTP_PORT
  - [ ] SMTP_USERNAME
  - [ ] SMTP_PASSWORD
  - [ ] SMTP_EMAIL

### Tasks

- [ ] Add deployment steps to the project
- [ ] Configure all needed secrets + SMTP server
- [ ] Setup receivers list (who wants to recieves notification emails? See docs [here](https://github.com/sidestream-tech/unified-auctions-ui/blob/b0fef054a27a7f8381decea0dde04bb3e7289a94/notification-service/README.md#environment-variables))

Issue #536, opened 2022-11-14
Cleanup the repo


valiafetisov:
### Goal

Unused services removed

### Context

We've accumulated quite some code since the start of the project. This is especially visible by the amount of folders on in the root directly. The proposal is to remove unused/stale code. More specifically:

- [x] Remove `/chaos-proxy` folder
- [x] Remove `/reverse-proxy` folder (since we actually don't need docker-compose to set project up)
  - [x] Remove `/docker-compose.yaml`
- [x] Decide if we want to move `notification-service` to another repo or just remove it (together with @LukSteib), since it's out of scope of the original repo
- [x] Propose other changes in the issue (eg make sure all docs are up-to-date)

### Tasks

- [x] Open a new PR for each major removal so we can keep of those in the commit history

LukSteib:
>  Decide if we want to move notification-service to another repo or just remove it (together with @LukSteib), since it's out of scope of the original repo

I would suggest to move it to another repo. This gives us the ability to decide wether to deploy and actually run it or not at some later point

zoey-kaiser:
> I would suggest to move it to another repo. This gives us the ability to decide wether to deploy and actually run it or not at some later point

Do I create a new repo under `sidestream-tech` for this or should I contact someone else? 

Todo for moving the notification service:

- [ ] Create a new repository named `maker-notification-service` (or another name)
- [ ] Move the files into the new repository
- [ ] Adjust the Readme (if needed)
- [ ] Remove the `notification service` from `sidestream_tech/unified-auctions-ui`

valiafetisov:
> Do I create a new repo under sidestream-tech for this or should I contact someone else?

Can you please first validate if you can actually split those? Does `notification-service` heavily depends on the core (if at all)? If so, what would be the way forward to remove/replace the dependency?

zoey-kaiser:
> Can you please first validate if you can actually split those? Does `notification-service` heavily depends on the core (if at all)? If so, what would be the way forward to remove/replace the dependency?

It does not depend on the core! https://github.com/sidestream-tech/unified-auctions-ui/blob/main/notification-service/package.json#L30

Therefore a move should be very straightforward.

PR #547, opened 2022-11-22
Removed notification service

zoey-kaiser:
Contributes to #536
Closes https://github.com/sidestream-tech/unified-auctions-ui/issues/245

Checklist:


zoey-kaiser:
To ensure we don't lose to existing code I moved the `notification-service` to a repository here: 

https://github.com/zoey-kaiser/notification-service

valiafetisov:
- [ ] The new repo is missing licence file, which I believe obligatory
- [ ] I would also explain and link the original repo `sidestream-tech/unified-auctions-ui` in the readme
- [ ] I don't think it's the intended result to move `notification-service` to your private repo, although I don't feel to be the right person to comment on it. I expected that it should still belong to @sidestream-tech, but might be archived immediately since we don't plan to support it as of now, but still want to preserve the work. I guess, if you want to, you can fork it from there. But as I am not sure, I would pull @BracketJohn to resolve this question 🙂 

A little context for @BracketJohn: we're cleaning up the repository from unused or unrelated code as outlined in #536 and one of the proposals was to move `notification-service` into another repository. Maybe you have a another suggestion than outlined above?

BracketJohn:
>  But as I am not sure, I would pull @BracketJohn to resolve this question 🙂

Yeah, let's please not move the code out of the sidestream-org,  instead:
1. commit 1: make an exact copy of the auctions ui and push it into a new repo `notification-service` (you could also do this by adding the new, empty repo as a new origin, that would preserve history but I don't think that that has any implications)
2. commit 2 (can be done via squashed PR to allow reviews):  remove everything besides the notification service + relevant meta-docs (license, readme, ....) you may want to keep in one move commit (still no alterations!)
4. do any alterations you may wish to do, e.g., by adding a fat disclaimer at the top of the readme with an explanation of the situation (eg see https://github.com/sidebase/sidebase-libs)
5. archieve the repo
6. delete the code in this repo, reference the moved location at least in the issue / PR, but maybe also somewhere in the readme (like: Archieved Services: notification-service, moved unaltered with same licensed status to: URL)

This should ensure the integrity of licensing, originality, ... much better than the current approach. Ensuring this integrity is important to make sure that the code stays in the public domain _clearly_, i.e., people should have a look at the notification-service and feel comfortable using it. This would not be the case with your current copy of the repo, as the License is missing. And even if you add a license now: From when does it apply? There's bigger disputes in the OS-community about sudden license changes with unclear implications.

Let me know if the above is too dense or if you disagree

zoey-kaiser:
I create a new repository and moved the code there. Afterwards I added a disclaimer and and archived the repository.

The license was never changed and I added links back to the "real" unified-auctions-ui.

You can find the new repository here: https://github.com/sidestream-tech/notification-service

I think we can now go ahead with the removal of the notification-service in this PR!

valiafetisov (approved):

PR #549, opened 2022-11-23
Adjust table width

### LINKED ISSUE #502: Improve Table widths
### Goal
We noticed that when switching between views on the pages, the column width of the table would change greatly. We want to introduce fixed width (or min widths) that fix this issue. 

However, we need to keep in mind that the table still works on:
- Mobile view
- Explanations on mode
- Explanations off mode

### Assets

https://user-images.githubusercontent.com/30908158/196419818-7c79da72-be71-41a3-8f4c-002bf5195ee7.mov

### Tasks

- [x] Introduce a fix
- [x] Test it in all circumstances
  - [x] Mobile View
  - [x] Explanations On mode
  - [x] Explanations Off mode

aomafarag:
Closes #502.

[table_width.webm](https://user-images.githubusercontent.com/85403429/204101515-52e8c4be-151d-486d-a734-cc8b34c5f890.webm)

Checklist:


frontend/components/common/formatters/TimeTill.vue
```diff
@@ -40,10 +50,15 @@ export default Vue.extend({
         calculateTime(): void {
             if (this.parsedDate) {
                 const now = new Date();
-                this.timeTill = formatInterval(now, this.parsedDate);
-
-                const duration = this.parsedDate.getTime() - now.getTime();
-                this.isEndingSoon = duration < ENDING_SOON_THRESHOLD && duration > 0;
+                if (this.isCountUp) {
+                    ++this.secondsElapsed;
+                    const newDate = new Date(now.getTime() + this.secondsElapsed * 1000);
```
valiafetisov, line 55: Why is `secondsElapsed` needed? 

The whole purpose of `parsedDate` is to not rely on `calculateTime` being called every second (which is not guaranteed by the browser). So why do you introduce a completely different logic to calculate time for `isCountUp` and completely ignore value of the `parsedDate` prop when in actuality you only need to change formatting?

frontend/components/auction/surplus/SurplusAuctionsTable.vue
```diff
@@ -10,6 +10,8 @@
             :custom-row="customRowEvents"
             :get-popup-container="() => $el"
             :locale="{ emptyText: 'No active auctions' }"
+            :table-layout="auto"
+            :scroll="{ x: 'max-content' }"
```
valiafetisov, line 14: What is `max-content` string? According to the docs, `scroll` keys only accept numbers or a boolean:
Screenshot 2022-11-28 at 09 29 32
https://2x.antdv.com/components/table#API

frontend/helpers/tillDuration.ts
```diff
@@ -20,11 +20,12 @@ export const formatSeconds = function (seconds: number): string {
     return formatDuration(duration);
 };
 
-export const formatInterval = function (startDate: Date, endDate: Date): string {
-    if (startDate < endDate) {
+export const formatInterval = function (startDate: Date, endDate: Date, isCountUp: boolean = false): string {
+    if (startDate < endDate || isCountUp) {
+        // switch startDate and endDate to act as a count-up timer
```
valiafetisov, line 25: Do we actually use `formatDistance` format anywhere else then? Otherwise I would suggest to completely remove `formatDistance` hack from here completely and use `0h 0m 0s` format everywhere

valiafetisov:
> should we make `TimeTill` a count-up timer for the `Last updated` row in [`WalletTable`](https://github.com/sidestream-tech/unified-auctions-ui/blob/main/frontend/components/wallet/WalletTable.vue#L46) as well?

Yes

frontend/helpers/tillDuration.ts
```diff
@@ -20,17 +20,26 @@ export const formatSeconds = function (seconds: number): string {
...
-            end: isCountUp ? startDate : endDate,
+            start: startDate,
+            end: endDate,
         });
         return formatDuration(duration);
     } else {
-        return formatDistance(endDate, startDate, {
-            addSuffix: true,
+        // check if difference between start and end dates is more than one month
+        if (!isSameMonth(startDate, endDate)) {
```
valiafetisov, line 32: That's not what you're checking here, [`isSameMonth` checks if two dates are in the same calendar month](https://date-fns.org/v2.29.3/docs/isSameMonth)

Instead, you can use [differenceInDays](https://date-fns.org/v2.29.3/docs/differenceInDays) or [differenceInMonths](https://date-fns.org/v2.29.3/docs/differenceInMonths)

valiafetisov (approved):

LukSteib (approved):

PR #561, opened 2022-12-07
patch: Improve autorouter

### LINKED ISSUE #558: Improve uni v3 auto router implementation
### Goal
Fix the known auto router related problems

### Context
During the implementation of the auto router (see #531) we faced the following problems that have been ignored for now:
* We're regenerating pools in a different moment than when we fetch price (see [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/531#discussion_r1039906540))
* We cannot assume that tokens used for the route are always available via the maker chainlog (see [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/531#discussion_r1039876930))

This issue is meant to fix these two problems and improve the current implementation

### Tasks
- [ ] Fix pool regeneration problem
- [ ] Fix unknown token problem

KirillDogadin-std:
Closes #558

Checklist:


core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
```diff
@@ -91,6 +91,11 @@ const _fetchAutoRouteInformation = async function (
             throw new Error('Only V3 routes are supported.');
         }
         const fees = bestRoute.route.pools.map(pool => pool.fee);
+        const pools = bestRoute.route.pools.map(pool => ({
+            addresses: [pool.token1.address, pool.token0.address],
+            fee: pool.fee,
+            routes: [pool.token1.symbol, pool.token0.symbol],
+        }));
```
KirillDogadin-std, line 98: this intends to address https://github.com/sidestream-tech/unified-auctions-ui/pull/531#discussion_r1039876930

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -39,20 +37,18 @@ const getMarketPrice = async function (
...
-    collateralAmount: BigNumber
+    collateralAmount: BigNumber,
+    preloadedPools?: Pool[]
 ): Promise<BigNumber> {
-    // convert collateral into DAI
-    const { route, fees } = await getRouteAndGasQuote(network, collateral.symbol, collateralAmount, marketId);
-    if (!route) {
-        throw new Error(`No route found for ${collateral.symbol} to DAI`);
+    if (!preloadedPools) {
+        throw new Error(`pools required to get market price for callee type "${marketId}"`);
```
KirillDogadin-std, line 44: this intends to address https://github.com/sidestream-tech/unified-auctions-ui/pull/531#discussion_r1039906540

core/src/calleeFunctions/helpers/uniswapAutoRouter.ts
valiafetisov: Thanks!

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -40,23 +39,28 @@ const getMarketPrice = async function (
...
+            network,
+            collateral.symbol,
+            marketId,
+            collateralAmount,
+            generatedPools
+        );
+        return { price: daiAmount.dividedBy(collateralAmount), pools: generatedPools };
+    }
+    if (totalPrice && pools) {
+        return { price: totalPrice.dividedBy(collateralAmount), pools: pools };
```
valiafetisov, line 61: The only remaining problem I see now is that `totalPrice` here is not the adjusted price that takes exchange fees into account (so we can properly compare exchange prices between each other), but the normal one

core/src/calleeFunctions/UniswapV3Callee.ts
```diff
@@ -40,23 +39,28 @@ const getMarketPrice = async function (
...
-): Promise<BigNumber> {
-    // convert collateral into DAI
-    const { route, fees } = await getRouteAndGasQuote(network, collateral.symbol, collateralAmount, marketId);
-    if (!route) {
-        throw new Error(`No route found for ${collateral.symbol} to DAI`);
-    }
-    const pools = await routeToPool(network, route, collateral.symbol, fees);
-    const daiAmount = await convertCollateralToDaiUsingPool(
+): Promise<{ price: BigNumber; pools: Pool[] }> {
+    const { route, fees, totalPrice, pools } = await getRouteAndGasQuote(
```
valiafetisov, line 43: Nit: I would remove `getRouteAndGasQuote` whatsoever and move its logic here, inside `getMarketPrice`. Because it's just adds a very strange layer of abstraction which doesn't add any value, but complicates understanding of what's going on

core/src/calleeFunctions/index.ts
```diff
@@ -64,63 +62,58 @@ export const getPools = async (
         return await routeToPool(network, calleeConfig.route, collateral.symbol);
     }
     if ('automaticRouter' in calleeConfig) {
-        const { route, fees } = await fetchAutoRouteInformation(network, collateral.symbol, amount.toFixed());
-        if (!route) {
-            throw new Error('No automatic route can be found');
+        const { pools } = await fetchAutoRouteInformation(network, collateral.symbol, amount.toFixed());
```
valiafetisov, line 65: - Where do we still use `getPools` if we the proposal is to never use them separately from the price?
- Why do we ever send amount as a string?

core/src/types.ts
```diff
@@ -116,7 +116,7 @@ declare interface MarketDataBase extends Partial<ExchangeFees> {
 export declare interface Pool {
     addresses: string[];
     fee: number;
-    routes: string[];
+    routes: (string | undefined)[];
```
valiafetisov, line 119: In which case it can be an array of `undefined`? 😓

core/src/calleeFunctions/index.ts
KirillDogadin-std: > Where do we still use getPools if we the proposal is to never use them separately from the price?

`getCalleeData` for UniswapV3 has 2 use cases:

  - autorouter
    - the prices and pools are extracted at the same time. No need to perform other actions (`getPools` is not called)
  - hardcoded
    - the route is provided, prices are not extracted with pools (because alpharouter is not called) and we have to do the old fashioned extraction from the hardcoded route. But we also have to unify the format to use pools primarily. So the `getPools` is called for the sake of converting route to pool. 

I will refactor the code a bit so that `getPools` cannot be called for callee configs where autorouter is enabled.

core/src/types.ts
KirillDogadin-std: uniswap autorouter does not promise to return the symbol of the token. It's optional there.

core/src/calleeFunctions/index.ts
KirillDogadin-std: removed unused arg (amount)

core/src/calleeFunctions/UniswapV3Callee.ts
KirillDogadin-std: i observe that the autorouter returns 2 amounts: 

```ts
    /**
     * The quote for the swap.
     * For EXACT_IN swaps this will be an amount of token out.
     * For EXACT_OUT this will be an amount of token in.
     */
    quote: CurrencyAmount;
    /**
     * The quote adjusted for the estimated gas used by the swap.
     * This is computed by estimating the amount of gas used by the swap, converting
     * this estimate to be in terms of the quote token, and subtracting that from the quote.
     * i.e. quoteGasAdjusted = quote - estimatedGasUsedQuoteToken
     */
    quoteGasAdjusted: CurrencyAmount;
```

My understanding of the above does not combine well with your claim.

The documentation claims that quote is the amount of "token out". If the value does not account for exchange fees, then the amount out is different from `quote` and hence the documentation is not valid.

Did you perhaps mean the gas fees? should i use `quoteGasAdjusted` value instead of `quote`?

LukSteib (approved):
Hard for me to review properly 
* don't have a test case where the auto router uses a route with an unknown token 
* cannot properly test regeneration of pools

However, 
* was able to participate in `goerli` auction via auto router
* UI is not broken but experience is quite similar to current state of prod

Hence approving this PR

LukSteib (changes requested):
- [ ] Live testing this as we currently have auctions on mainnet. Enabling the auto-router on the `RENBTC-A` auctions results in UI crash (see below)


console error: 
```
auctions.ts?b334:336 Uncaught TypeError: Cannot read properties of undefined (reading 'forEach')
    at Store.updateAuctionsPrices (auctions.ts?b334:336:1)
    at Array.wrappedActionHandler (vuex.esm.js?2f62:851:1)
    at Store.dispatch (vuex.esm.js?2f62:516:1)
    at Store.boundDispatch [as dispatch] (vuex.esm.js?2f62:406:1)
    at local.dispatch (vuex.esm.js?2f62:779:1)
    at eval (auctions.ts?b334:238:1)
```

Screenshot UI: 
image

KirillDogadin-std:
fixed the error, the auction seems to be already gone from the mainnet. I used fork on hardhat instead.

valiafetisov:
> CI tells that it's not mergable. Idk how so. Any thoughts?

<details><summary>Executing commands on the fresh clone worked for me</summary>

```
user@user ~ % cd /tmp      
user@user /tmp % git clone https://github.com/sidestream-tech/unified-auctions-ui
Cloning into 'unified-auctions-ui'...
remote: Enumerating objects: 8097, done.
remote: Counting objects: 100% (495/495), done.
remote: Compressing objects: 100% (291/291), done.
remote: Total 8097 (delta 313), reused 320 (delta 201), pack-reused 7602
Receiving objects: 100% (8097/8097), 9.99 MiB | 11.54 MiB/s, done.
Resolving deltas: 100% (5913/5913), done.
user@user /tmp % cd unified-auctions-ui             
user@user unified-auctions-ui % git fetch origin refs/pull/561/head
From https://github.com/sidestream-tech/unified-auctions-ui
 * branch            refs/pull/561/head -> FETCH_HEAD
user@user unified-auctions-ui % git merge 4521483a2ba76b1d00d0414ba3b64f6df9cfbc4d
Updating 89db7a2..4521483
Fast-forward
 .../src/calleeFunctions/CurveLpTokenUniv3Callee.ts | 16 ++++--
 core/src/calleeFunctions/UniswapV2CalleeDai.ts     | 14 +++--
 .../calleeFunctions/UniswapV2LpTokenCalleeDai.ts   |  4 +-
 core/src/calleeFunctions/UniswapV3Callee.ts        | 45 ++++++++++------
 core/src/calleeFunctions/WstETHCurveUniv3Callee.ts | 10 ++--
 .../calleeFunctions/helpers/uniswapAutoRouter.ts   | 14 ++++-
 core/src/calleeFunctions/helpers/uniswapV3.ts      | 27 +---------
 core/src/calleeFunctions/index.ts                  | 61 +++++++++-------------
 core/src/calleeFunctions/rETHCurveUniv3Callee.ts   | 16 ++++--
 core/src/types.ts                                  |  7 ++-
 10 files changed, 117 insertions(+), 97 deletions(-)
```

</details>

So the state of the branch doesn't seem to be a problem. @DeusAvalon?

LukSteib:
As discussed orally, will document the other error experienced during live test yesterday:

* network: `mainnet`
* related auction: https://etherscan.io/tx/0xbbe894d1fcaa1e1e917cb121b7222f0aeaeacbc28f2101afb2402e1c50db18a7
  * taken at block height: 16176353
* Error experienced when trying to execute swap tx with unsiwap v3 callee being selected
* Error message: `Transaction error: "execution reverted: Too little received (cannot estimate gas; transaction may fail or may require manual gas limit)"``

KirillDogadin-std:
could you reproduce this on fork? 

e.g. on block 16176345

i fail to see the same error.

LukSteib:
> could you reproduce this on fork?

As discussed verbally, currently there is no chance for me to start simulation with fork at specific block height. Hence created #568 as intermediate issue.

KirillDogadin-std:
> As discussed verbally, currently there is no chance for me to start simulation with fork at specific block height. Hence created https://github.com/sidestream-tech/unified-auctions-ui/issues/568 as intermediate issue.

did #232 ever get completely resolved? might it be the case that the calculation is just still screwed and in some cases the profit is just really not sufficient, but calculated incorrectly?

valiafetisov:
> might it be the case that the calculation is just still screwed and in some cases the profit is just really not sufficient, but calculated incorrectly?

Yes, might be the case that profit amount is calculated incorrectly. But this is always about the precise amount and not about it being negative – we did not have internal or external complains about failed transactions (except for one issue which was long time fixed)

KirillDogadin-std:
> Yes, might be the case that profit amount is calculated incorrectly. But this is always about the precise amount and not about it being negative – we did not have internal or external complains about failed transactions (except for one issue which was long time fixed)

i have just tested this on `main` branch. Same result. Therefore i would say that the core issue here is not related to this pr and should be resolved in the different issue's scope.

valiafetisov (approved):

KirillDogadin-std:
> Via simulation: For other collateral types like RENBTC-A or LINK-A enabling the auto router via UI resulted in an endless (killed process after 5min) loading state without getting actual prices (no console error thrown).

this is due to alpha router not being able to load in the data and failing.

currently the ui does not notify that this happened:
  - we can scope this in the other issue and develop some smart way of showing this
  - i can hack-patch it here and disable loading if the loading failed

LukSteib:
> we can scope this in the other issue and develop some smart way of showing this

Fine for me to scope it out. So for me this results in two follow up issues: 
* investigate problem with profit calculation
* better error handling for alpha router

LukSteib (approved):

PR #562, opened 2022-12-07
Handle multiple `window.ethereum` providers

### LINKED ISSUE #551: Fix Metamask wallet support while Coinbase extension installed
### Goal

MetaMask extension works even if Coinbase is also installed

### Context

It was discovered that as soon as Coinbase extension is installed, it also modifies global `window.ethereum` object which makes it incompatible with Metamask. More specifically, [Metamask docs specify `window.ethereum.isConnected` function](https://docs.metamask.io/guide/ethereum-provider.html#methods) which no longer exist after `Coinbase` overwrites the object. I don't know if it's intentional or not (at least sounds like a war between extensions), but we should figure out how those two can coexists or, if they can't, indicate to the user how to solve the problem.

### Assets

- Potential solution https://www.npmjs.com/package/@metamask/detect-provider
- Related answer https://stackoverflow.com/a/70796779

### Tasks

- [x] Investigate what can be used instead of `isConnected` or if we should drop support for "reconnection" after page restart
- [x] Investigate how Metamask can still be accessed even if Coinbase is installed
  - [x] Otherwise, propose how the conflict can be discovered
- [x] Discuss
- [x] Implement

aomafarag:
Closes #551.

Checklist:

### DIFF
--- a/frontend/lib/wallets/MetaMask.ts
+++ b/frontend/lib/wallets/MetaMask.ts
@@ -2,6 +2,7 @@ import { ethers } from 'ethers';
 import { message } from 'ant-design-vue';
 import { getChainIdByNetworkType, getNetworkTypeByChainId } from 'auctions-core/src/network';
 import { setSigner } from 'auctions-core/src/signer';
+import { formatToHexWithoutPad } from 'auctions-core/helpers/format';
 import MetaMaskLogo from '~/assets/icons/wallets/metamask.svg';
 import AbstractWallet from '~/lib/wallets/AbstractWallet';
 
@@ -10,7 +11,8 @@ export default class MetaMask extends AbstractWallet {
     public static icon = MetaMaskLogo;
     public static downloadUrl = 'https://metamask.io/';
 
-    private signer?: ethers.providers.JsonRpcSigner;
+    private static provider?: any;
+
     private addresses?: string[];
 
     public static get isInterfaceReady() {
@@ -21,7 +23,13 @@ export default class MetaMask extends AbstractWallet {
         if (!window.ethereum) {
             return false;
         }
-        if (typeof window.ethereum?.isConnected !== 'function') {
+        if (window.ethereum.providers?.length) {
+            const metaMask = window.ethereum.providers.find((p: any) => p.isMetaMask);
+            if (metaMask) {
+                return metaMask.isConnected();
+            }
+        }
+        if (typeof window.ethereum.isConnected !== 'function') {
             return false;
         }
         return window.ethereum.isConnected();
@@ -30,6 +38,12 @@ export default class MetaMask extends AbstractWallet {
     public static get isLoggedIn() {
         if (!window.ethereum) {
             return false;
+        }
+        if (window.ethereum.providers?.length) {
+            const metaMask = window.ethereum.providers.find((p: any) => p.isMetaMask);
+            if (metaMask) {
+                return metaMask.selectedAddress;
+            }
         }
         return window.ethereum.selectedAddress;
     }
@@ -41,7 +55,20 @@ export default class MetaMask extends AbstractWallet {
     }
 
     getProvider(): ethers.providers.JsonRpcProvider {
-        return new ethers.providers.Web3Provider(window.ethereum, 'any');
+        if (!MetaMask.provider) {
+            MetaMask.provider = window.ethereum;
+            // handle case of MetaMask and Coinbase Wallet extensions both installed
+            if (typeof window.ethereum !== 'undefined') {
+                if (window.ethereum.providers?.length) {
+                    window.ethereum.providers.forEach((p: any) => {
+                        if (p.isMetaMask) {
+                            MetaMask.provider = p;
+                        }
+                    });
+                }
+            }
+        }
+        return new ethers.providers.Web3Provider(MetaMask.provider, 'any');
     }
 
     async getSigner(): Promise<ethers.providers.JsonRpcSigner> {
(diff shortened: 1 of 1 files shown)


LukSteib (approved):
* checked metamask only setup -> was able to connect wallet ✔️
* checked metamask + coinbase setup -> was able to connect with both wallets and execetue txs (on goerli) ✔️

frontend/lib/wallets/MetaMask.ts
```diff
@@ -41,7 +55,20 @@ export default class MetaMask extends AbstractWallet {
...
+        if (!MetaMask.provider) {
+            MetaMask.provider = window.ethereum;
+            // handle case of MetaMask and Coinbase Wallet extensions both installed
+            if (typeof window.ethereum !== 'undefined') {
+                if (window.ethereum.providers?.length) {
+                    window.ethereum.providers.forEach((p: any) => {
+                        if (p.isMetaMask) {
+                            MetaMask.provider = p;
+                        }
+                    });
```
KirillDogadin-std, line 67: if there's supposed to be just 1 metamask object in this array, you can stop the loop after the `if` condition is hit.

frontend/lib/wallets/MetaMask.ts
```diff
@@ -31,6 +39,12 @@ export default class MetaMask extends AbstractWallet {
         if (!window.ethereum) {
             return false;
         }
+        if (window.ethereum.providers?.length) {
+            const metaMask = window.ethereum.providers.find((p: any) => p.isMetaMask);
```
KirillDogadin-std, line 43: nit: i'm a strong proponent of complete names. Hence i suggest to name it `provider` instead of `p`

KirillDogadin-std (dismissed):

frontend/lib/wallets/MetaMask.ts
```diff
@@ -10,7 +11,8 @@ export default class MetaMask extends AbstractWallet {
     public static icon = MetaMaskLogo;
     public static downloadUrl = 'https://metamask.io/';
 
-    private signer?: ethers.providers.JsonRpcSigner;
+    private static provider?: any;
```
valiafetisov, line 14: First of all, why do you now need to have static variable that holds provider? Can this instead be a static [getter] function?

Can you please avoid using `any` as much as possible? Types are helpful when you use them to reinforce your code. What do you plan to store here?

frontend/lib/wallets/MetaMask.ts
```diff
@@ -21,7 +23,13 @@ export default class MetaMask extends AbstractWallet {
...
             return false;
         }
-        if (typeof window.ethereum?.isConnected !== 'function') {
+        if (window.ethereum.providers?.length) {
+            const metaMask = window.ethereum.providers.find((provider: any) => provider.isMetaMask);
+            if (metaMask) {
+                return metaMask.isConnected();
+            }
+        }
+        if (typeof window.ethereum.isConnected !== 'function') {
```
valiafetisov, line 32: `isConnected` function purpose is to check whether the website was previously connected to the metamask, so on the page refresh we can reconnect to it. It prevents users who previously haven't connected their metamask seeing a new dialog as soon a they open the page.

Currently, you're using `window.ethereum.isConnected` which wouldn't be available in case coinbase extension is installed. So my recommendation would be to get (static) provider here first and then check if this function is there.

frontend/lib/wallets/MetaMask.ts
```diff
@@ -41,7 +55,19 @@ export default class MetaMask extends AbstractWallet {
     }
 
     getProvider(): ethers.providers.JsonRpcProvider {
-        return new ethers.providers.Web3Provider(window.ethereum, 'any');
+        if (!MetaMask.provider) {
+            MetaMask.provider = window.ethereum;
```
valiafetisov, line 59: this would already assign a provider, which might not be a metamask provider (since it can be coinbase, as we learned). Can this be done more safe? Ie, is there a way to check if it's a metamask by the metamask API itself?

valiafetisov:
> I propose combining both `isConnected()` and `isMetaMask/isCoinbaseWallet` to set them apart.

Proposed pseudo-code:
```
// inside metamask
const getProvider = function () {
    if (window?.ethereum?.providers) {
        return window?.ethereum?.providers.find(provider => provider.isMetaMask);
    }
    if (window?.ethereum?.isConnected) {
        return window?.ethereum;
    }
    throw new ...; // we should show "download error" here on the frontend
}
```
would this work?

frontend/lib/wallets/MetaMask.ts
KirillDogadin-std: according to

> First of all, why do you now need to have static variable that holds provider? 

and [this](https://github.com/sidestream-tech/unified-auctions-ui/pull/562#discussion_r1045661166) and [this](https://github.com/sidestream-tech/unified-auctions-ui/pull/562#issuecomment-1347918447)

i've merged the static variable and function into static getter function.

LukSteib (changes requested):
* Test scenario 1 - only metamask extension: 
   * Error `Please install Metamask first from https://metamask.io/` thrown, not able to connect 🔴 
* Test scenario 2 - metamask and coinbase extension
   * Metamask:   Error `Please install Metamask first from https://metamask.io/` thrown, not able to connect 🔴 
   * Coinbase: Works

Please ensure that all of the cases work.

KirillDogadin-std:
i fail to reproduce this, please throw info in. for me both work. Did you by chance leave metamask disabled in the browser? If this is not the case - can you provide a video?

LukSteib:
>  fail to reproduce this, please throw info in. for me both of the cases work.

My bad. Metamask was stuck in the loading state and this is an expected scenario to throw the reported error above. 
Now both test scenarios looking good ✅ 

Only thing I noticed (probably related to `isConnected`):
When Connect Metamask -> Disconnect Metamask -> Connect Coinbase -> Disconnect Coinbase -> Site refreshes automatically and automatically connects back to Metamask (see recording below). I'd also be ok with opening a new issue for that and put it out of scope if this is too envolved. Leave it up to you:

https://www.loom.com/share/5db4debcec1f4bcfb0a9c63c5f4596f9

KirillDogadin-std:
I would have to perform additional investigations and look into it depeer anyways (read: "i dont have an instant idea how to fix it"). I would give it back to @valiafetisov for code re-review to pinpoint the already present problems with the code and if he suggests a quick-fix (read: "has an instant idea") - i'll add it as well.

valiafetisov (dismissed):

PR #563, opened 2022-12-07
Fix initial fetching/loading state

### LINKED ISSUE #550: Fix "not found" messages before loading
### Goal

Not found messages in the UI no longer displayed on load.

### Context

Currently, 
- When a page to an auction is opened directly via url, the UI first show "This auction was not found" error and only later (a few seconds later) change it into the loading state
- When the main page is opened with the table or auctions, we also first display "No active auctions" and only a few moments later show loading state

Instead, the pages should always show loading state first or somehow indicate proper state of the auction
Please note that we once had a similar problem that was previously fixed: https://github.com/sidestream-tech/unified-auctions-ui/pull/90#issuecomment-1063130777

### Assets
- Example url https://unified-auctions.makerdao.com/collateral?network=mainnet&auction=WBTC-A%3A26
- Example screenshot
  Screenshot 2022-11-28 at 13 31 22

### Tasks

- [x] Come up with several approaches to fix it, choose the best (we also don't want to show loading when it's not actually loading yet)
- [x] Fix auction page error
- [x] Fix auction table error

aomafarag:
Closes #550.

Checklist:

### DIFF
--- a/frontend/store/auctions.ts
+++ b/frontend/store/auctions.ts
@@ -54,11 +54,9 @@ const getInitialState = (): State => ({
 export const state = (): State => getInitialState();
 
 export const getters = {
-    listAuctions(state: State): AuctionTransaction[] {
-        return Object.values(state.auctionStorage);
-    },
-    listAuctionTransactions(state: State, getters: any, _rootState: any): AuctionTransaction[] {
-        const auctions = Object.values(state.auctionStorage);
+    listAuctionTransactions(state: State, getters: any, _rootState: any, rootGetters: any): AuctionTransaction[] {
+        const network = rootGetters['network/getMakerNetwork'];
+        const auctions = Object.values(state.auctionStorage).filter(auction => auction.network === network);
         return auctions.map(auction => {
             const isRestarting = getters.isAuctionRestarting(auction.id);
             return {


LukSteib (dismissed):
Checked with the following deep links to an auction - good experience without the auction not found indicator:

* goerli: http://localhost:3000/collateral?network=goerli&auction=WSTETH-A%3A1
* mainnet: http://localhost:3000/collateral?network=mainnet&auction=USDC-A%3A185

frontend/components/auction/collateral/CollateralAuctionFlow.vue
```diff
@@ -28,7 +28,7 @@
                         :take-events="selectedTakeEvents"
                         :wallet-address="walletAddress"
                         :auction-id="selectedAuctionId"
-                        :are-auctions-fetching="areAuctionsFetching || isSelectedAuctionFetching"
+                        :are-auctions-fetching="areAuctionsFetching"
```
valiafetisov, line 31: Why was this removed? When all auctions are not fetched (but only a single auction is) `areAuctionsFetching` will not be set to my knowledge

valiafetisov (commented):
Can the vault fetching, surplus and debt auctions be also fixed in the same PR, since I assume it has the same problem

frontend/components/auction/collateral/CollateralAuctionFlow.vue
valiafetisov: But that was intended to have separate loading indicators. The reason is to not mislead user by showing loading on the table with auctions when in actuality we only fetch a single auction

valiafetisov:
Simpler solution would be to set both `areAuctionsFetching` and `isSingleAuctionFetching` to `false` whenever any of them are finished. This should not affect [the problem I described in the comment](https://github.com/sidestream-tech/unified-auctions-ui/pull/563#discussion_r1043319904), right?

valiafetisov:
> we wrongly display `No active auctions`

This is unrelated to the scope of the issue, right?

In sort: the sentence can be better, but the behaviour is intended. We don't want to fetch all auctions when user opens a link to a single auction. The reason is in case of the market crash:
 - most of the users expected to come from twitter and will already have URL
 - we don't want them to load all 200 auctions when their UI becomes unresponsive

So proper solution to this problem (that you can fix/discuss in another PR) would be a better wording, eg `No auctions were fetched yet`. Or, in order to display finished auction in the table, the finished auction fetching should be refactored like we're doing it with vaults: in case auction is active it's values are fetched and in case it's finished, the events are fetched in one call, in one "combined" type

valiafetisov:
> Currently, when a single auction is opened directly via URL, the auctions table displays the loading animation for a few seconds until updateSingleAuction sets areAuctionsFething to false. This only happens initially, not each time the single auction is updated. Is this OK?

I guess it's fine in the scope of this issue. 

> Should I change that?

Good catch! You certainly can, but in another PR 🙂

valiafetisov (approved):

LukSteib (approved):

PR #565, opened 2022-12-07
Filter auctions based on network

### LINKED ISSUE #553: Fix problem wrongly displayed auctions
### Goal
Only auctions of the selected network are displayed

### Context
We figured that switching network (in example case from `goerli` to `mainnet`) results in the situation that `goerli` auctions keep being displayed although the network both on UI and in wallet have been switched.
The problem can be replicated on staging as well as production 

We need to ensure that this is not the case as it can be very confusing for a user seeing auctions that shouldn't be there. 

### Tasks
- [ ] Investigate the problem of auctions being displayed for the wrong network
- [ ] Fix the problem 

### Assets
Problem replication:
https://www.loom.com/share/e4f33d1311c94427a603496ef0725481

aomafarag:
Closes #553.

Checklist:

### DIFF
--- a/frontend/store/auctions.ts
+++ b/frontend/store/auctions.ts
@@ -54,11 +54,11 @@ const getInitialState = (): State => ({
 export const state = (): State => getInitialState();
 
 export const getters = {
-    listAuctions(state: State): AuctionTransaction[] {
-        return Object.values(state.auctionStorage);
-    },
-    listAuctionTransactions(state: State, getters: any, _rootState: any): AuctionTransaction[] {
-        const auctions = Object.values(state.auctionStorage);
+    // `rootGetters` is exposed to the `getters` module as the 4th parameter, right after `rootState`
+    // Hence the dummy _rootState parameter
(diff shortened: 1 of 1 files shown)


LukSteib (approved):
Could not reproduce the problem anymore while switching back and forth between goerli and mainnet ✅

frontend/store/auctions.ts
```diff
@@ -54,11 +54,11 @@ const getInitialState = (): State => ({
 export const state = (): State => getInitialState();
 
 export const getters = {
-    listAuctions(state: State): AuctionTransaction[] {
-        return Object.values(state.auctionStorage);
-    },
-    listAuctionTransactions(state: State, getters: any, _rootState: any): AuctionTransaction[] {
-        const auctions = Object.values(state.auctionStorage);
+    // `rootGetters` is exposed to the `getters` module as the 4th parameter, right after `rootState`
+    // Hence the dummy _rootState parameter
```
valiafetisov, line 58: If you want to leave comment to the reviewer, you can make a comment in the PR itself, not in the code 🙂 
```suggestion
```

valiafetisov (approved):

PR #569, opened 2022-12-14
feat: Specify block on vault liquidation simulation

### LINKED ISSUE #568: Adjust `Fork block with active WSTETH-A auction` simulation
### Goal
Have a simulation that allows to specify block height via input and then forks mainnet at the specified block number

### Context
We oftentimes run into the situation that we experience some behavior on mainnet and then want to reproduce it by running a fork at a specific block height.
Currently none of our simulations allows for this scenario which we want to address in this issue.
Since the Fork block with active WSTETH-A auction does it with a hardcoded block height it seems reasonable to adjust this simulation in the way described above

### Tasks
- [ ] Ensure to have simulation that allows user to specify block height and then forks it
- [ ]  If we adjust Fork block with active WSTETH-A auction for that, please rename the simulation to Fork at specific block

KirillDogadin-std:
Closes #568

Checklist:


LukSteib (changes requested):
In this PR you have changed the existing "Create collateral auction" in a way that it allows block height input. However, I would not touch this simulation as this allows to test vault creation and liquidation on the latest block and should be kept like that. 
Instead my suggestion as in the issue was to adjust the WSTETH simulation in way to make it more generic. This one will then allow to "replay" certain moments of time in the past by specifying the block height and then just forking the chain. No need to have the vault creation and liquidation here as we specify blocks that are of interest to us and where there have been active auctions.

So my suggestion is to ajdust this PR in accordance to issue

KirillDogadin-std:
I would vouch to leave the liquidation simulation with the number choice as well because somethimes the latest block falls into a "funny" scenario where the protocol's contract cannot perform an action because one of the conditions are not met. So rolling back several blocks usually allows to use the simulation again. But currently code change is necessary.

LukSteib:
> I would vouch to leave the liquidation simulation with the number choice as well

Agree as long as the latest block is always pre-selected as default. Otherwise it comes with overhead to first lookup the current block number in order to stay on top of things.

KirillDogadin-std:
Yes, it is already the case

LukSteib (approved):

core/simulations/helpers/promptToGetNumber.ts
```diff
@@ -0,0 +1,16 @@
+import prompts from 'prompts';
+
+const promptToGetNumber = async (title: string, initial: number) => {
+    const { number } = await prompts([
+        {
+            type: 'number',
+            name: 'number',
+            message: title,
+            initial,
+            min: 1,
```
valiafetisov: If initial is set (which I assume is selected via just pressing enter), is there a way to set the latest block (without checking it online and typing in?). Eg, by typing `current` or `-1` (https://docs.ethers.org/v5/api/providers/types/#providers-BlockTag)?

KirillDogadin-std: i don't get the question here.

the latest block is already pre-computed and sent as `initial` arg to this funciton. If the latest block is "100500", then there will be a graytext saying "100500" and one will not have to do anything.

valiafetisov: This is in case no specific block is suggested, right? But how to enter "latest block" in case of the suggested `14052140` (above in your code)?

KirillDogadin-std: did not find a simple & straightforward solution. Adjusted the code so that it suggests the latest block instead of the specific block with an active auction (since we have the simulation now that creates auctions, this was redundant anyway)

LukSteib (dismissed):

core/simulations/configs/vaultLiquidation.ts
```diff
@@ -19,7 +22,9 @@ const simulation: Simulation = {
         {
             title: 'Reset blockchain fork',
             entry: async () => {
-                await resetNetworkAndSetupWallet();
+                const latestBlock = await hre.ethers.provider.getBlockNumber();
```
valiafetisov, line 25: Can you please move it to the `promptToGetBlockNumber` 😉 ?
So that you only fetch it if `max` is undefined. I would also: 
- rename `promptToGetNumber` to `promptToGetBlockNumber`
- move `title` to there and make it a default
- restructure arguments to be a single `options` objects, since most of them are optional

core/simulations/helpers/promptToGetNumber.ts
valiafetisov: Nit: I guess it's not a requirement now, but as suggested you can accept `-1` input as the latest block (also min can be 0 actually)

KirillDogadin-std: - adjusted the min to be 0
- i don't think allowing `-1` as latest block is any better than suggesting one by default: at least if the implementation that is currently present is not flawed / bad, there is no benefit (afai can see) in ux/ui.

core/simulations/helpers/promptToGetNumber.ts
```diff
@@ -0,0 +1,31 @@
...
+
+import hre from 'hardhat';
+
+interface BlockNumberPromt {
+    title?: string;
+    initial?: number;
+    max?: number;
+    min?: number;
+}
+const promptToGetNumber = async (params?: BlockNumberPromt) => {
```
valiafetisov, line 11:
```suggestion
const promptToGetBlockNumber = async (params?: BlockNumberPromt) => {
```

core/simulations/helpers/promptToGetNumber.ts
```diff
@@ -0,0 +1,31 @@
...
+    const title: string = params?.title || 'Block number to fork from';
+    const min = params?.min || 0;
+    const initial = params?.initial;
+    const max = params?.initial;
+    // only compute latest block if we need to fill values
+    const latestBlock = max && initial ? undefined : await hre.ethers.provider.getBlockNumber();
```
valiafetisov, line 17: Providing `initial` shouldn't automatically set max to it as well (currently, `params.max` is unused)

```suggestion
    const max = params?.max ?? await hre.ethers.provider.getBlockNumber();
```

core/simulations/helpers/promptToGetNumber.ts
```diff
@@ -0,0 +1,31 @@
...
+    const { number } = await prompts([
+        {
+            type: 'number',
+            name: 'number',
+            message: title,
+            initial: initial || latestBlock,
```
valiafetisov, line 23: Nit: please prefer `??` over `||` when possible. Currently, if `initial` is set to 0 it wouldn't be processed correctly

Also
```suggestion
            initial: initial || max,
```

core/simulations/helpers/promptToGetBlockNumber.ts
```diff
@@ -0,0 +1,29 @@
...
+    title?: string;
+    initial?: number;
+    max?: number;
+    min?: number;
+}
+const promptToGetBlockNumber = async (params?: BlockNumberPromt) => {
+    const title: string = params?.title || 'Block number to fork from';
+    const min = params?.min || 0;
+    const initial = params?.initial;
+    const max = params?.initial ?? (await hre.ethers.provider.getBlockNumber());
```
valiafetisov, line 15:
```suggestion
    const max = params?.max ?? (await hre.ethers.provider.getBlockNumber());
```

valiafetisov (approved):

PR #573, opened 2023-01-02
Provide RPC URL without env variables

### LINKED ISSUE #571: Do not require frontend required env variables
### Goal
Frontend does not require env variables

### Context
This issue forms part of our roadmap to package the Unified Auctions UI into a standalone app. 
A first step to achieve that is to not require env variables to get the frontend started any longer.

### Tasks
- [ ] Refactor the frontend / core to accept `RPC_URL` async and to not require it on start
   - [ ] no error is shown
   - [ ] as soon as it's needed, everything is cleared (and refetched, if needed)
- [ ] Create UI dialog that is automatically shown if no `RPC_URL` is provided

aomafarag:
Closes #571.

Screenshot from 2023-01-02 13-50-08

Checklist:


LukSteib (commented):
* Tested with an infura RPC URL and was able to switch between `goerli` and `main` ✔️ 
* Inserting non RPC URL throws an error ✔️ 
* Tested with an alchemy mainnet RPC URL. Was not able to switch to test network. Probably expected. The network indicator in the UI is displaying `Main-like` (see screenshot below). Is that expected?


image

LukSteib:
> This is expected; it has always been the case.

Agree, fine

> The actual reason is a mistake on my part: I only setup the network when the RPC URL is provided. With the newest commit, I setup the wallet, auctions, etc. as well, so this problem should be resolved now.

Not getting this error anymore. Able to fetch auctions and do tx. 

> However, currently, I'm getting this error in the console:
Failed to load resource: the server responded with a status of 429 ()

My console is also flooded with the following error (most probably the same):

`POST {RPC-URL} 429`

Although this might not be due to us, can you please do an investigation where this error emerges (e.g. checking the alchemy docs) to have more insights on it.

LukSteib (dismissed):

valiafetisov:
- [ ] Looking at the screenshot, let's change how the modal looks
   - [ ] Different title
   - [ ] Text section with external (target=_blank) links
   - [ ] Loading state that checks if url is valid (eg you can send a basic post request to the the url to get chain id, [eg](https://getblock.io/docs/available-nodes-methods/ETH/JSON-RPC/eth_chainId/))
   - [ ] Error message
   - [ ] Previously set url can be found in the same input element
   - [ ] Reconnect button
- [ ] Please add more stories for the modal

MakerDAO new collateral@2x (11)

core/src/rpc.ts
```diff
@@ -28,7 +29,10 @@ export const setupRpcUrlAndGetNetworks = async function (
     isDev = false
 ): Promise<{ networks: NetworkConfig[]; defaultNetwork: string; defaultChainId: string }> {
     if (!rpcUrl) {
-        throw new Error(`No RPC_URL env variable was provided`);
+        throw new Error(`Invalid RPC URL`);
+    }
+    if (getNetworks()) {
+        resetNetworks();
     }
```
valiafetisov, line 36: What does this try to accomplish? Why do you need to reset networks in the first place? Why is it related to the execution of the `getNetworks`?

frontend/components/layout/Header.vue
```diff
@@ -19,6 +19,8 @@
                     </label>
 
                     <div class="flex space-x-4">
+                        <button class="inline-block" @click="$emit('openRpcUrlInputModal')">Change RPC URL</button>
```
valiafetisov, line 22: Let's put it into the networks dropdown? Since this is also added to the regular "production version" of the website it should work well on mobile as well

frontend/components/modals/RpcUrlInputModal.vue
```diff
@@ -0,0 +1,54 @@
...
+            newRpcUrl: '',
+        };
+    },
+    methods: {
+        submit() {
+            this.$emit('setRpcUrl', this.newRpcUrl);
+            this.close();
+        },
+        close() {
+            this.$emit('close');
```
valiafetisov, line 50: Nit: this should be better set via synced props (to avoid extra work of handling event in the parent)?

```suggestion
            this.$emit('update:isShown', false);
```

frontend/plugins/vuex-persist.client.js
```diff
@@ -1,11 +1,11 @@
...
 import Cookies from 'js-cookie';
 
-export default ({ store, isDev }) => {
-    window.onNuxtReady(() => {
+export default ({ store }) => {
+    window.onNuxtReady(async () => {
         new VuexPersistence({
             storage: window.localStorage,
-            modules: ['preferences'],
+            modules: ['network', 'preferences'],
```
valiafetisov, line 8: Putting the whole `network` store into the local storage would create a lot of side effects (think about `isChangingNetwork` being forever stack being `true`). Instead, please store it inside `preferences`

frontend/plugins/vuex-persist.client.js
```diff
@@ -14,7 +14,7 @@ export default ({ store, isDev }) => {
             },
             modules: ['cookies'],
         }).plugin(store);
+        await store.dispatch('network/setup');
```
valiafetisov, line 17: Why did you remove `isDev` parameter, which adds extra network to our dropdown (localhost:8545) depending on the environment

frontend/store/network.ts
```diff
@@ -114,20 +122,36 @@ export const actions = {
         await dispatch('surplus/setup', undefined, { root: true });
         await dispatch('debt/setup', undefined, { root: true });
     },
-    async setupNetworks({ commit }: ActionContext<State, State>, isDev?: boolean) {
+    async setupNetworks({ getters, commit, dispatch }: ActionContext<State, State>, oldRpcUrl?: string) {
```
valiafetisov, line 125: Why do you need to pass `oldRpcUrl` as an argument? Wouldn't it be stored in the store at this point?

valiafetisov:
> Failed to load resource: the server responded with a status of 429 ()
> I think this has to do with Alchemy, not our code. @valiafetisov, please confirm.

Why alchemy? Our secrets should contain valid infura RPC URL which doesn't have rate limits (since we're paying for it)

frontend/store/network.ts
```diff
@@ -114,20 +122,36 @@ export const actions = {
...
+        if (!getters.getRpcUrl) {
+            return;
+        }
         commit('setIsChangingNetwork', true);
         if (networkChangeTimeoutId) {
             clearTimeout(networkChangeTimeoutId);
         }
-        const { networks, defaultChainId, defaultNetwork } = await setupRpcUrlAndGetNetworks(
-            process.env.RPC_URL,
```
valiafetisov, line 123: We should still fallback to the `process.env.RPC_URL` (cause we still want to keep https://unified-auctions.makerdao.com running, where we provide our RPC url) this PR should not affect default production behaviour of the website

LukSteib:
I'll do another review once the adjusted modal and flows have been implemented as outlined [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/573#issuecomment-1380444933)

valiafetisov:
The first official link from google: https://ethereum.org/en/run-a-node/

core/src/rpc.ts
valiafetisov: Ok, but why do you need to check `if (getNetworks())` first before resetting it?
If you _are required to check it_, why it's not part of the `resetNetworks` function itself?

LukSteib (changes requested):
- [ ] When providing an invalid RPC URL in the first place the UI crashes and user is kind of stuck in a dead end up until actively re-loading the page. Let's try to better catch this problem (see s1 below) 
- [ ] NIT (if easily configurable): We might only ask for RPC URL when accessing a page that actually requires fetching blockchain information. In other words let's allow a user to access the `/` page without immediately requiring RPC URL. Only if navigating to e.g. `/collateral`. As indicated in beginning. Don't know when we start fetching things so let's not address if this complicates things heavily.
- [x] NIT: When providing a new RPC URL, which is valid, I'd suggest to show a toast notification like "Successfully changed RPC URL" and automatically close the modal



s1: 
image

LukSteib:
> Can you provide a screencast of how you arrived at this?

Check recording below. It's on the very first input of an RPC URL. This time the UI didn't break but still the user is somewhat stuck. 

https://www.loom.com/share/303249f297b34ebca3fbe1f28bd242e0

LukSteib (approved):
> Technically possible, but will require some adjustment and refactoring, since we dispatch all setup functions once Nuxt launches the app.

As said, fine to not address. Approved from my side

valiafetisov:
This PR currently breaks our production flow: when opened it looks like this:

Screenshot 2023-01-31 at 10 13 06

Instead: 
- [ ] We have to make sure that our production logic stays exactly the same as now. In other words: if `RPC_URL` env var is set and is not emtpy, it should be set up automatically, no modal should be shown
- [ ] RPC url stored in the local storage should be overwritten by the new `RPC_URL` env variable (if it's not empty) – so in case we change our provider or deprecate old infura url, users will not experience errors

In order to test this, you can clear your local storage in the browser settings

valiafetisov:
No, that means that initial value of that url should be preferably taken from env var over local storage. So even if localstorage  have something, but env var also have something, env var should be preferred

frontend/store/network.ts
```diff
@@ -114,8 +114,15 @@ export const actions = {
         await dispatch('surplus/setup', undefined, { root: true });
         await dispatch('debt/setup', undefined, { root: true });
     },
-    async setupNetworks({ rootGetters, commit }: ActionContext<State, State>) {
-        const rpcUrl = rootGetters['preferences/getRpcUrl'] || process.env.RPC_URL;
+    async setupNetworks({ rootGetters, commit, dispatch }: ActionContext<State, State>) {
+        let rpcUrl: string | undefined;
+        if (process.env.RPC_URL && !rootGetters['preferences/getWasEnvRpcUrlUsed']) {
```
valiafetisov, line 119: Why do you need to introduce and check `wasEnvRpcUrlUsed`?  If `RPC_URL` is provided, it should overwrite locally stored value in any case

valiafetisov: Also, please don't forget to tick [the requests](https://github.com/sidestream-tech/unified-auctions-ui/pull/573#issuecomment-1410027409) that were implemented

valiafetisov (approved):

PR #575, opened 2023-01-05
Configure router via env var

### LINKED ISSUE #574: Switch router behavior via env var
### Goal

Manipulate router's history mode via an environment variable.

### Context

Currently, our URLs are, for example, of this format `/collateral?auction=ETH-A:1`. However,  this is not supported when, say, a file is opened locally  (e.g. file:///users/User/index.html) . Without a server, we need an environment variable to tell `vue-router` to use `#/collateral?auction=ETH-A:1` instead.

### Assets

NuxtJS `router` property: https://nuxtjs.org/docs/configuration-glossary/configuration-router.

### Tasks

- [x] Add environment variable
- [x] Use it to configure router

aomafarag:
Closes #574.

Checklist:

### DIFF
--- a/frontend/nuxt.config.js
+++ b/frontend/nuxt.config.js
@@ -4,6 +4,10 @@ const PREVIEW_IMAGE = (process.env.FRONTEND_ORIGIN || '') + '/preview.jpeg';
 const SITE_TITLE = 'Unified Auctions';
 const SITE_DESCRIPTION = `The "${SITE_TITLE}" is the portal to all Maker Auctions related services. Easily interact with the Maker Protocol through streamlined interfaces, inform yourself about how the protocol works and receive updates on current auctions.`;
 const TWITTER_HANDLE = '@MakerDAO_SAS';
+// mode: https://v3.router.vuejs.org/api/#mode
+const ROUTER_HISTORY_MODES = ['history', 'hash', 'abstract'];
+// Nuxt has 'history' as default
+const HISTORY_MODE = ROUTER_HISTORY_MODES.includes(process.env.HISTORY_MODE) ? process.env.HISTORY_MODE : 'history';
 
 export default {
     // Disable server-side rendering: https://go.nuxtjs.dev/ssr-mode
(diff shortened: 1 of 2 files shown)


frontend/nuxt.config.js
```diff
@@ -1,9 +1,15 @@
...
 
 const PREVIEW_IMAGE = (process.env.FRONTEND_ORIGIN || '') + '/preview.jpeg';
 const SITE_TITLE = 'Unified Auctions';
 const SITE_DESCRIPTION = `The "${SITE_TITLE}" is the portal to all Maker Auctions related services. Easily interact with the Maker Protocol through streamlined interfaces, inform yourself about how the protocol works and receive updates on current auctions.`;
 const TWITTER_HANDLE = '@MakerDAO_SAS';
+// mode: https://v3.router.vuejs.org/api/#mode
+const ROUTER_HISTORY_MODES = ['history', 'hash', 'abstract'];
+// Nuxt has 'history' as default
+const HISTORY_MODE = ROUTER_HISTORY_MODES.includes(process.env.HISTORY_MODE) ? process.env.HISTORY_MODE : 'history';
+const BASE = HISTORY_MODE === 'hash' ? path.join(__dirname, 'dist') : '';
```
valiafetisov, line 12: Using `path.join(__dirname, 'dist')` will produce html file with hardcoded absolute path like `/Users/user/unified-auctions-ui/frontend/dist/favicon.ico`. This path will point to this specific folder on this specific computer and will not be distributable (eg you can not publish this file or send to someone or even move it to another folder). 

Instead, the only change you need to make in order to make it work for `file://` protocol is to make path relative instead of absolute

```suggestion
const BASE = HISTORY_MODE === 'hash' ? '.' : '';
```

valiafetisov: Since I've tested this already in order to suggest it, I've directly committed it via [3e1931d](https://github.com/sidestream-tech/unified-auctions-ui/pull/575/commits/3e1931d98df3c8da3df11c933f37abd4e09eaa3d)

valiafetisov (approved):

PR #579, opened 2023-01-10
Add and configure Electron to package app

### LINKED ISSUE #578: Add Electron and research how to integrate extensions
### Goal

Package the app with Electron and support all extensions.

### Context

We rely on browser extensions to support the wallets. After packaging the app with Electron, we need to find how to re-integrate them.

### Assets

Electron docs: https://www.electronjs.org/docs/latest/

### Tasks

- [x] Package the app with Electron
- [x] Research how to re-integrate extensions

aomafarag:
Closes #578.

Checklist:

Screenshot from 2023-01-10 15-30-39


frontend/main.js
```diff
@@ -0,0 +1,32 @@
...
+const serve = require('electron-serve');
+
+const loadURL = serve({ directory: 'dist' });
+
+let mainWindow;
+
+function createWindow() {
+    mainWindow = new BrowserWindow({
+        width: 800,
+        height: 600,
```
valiafetisov, line 11: Please set it to the monitor size

frontend/main.js
```diff
@@ -0,0 +1,32 @@
...
+    loadURL(mainWindow);
+}
+
+app.whenReady().then(() => {
+    createWindow();
+    app.on('activate', () => {
+        if (BrowserWindow.getAllWindows().length === 0) {
+            createWindow();
+        }
+    });
```
valiafetisov, line 25: Why is it needed to create window two times?

frontend/main.js
```diff
@@ -0,0 +1,32 @@
...
+});
+
+app.on('window-all-closed', () => {
+    if (process.platform !== 'darwin') {
+        app.quit();
+    }
```
valiafetisov, line 31: Why do you quit only it's not mac?

frontend/main.js
```diff
@@ -0,0 +1,32 @@
+const { app, BrowserWindow } = require('electron');
+const serve = require('electron-serve');
+
+const loadURL = serve({ directory: 'dist' });
```
valiafetisov, line 4: Why did you decide to use web server instead of the basic file viewing?

frontend/package.json
```diff
@@ -75,6 +78,7 @@
     "babel-core": "7.0.0-bridge.0",
     "babel-jest": "^27.4.2",
     "babel-loader": "^8.2.2",
+    "electron": "^22.0.0",
```
valiafetisov, line 81: All new dependancies should at least be dev dependencies. But I would actually recommend to create another folder outside of the frontend, eg `/electron` to create its own package json and have its own place for the main.js file. Because there is no reason to keep them in one place

valiafetisov (commented):
It's nice that you were able to start the project!

But remember, the ultimate goal of this PR should be packaging. Ie, a command that outputs applications suitable to run on 3 different platforms: windows, mac os and ubuntu.

frontend/main.js
valiafetisov: When you apply some hacks (even if they are suggested by the documentation), but are not clear by default, please try to explain it. In this case, please add a comment with the link

frontend/main.js
valiafetisov: Same here 🙂

electron/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "auctions-electron",
+  "version": "1.0.0",
+  "description": "auctions electron app packaging",
+  "main": "main.js",
+  "scripts": {
+    "start": "electron ."
+  },
```
valiafetisov, line 8: As discussed, please add build command that when executed do all necessary steps from building frontend with correct env vars to outputing `electron/distr` folder with built and ready-to-distribute apps

frontend/package-lock.json
```diff
@@ -12847,9 +12847,9 @@
       "dev": true
     },
     "node_modules/@types/node": {
-      "version": "15.12.4",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-15.12.4.tgz",
-      "integrity": "sha512-zrNj1+yqYF4WskCMOHwN+w9iuD12+dGm0rQ35HLl9/Ouuq52cEtd0CH9qMgrdNmi5ejC1/V7vKEXYubB+65DkA=="
+      "version": "16.18.11",
```
valiafetisov, line 12850: Why did the `frontend/package-lock.json` changed? Please reset it

valiafetisov:
> From the project's root, run the following commands in order in a terminal:

- [ ] Please instead make a single `npm run start` command out of this, please also make sure to install required dependencies (eg via postinstall script)
- [ ] Please add a readme for this subfolder

valiafetisov:
Why [this comment is marked as resolved](https://github.com/sidestream-tech/unified-auctions-ui/pull/579#discussion_r1068154250), but there is no new commits? Have you worked on the script to produce packaged apps?

> There are 3 popular libraries for packaging/building Electrons apps

Why do you reference `Electron Forge` and `electron-packager` on the same level if `Electron Forge` [actually depends on `electron-packager`](https://www.electronforge.io/core-concepts/build-lifecycle#step-1-package)?

The decision which tool to use is up to you. Since you didn't provide any reasoning for or against any of them yet, I can't even help. Do you have trouble setting one of them up? Or you haven't tried any yet? So what do you wait for? I personally worked with `builder` before and it was ok.

> Regardless of the library, .dmg builds can only be done on macOS: https://stackoverflow.com/questions/11736368/how-to-sign-a-mac-os-x-application-in-linux/12156576#12156576.

The link says that you can't sign an app outside of the mac ox, but we don't need to sign the app. For the `builder`, see [How to Disable Code Signing During the Build Process on macOS](https://www.electron.build/code-signing.html#how-to-disable-code-signing-during-the-build-process-on-macos)

What about the Ci scripts that you've mentioned in the daily? _After you made it work on your local machine_ for the whatever single target you want (ie debian if you're on debian, windows if you're on windows), you can have a look into setting up github actions, eg https://github.com/marketplace/actions/electron-builder-action or https://medium.com/@johnjjung/building-an-electron-app-on-github-actions-windows-and-macos-53ab69703f7c. As far as I remember on mac os it was possible to build all 3 platforms.

electron/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "unified-auctions",
+  "version": "1.0.0",
+  "description": "MakerDAO Unified Auctions UI",
+  "author": "SIDESTREAM GmbH",
+  "main": "main.js",
+  "build": {
+    "appId": "com.MakerDAO.UnifiedAuctions",
```
valiafetisov, line 8: Let's use
```suggestion
    "appId": "com.makerdao.unified-auctions",
```

electron/package.json
```diff
@@ -0,0 +1,25 @@
...
+    "mac": {
+      "category": "public.app-category.finance"
+    }
+  },
+  "scripts": {
+    "preinstall": "cd ../frontend/ && npm ci && npm run build && npm run generate && cd ../electron/",
```
valiafetisov, line 14: Name `preinstall` is not just a name, it's a special script name that will be run every time after npm installation. Is it desired?

> These all run after the actual installation of modules into node_modules, in order, with no internal actions happening in between

https://docs.npmjs.com/cli/v9/using-npm/scripts#life-cycle-operation-order

electron/main.js
```diff
@@ -0,0 +1,37 @@
+const { app, BrowserWindow } = require('electron');
+const serve = require('electron-serve');
+
+const loadURL = serve({ directory: '../frontend/dist' });
```
valiafetisov, line 4: The path can be defined differently for different OSs. It's recommended to use `path` node package to do any cross-os operations with file path (eg `path.join()`)

valiafetisov (commented):
You can enable debugging on the client and on the server to figure out what's is happening with the "blank screen" window. Since it seems like a "common problem", not a "known problem"

frontend/nuxt.config.js
```diff
@@ -48,8 +49,8 @@ export default {
             { name: 'twitter:description', content: SITE_DESCRIPTION },
             { name: 'twitter:image', content: PREVIEW_IMAGE },
         ],
-        link: [{ rel: 'icon', type: 'image/x-icon', href: '/favicon.ico' }],
-        script: [{ src: '/js/HeapIO.js' }],
+        link: [{ rel: 'icon', type: 'image/x-icon', href: path.join(__dirname, 'dist/favicon.ico') }],
+        script: [{ src: path.join(__dirname, 'dist/js/HeapIO.js') }],
```
valiafetisov, line 53: I think `__dirname` is actually deprecated by node and you should use [`import.meta.url` instead](https://flaviocopes.com/fix-dirname-not-defined-es-module-scope/)

But why do you actually need to do that?

frontend/package-lock.json
valiafetisov: Can you please reset changes made to that file? This commend was not addressed (or it was changed again?)

electron/package.json
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "files": [
+      "main.js",
+      "dist/**/*"
+    ]
+  },
+  "scripts": {
+    "preinstall": "cd ../frontend && npm ci && npm run build && npm run generate",
+    "postinstall": "electron-builder install-app-deps",
+    "prepare": "npx copyfiles -a -u 2 \"../frontend/dist/**/*\" .",
```
valiafetisov, line 22: When is this command actually executed?

electron/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "start": "npm ci && electron .",
```
valiafetisov, line 23: Does `npm start` starts a development version of the nuxt or what's the point of that command?

electron/package.json
```diff
@@ -0,0 +1,32 @@
+{
+  "name": "unified-auctions",
```
valiafetisov, line 2:
```suggestion
  "name": "unified-auctions-app",
```

valiafetisov (commented):
- [ ] Where was [this suggestion](https://www.electron.build/code-signing.html#how-to-disable-code-signing-during-the-build-process-on-macos) applied? Have you applied and removed it later on? Do you have a commit with it?

electron/package.json
valiafetisov: You should not normally mix installation with running/building commands. Otherwise, why is it not a default behaviour? 🙂. Expected flow: developer cds into electron folder run `npm run install` only once to install every required dependency. Then, they should be able to run `npm run build` which will rebuild the frontend and build a new app

electron/package.json
valiafetisov: I wasn't aware that the app will be named after this field, so I changed it back

frontend/nuxt.config.js
valiafetisov: Please remove those changes and merge logic from #575

frontend/nuxt.config.js
```diff
@@ -108,4 +109,8 @@ export default {
             },
         },
     },
+    router: {
+        mode: 'hash',
+        base: path.join(__dirname, 'dist'),
+    },
```
valiafetisov, line 115: Please remove those changes too

frontend/nuxt.config.js
valiafetisov: Please remove those changes and merge logic from #575. Please note that you will need to provide correct env var for the npm run build to activate correct router mode

frontend/package-lock.json
valiafetisov: This was still not addressed!

frontend/nuxt.config.js
valiafetisov: Where do you provide correct `ENABLE_FILE_PROTOCOL` env variable to the frontend build command? Or do you expect the developer to change it in the env file?

valiafetisov:
Please merge main into this branch and validate that the whole application works end-to-end 🙂

LukSteib:
> There is a weird bug: when setting the RPC URL to our Infura ID, the Cannot verify RPC URL error is thrown. However, everything works perfectly with Alchemy.

Might that be due to the fact that our Infura ID is restricted by origin and once packaged as electron app we violate that origin restriction?

LukSteib (changes requested):
- [x] NIT: can we re-name the application to `Unified Auctions UI` as it already became our operating title
- [x] For the 3rd party resources on the landingpage, can we open them in browser and not inside the application? (e.g. makerburn)
   - [x] same for the links in the footer. Additionally here the Terms and Conditions page just shows blank page. This is probably due to us providing it as a pdf. Maybe @valiafetisov has a good suggestion how to go about this 

--- 
Other than that I was able to connect a wallet via WalletConnect and execute a tx on goerli. 
Basically I was able to reproduce all of the steps outlined in former WalletConnect PR (see [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/532#issuecomment-1318947281))

One thing I noticed: The confirmation of a certain tx takes quite long via our app. While confirmation is given already in wallet app (on mobile phone) the confirmation in our app takes ~30sek longer.

valiafetisov:
> For the 3rd party resources on the landingpage, can we open them in browser and not inside the application? (e.g. makerburn)

Good code snippet for that is https://stackoverflow.com/a/34503175 that can be used here. But as you can't directly insert this code snippet into our frontend, you should use something like `executeJavaScript` to inject this logic

> While confirmation is given already in wallet app (on mobile phone) the confirmation in our app takes ~30sek longer.

That's expected, since we wait for a few other transactions on top to be finished to make sure that it is correct confirmation

valiafetisov:
> For the 3rd party resources on the landingpage, can we open them in browser and not inside the application? (e.g. makerburn)

cc @aomafarag, implemented it via 2962806

valiafetisov:
Currently I'm seeing this after building the project, installing it and running:

Screenshot 2023-01-31 at 18 24 08

LukSteib:
> Currently I'm seeing this after building the project, installing it and running:

Might be related to the experience I just had: I've removed the old application (of first install). Based on latest changes I built, installed and ran the application. On startup I was not prompted with RPC popup but it already had the one I inserted on first run. 
Is this expected behavior?

valiafetisov:
> Might be related to the experience I just had: I've removed the old application (of first install)

Which application should I remove? dist? out? We [already remove `out` folder](https://github.com/sidestream-tech/unified-auctions-ui/blob/2f793bb1b970d33c627d66efe2e46c5413a31d39/electron/package.json#L24) before building starts. If something else needs to be removed, it should be specified in the code

> On startup I was not prompted with RPC popup but it already had the one I inserted on first run.
Is this expected behavior?

If it's a newly build dmg application I guess it should have it's own localstorage, in case of the `npm start` it might be the same application that also preserves the local storage data

valiafetisov:
- Removed `frontend/dist`, rebuilt, reinstalled, experienced the [same problem](https://github.com/sidestream-tech/unified-auctions-ui/pull/579#issuecomment-1410780323)
- Cleaned up local storage, closed, opened, experienced the same problem

KirillDogadin-std:
```bash

export RPC_URL = our url
npm run prebuild
npm start
```

LukSteib:
> should this work without the export of the RPC_URL env var?

Yes, every user should be able to insert their own RPC_URL on startup of the application.

KirillDogadin-std:
> Yes, every user should be able to insert their own RPC_URL on startup of the application.

so instead of the error popping up , the below attached pop up should appear, right?

image

LukSteib:
> so instead of the error popping up , the below attached pop up should appear, right?

Exactly

KirillDogadin-std:
1. For some reason the rpc url is set to localhost with no apparent reason
  - i did not find the reason for this. tried searching for localhost/127... in the src code (frontend, core)
  - tried to check my env vars
2. I suggest removing `ChangePageNetworkModal` from the default layout as a quick&dirty fix:
  - it pops up in case the network is not valid, we can just show the rpc url modal instead at all times (in case of error and in case of initiated switch)

valiafetisov:
> I suggest removing ChangePageNetworkModal from the default layout

How would is solve the problem? The user would still need to figure out how to change the network. Instead, how is the "provide new rpc url" modal now triggered? Should we instead drive it from the fact that no networks are there?

KirillDogadin-std:
> How would is solve the problem? 

> Instead, how is the "provide new rpc url" modal now triggered?

the condition for trigger of `ChangePageNetworkModal` goes to extend the condition of the rpc modal. so instead of the error the change network is shown.

> Should we instead drive it from the fact that no networks are there 🤪?

the root problem is that "no networks are there" does not happen and instead the `localhost` appears from god knows where. Ideally the mechanism that inserts localhost url to the rpc url variable should be removed, but i can't find it. what i found out from debugging is that the `preferences/getRpcUrl` is called twice. At first it returns nothing, but then immediately gets called again and at this point already returns localhost. Sadly, i did not see `setRpcUrl` action being called. So here is the confusion.

Quick & dirty fix is removing the error modal and just showing the rpc modal instead (on the same conditions)

valiafetisov:
Was actually able to build and run correct application. The problem was a combination of two things: 

- (dev setup issue) `npm run build` is actually using the `forntend/.env` so before building the app locally, they should be commented out
  - couldn't find a good solution for now. I guess it's not a problem when we would build app in the CI
- (logic issue) The rpc url is stored in localstorage and if you already had it stored there and it's invalid, it will be in this state when you can't do anything and invalid error is shown
  - we should actually make sure that logically this kind of limbo doesn't exist: if rpc url is invalid, we should show rpc dialog
    - but because it's tricky (eg to distinguish network error from origin protection), I went with another fix
  - I now store rpc url under a different key for every new application build. For the user, that means that whenever they download a new app, they also have to set rpc url again (which I think is expected!)

LukSteib (approved):
Worked for me as well. Approving this PR. 

Just to not forget and tackle in follow up issue: 
- [ ] `Terms and Conditions` and `Report a problem` links in footer are not working properly

valiafetisov:
Screenshot for the update notes
Screenshot 2023-02-02 at 14 58 18

LukSteib (approved):

PR #587, opened 2023-02-08
One inch integration

KirillDogadin-std:
Closes # .

Checklist:

### DIFF
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -347,7 +347,14 @@ export const bidWithCallee = async function (
     const calleeAddress = getCalleeAddressByCollateralType(network, auction.collateralType, marketId);
     const marketData = auction.marketDataRecords?.[marketId];
     const preloadedPools = marketData && 'pools' in marketData ? marketData.pools : undefined;
-    const calleeData = await getCalleeData(network, auction.collateralType, marketId, profitAddress, preloadedPools);
+    const oneInchData = marketData && 'oneInch' in marketData ? marketData.oneInch : undefined;
+    if ([!!oneInchData, !!preloadedPools].filter(i => i).length > 1) {
+        throw new Error('Only one of oneInchData and preloadedPools can be provided');
+    }
+    const calleeData = await getCalleeData(network, auction.collateralType, marketId, profitAddress, {
+        pools: preloadedPools,
+        oneInchParams: oneInchData ? { txData: oneInchData.calleeData, to: oneInchData.to } : undefined,
+    });
     const contractName = getClipperNameByCollateralType(auction.collateralType);
     const contractParameters = [
         convertNumberTo32Bytes(auction.index),
--- a/core/src/calleeFunctions/OneInchCallee.ts
+++ b/core/src/calleeFunctions/OneInchCallee.ts
@@ -2,36 +2,35 @@ import type { CalleeFunctions, CollateralConfig, Pool } from '../types';
 import { ethers } from 'ethers';
 import BigNumber from '../bignumber';
 import { getContractAddressByName, getJoinNameByCollateralType } from '../contracts';
-import { convertCollateralToDaiUsingPool, encodePools } from './helpers/uniswapV3';
-import { getPools } from '.';
+import { convertCollateralToDaiUsingPool } from './helpers/uniswapV3';
 import { routeToPool } from './helpers/pools';
 import { getRouteAndGasQuote } from './helpers/uniswapV3';
+import { getOneInchQuote } from './helpers/oneInch';
 
 const getCalleeData = async function (
     network: string,
     collateral: CollateralConfig,
     marketId: string,
     profitAddress: string,
-    params?: { pools?: Pool[]; oneInchParams?: {txData: string; to: string} }
+    params: { pools?: Pool[]; oneInchParams?: {txData: string; to: string} }
 ): Promise<string> {
     const marketData = collateral.exchanges[marketId];
-    if (marketData?.callee !== 'UniswapV3Callee') {
+    if (marketData?.callee !== 'OneInchCallee') {
         throw new Error(`getCalleeData called with invalid collateral type "${collateral.ilk}"`);
     }
-    const pools = params?.pools || (await getPools(network, collateral, marketId));
-    if (!pools) {
-        throw new Error(`getCalleeData called with invalid pools`);
+    if (!params.oneInchParams) {
+        throw new Error(`getCalleeData called with invalid txData`);
     }
     const joinAdapterAddress = await getContractAddressByName(network, getJoinNameByCollateralType(collateral.ilk));
     const minProfit = 1;
-    const uniswapV3pools = await encodePools(network, pools);
-    const typesArray = ['address', 'address', 'uint256', 'bytes', 'address'];
+    const typesArray = ['address', 'address', 'uint256', 'address', 'address', 'bytes'];
     return ethers.utils.defaultAbiCoder.encode(typesArray, [
         profitAddress,
         joinAdapterAddress,
         minProfit,
-        uniswapV3pools,
         ethers.constants.AddressZero,
+        params.oneInchParams.to,
+        params.oneInchParams.txData
     ]);
 };
 
@@ -42,21 +41,10 @@ const getMarketPrice = async function (
     collateralAmount: BigNumber
 ): Promise<BigNumber> {
     // convert collateral into DAI
-    const { route, fees } = await getRouteAndGasQuote(network, collateral.symbol, collateralAmount, marketId);
-    if (!route) {
-        throw new Error(`No route found for ${collateral.symbol} to DAI`);
-    }
-    const pools = await routeToPool(network, route, collateral.symbol, fees);
-    const daiAmount = await convertCollateralToDaiUsingPool(
-        network,
-        collateral.symbol,
-        marketId,
-        collateralAmount,
-        pools
-    );
+    const { tokenOut } = await getOneInchQuote(network, collateral.symbol, collateralAmount.toFixed(), marketId);
 
     // return price per unit
-    return daiAmount.dividedBy(collateralAmount);
+    return new BigNumber(tokenOut).dividedBy(collateralAmount);
 };
 
 const UniswapV2CalleeDai: CalleeFunctions = {
--- a/core/src/calleeFunctions/OneInchCallee.ts
+++ b/core/src/calleeFunctions/OneInchCallee.ts
@@ -2,9 +2,6 @@ import type { CalleeFunctions, CollateralConfig, Pool } from '../types';
 import { ethers } from 'ethers';
 import BigNumber from '../bignumber';
 import { getContractAddressByName, getJoinNameByCollateralType } from '../contracts';
-import { convertCollateralToDaiUsingPool } from './helpers/uniswapV3';
-import { routeToPool } from './helpers/pools';
-import { getRouteAndGasQuote } from './helpers/uniswapV3';
 import { getOneInchQuote } from './helpers/oneInch';
 
 const getCalleeData = async function (
@@ -12,13 +9,13 @@ const getCalleeData = async function (
     collateral: CollateralConfig,
     marketId: string,
     profitAddress: string,
-    params: { pools?: Pool[]; oneInchParams?: {txData: string; to: string} }
+    params?: { pools?: Pool[]; oneInchParams?: {txData: string; to: string} }
 ): Promise<string> {
     const marketData = collateral.exchanges[marketId];
     if (marketData?.callee !== 'OneInchCallee') {
         throw new Error(`getCalleeData called with invalid collateral type "${collateral.ilk}"`);
     }
-    if (!params.oneInchParams) {
+    if (!params || !params.oneInchParams) {
         throw new Error(`getCalleeData called with invalid txData`);
     }
     const joinAdapterAddress = await getContractAddressByName(network, getJoinNameByCollateralType(collateral.ilk));
--- a/core/src/calleeFunctions/helpers/oneInch.ts
+++ b/core/src/calleeFunctions/helpers/oneInch.ts
@@ -1,8 +1,12 @@
 import { ethers } from 'ethers';
-import { getContractSymbolByAddress } from '../../contracts';
+import { getCalleeAddressByCollateralType } from '../../constants/CALLEES';
+import { getCollateralConfigBySymbol } from '../../constants/COLLATERALS';
+import { getContractAddressByName, getContractSymbolByAddress } from '../../contracts';
+import { getChainIdByNetworkType } from '../../network';
 import { Pool } from '../../types';
 
 const EXPECTED_SIGNATURE = '0x12aa3caf'; // see https://www.4byte.directory/signatures/?bytes4_signature=0x12aa3caf
+const CALLEE_ADDRESS = '0x0'; // TODO: fill in
 
 export const getOneInchUrl = (chainId: number) => {
     return `https://api.1inch.io/v5.0/${chainId}`;
@@ -15,6 +19,28 @@ interface Protocol {
     img_color: string;
 }
 
+interface OneInchToken {
+    symbol: string;
+    name: string;
+    address: string;
+    decimals: 0;
+    logoURI: string;
+}
+interface OneInchSwapRepsonse {
+    fromToken: OneInchToken;
+    toToken: OneInchToken;
+    toTokenAmount: string;
+    fromTokenAmount: string;
+    protocols: OneInchSwapRoute[];
+    tx: {
+        from: string;
+        to: string;
+        data: string;
+        value: string;
+        gasPrice: string;
+        gas: string;
+    };
+}
 interface LiquiditySourcesResponse {
     protocols: Protocol[];
 }
@@ -42,18 +68,22 @@ export async function getOneinchValidProtocols(chainId: number) {
 }
 
 export async function getOneinchSwapParameters(
-    chainId: number,
-    fromTokenAddress: string,
-    toTokenAddress: string,
-    calleeAddress: string,
+    network: string,
+    collateralSymbol: string,
     amount: string,
-    slippage: string
-) {
+    slippage: string = '10'
+): Promise<OneInchSwapRepsonse> {
+    const chainId = parseInt(getChainIdByNetworkType(network) || '', 16);
+    if (Number.isNaN(chainId)) {
+        throw new Error(`Invalid chainId: ${chainId}`);
+    }
+    const toTokenAddress = await getContractAddressByName(network, 'DAI');
+    const fromTokenAddress = await getContractAddressByName(network, collateralSymbol);
     // Documentation https://docs.1inch.io/docs/aggregation-protocol/api/swap-params/
     const swapParams = {
         fromTokenAddress,
         toTokenAddress,
-        fromAddress: calleeAddress,
+        fromAddress: CALLEE_ADDRESS,
         amount,
         slippage,
         allowPartialFill: false, // disable partial fill
@@ -70,44 +100,46 @@ export async function getOneinchSwapParameters(
     return oneinchResponse;
 }
 
-export async function extractPoolsFromSwapResponseProtocols(
+export async function extractPathFromSwapResponseProtocols(
     network: string,
     oneInchRoutes: OneInchSwapRoute
-): Promise<Pool[]> {
-    return await Promise.all(
-        oneInchRoutes.map(async route => ({
-            routes: await Promise.all([
-                getContractSymbolByAddress(network, route.fromTokenAddress),
-                getContractSymbolByAddress(network, route.toTokenAddress),
-            ]),
-            fee: 3000,
-            addresses: [route.fromTokenAddress, route.toTokenAddress],
-        }))
+): Promise<string[]> {
+    const pathSteps = await Promise.all(
+        oneInchRoutes.map(
+            async route =>
+                await Promise.all([
+                    getContractSymbolByAddress(network, route.fromTokenAddress),
+                    getContractSymbolByAddress(network, route.toTokenAddress),
+                ])
+        )
     );
+    const path = [pathSteps[0][0]];
+    for (const step of pathSteps) {
+        path.push(step[1]);
+    }
+    return path;
 }
 
-export const fetchAutoRouteInformation = async function () {
-    // 0. get autoRouteData
-    // 1. extract route from response
-    // 2. trim route if needed
-    // 3. feel fees in
-    // 4. derive quote from input amount
-    // 5. derive quote gas adjusted from input amount, gas price and gas
-    // ( can have estimated gas from the api)
-};
-
-export const encodePools = async function (_network: string, pools: Pool[]): Promise<string> {
-    const types = [] as string[];
-    const values = [] as Array<string | number>;
-
-    for (const pool of pools) {
-        types.push('address');
-        values.push(pool.addresses[0]);
-
-        types.push('uint24');
-        values.push(pool.fee);
+export async function getOneInchQuote(network: string, collateralSymbol: string, amount: string, marketId: string) {
+    const chainId = parseInt(getChainIdByNetworkType(network) || '', 16);
+    if (Number.isNaN(chainId)) {
+        throw new Error(`Invalid chainId: ${chainId}`);
     }
-    types.push('address');
-    values.push(pools[pools.length - 1].addresses[1]);
-    return ethers.utils.solidityPack(types, values);
-};
+    const toTokenAddress = await getContractAddressByName(network, 'DAI');
+    const fromTokenAddress = await getContractAddressByName(network, collateralSymbol);
+    const calleeAddress = getCalleeAddressByCollateralType(network, getCollateralConfigBySymbol(collateralSymbol).ilk, marketId);
+    // Documentation https://docs.1inch.io/docs/aggregation-protocol/api/swap-params/
+    const swapParams = {
+        fromTokenAddress,
+        toTokenAddress,
+        fromAddress: calleeAddress,
+        amount,
+        protocols: (await getOneinchValidProtocols(chainId)).join(','),
+    };
+    const oneinchResponse = await executeOneInchApiRequest(chainId, '/quote', swapParams);
+    return {
+        estimatedGas: oneinchResponse?.estimatedGas,
+        route: await extractPathFromSwapResponseProtocols(network, oneinchResponse?.protocols),
+        tokenOut: oneinchResponse?.toTokenAmount,
+    };
+}
--- a/core/src/calleeFunctions/helpers/oneInch.ts
+++ b/core/src/calleeFunctions/helpers/oneInch.ts
@@ -1,13 +1,12 @@
 import { ethers } from 'ethers';
+import { getContractSymbolByName } from '../../contracts';
 import { Pool } from '../../types';
 
 const EXPECTED_SIGNATURE = '0x12aa3caf'; // see https://www.4byte.directory/signatures/?bytes4_signature=0x12aa3caf
 
-export const getOneInchUrl = (
-    chainId: number,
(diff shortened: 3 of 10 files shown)


LukSteib:
Tested with an `ETH-A` auction based on hardhat fork of latest block.

Documentation of findings during review:
* Was able to follow the guidelines above
* One Inch price resulted in most competitive price (see s1 below)
* Was able to execute auction with One Inch callee and received DAI profit
  * Related to known profit calc problem: The net profit indicator was not 100% accurate but quite ok (see s2 below)



s1: 
image

s2:
image

core/src/auctions.ts
```diff
@@ -347,7 +347,11 @@ export const bidWithCallee = async function (
     const calleeAddress = getCalleeAddressByCollateralType(network, auction.collateralType, marketId);
     const marketData = auction.marketDataRecords?.[marketId];
     const preloadedPools = marketData && 'pools' in marketData ? marketData.pools : undefined;
-    const calleeData = await getCalleeData(network, auction.collateralType, marketId, profitAddress, preloadedPools);
+    const oneInchData = marketData && 'oneInch' in marketData ? marketData.oneInch : undefined;
+    const calleeData = await getCalleeData(network, auction.collateralType, marketId, profitAddress, {
+        pools: preloadedPools,
+        oneInchParams: oneInchData ? { txData: oneInchData.calleeData, to: oneInchData.to } : undefined,
+    });
```
valiafetisov, line 354: The original proposal was:
- Remove `getCalleeData` from `bidWithCallee` function
- Execute `getCalleeData` inside `getMarketPrice` and store generated binary string (`calleeData`) inside `marketDataRecords[marketId]`, uniformly between callee types (ie, for all kind of callees)
  - The only additional value to replace `pools` that we will need to keep inside the `marketData` object will be `exchangePath` – I suggested to store generated string (like `LINK -> ETH -> DAI`)
  - This will remove the need for `pools` whatsoever
- Use pre-generated `calleeData` inside `bidWithCallee`

Please reach out to me/comment here if you didn't fully get the proposal

core/src/calleeFunctions/OneInchCallee.ts
```diff
@@ -0,0 +1,55 @@
...
+    collateral: CollateralConfig,
+    marketId: string,
+    collateralAmount: BigNumber
+): Promise<{ price: BigNumber; pools: undefined }> {
+    // convert collateral into DAI
+    const collateralAmountWad = collateralAmount.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0);
+    const { tokenOut } = await getOneInchQuote(network, collateral.symbol, collateralAmountWad, marketId);
+
+    // return price per unit
+    return { price: new BigNumber(tokenOut).dividedBy(collateralAmountWad), pools: undefined };
```
valiafetisov, line 47:
```diff
-    return { price: new BigNumber(tokenOut).dividedBy(collateralAmountWad), pools: undefined };
+    return { price: new BigNumber(tokenOut).dividedBy(collateralAmountWad), calleeData: await getCalleeData(/* params */) };
```

core/src/calleeFunctions/OneInchCallee.ts
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const UniswapV2CalleeDai: CalleeFunctions = {
+    getCalleeData,
```
valiafetisov, line 51: This means, we no longer need to even export it
```suggestion
```

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,197 @@
...
+        to: string;
+        data: string;
+        value: string;
+        gasPrice: string;
+        gas: string;
+    };
+}
+interface LiquiditySourcesResponse {
+    protocols: Protocol[];
+}
```
valiafetisov, line 48: Nit: there is no much sense to type http response if you don't validate received data (except maybe for code completion in the editor)

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,197 @@
...
+    network: string,
+    collateralSymbol: string,
+    amount: string,
+    marketId: string,
+    slippage = '10'
+): Promise<OneInchSwapRepsonse> {
+    let chainId = parseInt(getChainIdByNetworkType(network) || '', 16);
+    if (chainId === 1337) {
+        chainId = 1;
+    }
```
valiafetisov, line 82: You can use `isFork` to set it to main, that's at least a little bit less hacky

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,197 @@
...
+    const estimatedGas = new BigNumber(
+        (
+            await getOneInchQuote(
+                network,
+                collateral.symbol,
+                amount.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0),
+                marketId
+            )
+        ).estimatedGas
+    );
```
valiafetisov, line 186: Please don't do another request to 1inch just to estimate gas of the transaction. `swapData.tx.gasPrice` already contains enough data to calculate transaction cost

core/src/calleeFunctions/OneInchCallee.ts
```diff
@@ -0,0 +1,55 @@
...
+
+const getMarketPrice = async function (
+    network: string,
+    collateral: CollateralConfig,
+    marketId: string,
+    collateralAmount: BigNumber
+): Promise<{ price: BigNumber; pools: undefined }> {
+    // convert collateral into DAI
+    const collateralAmountWad = collateralAmount.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0);
+    const { tokenOut } = await getOneInchQuote(network, collateral.symbol, collateralAmountWad, marketId);
```
valiafetisov, line 44: To make it clear: there should be **only a singe API request** to 1inch per auction. Returned swap data already contains [`toTokenAmount`](https://docs.1inch.io/docs/aggregation-protocol/api/swap-params#description-of-response-parameters) value

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,197 @@
...
+        })
+    );
+    const path = [pathStepsResolves[0][0]];
+    for (const step of pathStepsResolves) {
+        path.push(step[1]);
+    }
+    return path;
+}
+
+export async function getOneInchQuote(network: string, collateralSymbol: string, amount: string, marketId: string) {
```
valiafetisov, line 131: Please remove this function

core/src/auctions.ts
KirillDogadin-std: > Execute getCalleeData inside getMarketPrice

then the `getMarketPrice` in this case does performs actions that do not correspond to its name which is e.g. problematic for `convertETHtoDAI` and/or all other functions (in `debt.ts`, `fees.ts`, etc) that only need the market price calculated.

> and store generated binary string (calleeData) inside marketDataRecords[marketId]

currently this function uses `getMarketDataRecords` call to receive this object (some of underlying functionality there is cached). So there is not object that is actually stored in the variable. So according to this proposal, i would have to instantiate such object, pass it around and manage its state which breaks the initial functional-procedural design paradigm.
  - E.g. i would have to return it from the `getMarketPrice` with all the `getCalleeData`s being called for every market (not only for the best one). 
  - After returning this object, i would have to propagate it upwards (or only the callee datas) to the enrichment functionality

If i interpret the consequences of the proposed approach correctly, then i do not see how the suggested here implementation is better than the one currently featured in the PR.

KirillDogadin-std: the callee data generation requires the profit address. Should i fill this in and reencode this on the stage of bidding?

core/src/calleeFunctions/OneInchCallee.ts
KirillDogadin-std: wip in https://github.com/sidestream-tech/unified-auctions-ui/pull/587#discussion_r1113282539

core/src/auctions.ts
valiafetisov: True! But can you reencode it at bidding uniformly for all callees? (I think different callees have different order of arguments for that, right?) What would be another workaround?

KirillDogadin-std: https://github.com/makerdao/exchange-callees/tree/master/src

all the callees there as of now comply with the position of profitAddress (i.e. first) so unpacking the data into address and bytes and packing it back should work up until we intend supporting a callee that violates this order.

___

i do not see the alternative workaround that is scoped to the core ( mb it's possible to also adjust store on frontend so it sends in the wallet address for querying the marketdata of the auctions, but this does not make sense because wallet should not be required for this)

KirillDogadin-std: additional complications which seem to require more than just minor adjustments.

1. To implement the idea here, it would require to call the 1inch api as part of the `getCalleeData` for getting the `txData`. But this request requires the amount of exchanged token for the api. So we have the case where _generally_ unrelated information is to be passed to `getCalleeData`. And this edgecase is to be handled based on types. Just like now. So i dont think we're making something better in this scenario
2. Pools are provided in some cases by the market configuration, in other by the Alpharouter - along with the market price, there can be preloaded pools, there can be none. Removing the pools (if i got the idea correctly) from the scope of `getMarketPrice` also then would require additional logic put in every function to map this. And then changing the whole chain of `pools` transformations that propagates to frontend.

These both along with the mentioned `profitAddress` above seem to require expanding the functionality of the `getMarketPrice` and `getCalleeData` and these would have to account for a variety of cases/inputs. This is already the case, just in another form and less hacks that fill in a dummy value to replace it later.

I would suggest to leave the solution as it is now (given that i understood your intent correctly and am seeing some consequences that you did not account), then if there're code structure problems - open a refactoring issue, list the points that are problematic, discuss them, come up with the solutions that would work out and are less hacky. Currently i'm not seeing the benefit of the proposed approach, i also did not find the comment that highlights the imperfections of current implementation.

KirillDogadin-std:
Goerli is apparently not supported by 1inch so testing it on this network will not work out.

see tabs at https://docs.1inch.io/docs/aggregation-protocol/api/swagger/

LukSteib:
Ran through the whole process described [here](https://github.com/sidestream-tech/unified-auctions-ui/pull/587#issuecomment-1430067216) again and was able to execute an `ETH-A` auction on local fork with 1inchCallee

valiafetisov:
> Goerli is apparently not supported by 1inch so testing it on this network will not work out.

So, how will it be handled in production, when the user switches to Goerli network? What will they see?

I imagine that the user should see the 1inch market with error/note in the list of markets (so only when they expand markets). I assume that this can be implemented using existing logic of handling errored markets (so without new custom 1inch logic). I also imagine that we would want to check network id before/instead of sending requests to 1inch (to avoiding reaching the quota).

KirillDogadin-std:
image

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+    };
+}
+interface LiquiditySourcesResponse {
+    protocols: Protocol[];
+}
+type OneInchSwapRoute = { name: string; part: number; fromTokenAddress: string; toTokenAddress: string }[][];
+
+export const executeOneInchApiRequest = async (
+    chainId: number,
+    endpoint: '/swap' | '/quote' | '/liquidity-sources',
```
valiafetisov, line 56:
```suggestion
    endpoint: '/swap' | '/liquidity-sources',
```

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+    const url = `${oneInchUrl}${endpoint}?${new URLSearchParams(params)}`;
+    await REQUEST_LIMITER.removeTokens(1);
+    const response = await fetch(url).then(res => res.json());
+    if (response.error) {
+        throw new Error(`failed to receive response from oneinch: ${response.error}`);
+    }
+    return response;
+};
+
+export async function getOneinchValidProtocols(chainId: number) {
```
valiafetisov, line 69: Can you please cache this for 24h

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+
+export async function getOneinchSwapParameters(
+    network: string,
+    collateralSymbol: string,
+    amount: string,
+    marketId: string,
+    slippage = '10'
+): Promise<OneInchSwapRepsonse> {
+    const isFork = getNetworkConfigByType(network).isFork;
+    const chainId = isFork ? 1 : parseInt(getChainIdByNetworkType(network) || '', 16);
```
valiafetisov, line 84: There is `getDecimalChainIdByNetworkType` for this purpose
```ts
    const chainId = isFork ? 1 : getDecimalChainIdByNetworkType(network);
```

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+import { getChainIdByNetworkType, getNetworkConfigByType } from '../../network';
+import { CollateralConfig } from '../../types';
+import BigNumber from '../../bignumber';
+import { getTokenAddressByNetworkAndSymbol } from '../../tokens';
+import { WAD_NUMBER_OF_DIGITS } from '../../constants/UNITS';
+import { RateLimiter } from 'limiter';
+
+const REQUEST_LIMITER = new RateLimiter({ tokensPerInterval: 1, interval: 'second' });
+const EXPECTED_SIGNATURE = '0x12aa3caf'; // see https://www.4byte.directory/signatures/?bytes4_signature=0x12aa3caf
+const SUPPORTED_1INCH_NETWORK_IDS = [1, 56, 137, 10, 42161, 100, 43114];
```
valiafetisov, line 14: Where does this comes from? Please add a comment

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+    amount: string,
+    marketId: string,
+    slippage = '10'
+): Promise<OneInchSwapRepsonse> {
+    const isFork = getNetworkConfigByType(network).isFork;
+    const chainId = isFork ? 1 : parseInt(getChainIdByNetworkType(network) || '', 16);
+    if (!isFork && !SUPPORTED_1INCH_NETWORK_IDS.includes(chainId)) {
+        throw new Error(`1inch does not support network ${network}`);
+    }
+    if (Number.isNaN(chainId)) {
```
valiafetisov, line 88: When can this happen? Can you then please move this error to the `networks.ts` function

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+export async function getOneInchMarketData(
+    network: string,
+    collateral: CollateralConfig,
+    amount: BigNumber,
+    marketId: string
+) {
+    const swapData = await getOneinchSwapParameters(
+        network,
+        collateral.symbol,
+        amount.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0),
```
valiafetisov, line 145: Why WAD if you're passing the amount of collateral?

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+        marketId
+    );
+    const path = await extractPathFromSwapResponseProtocols(network, swapData.protocols);
+    const calleeData = swapData.tx.data;
+    const estimatedGas = swapData.tx.gas;
+    const exchangeFeeEth = new BigNumber(swapData.tx.gasPrice).multipliedBy(estimatedGas);
+    const exchangeFeeDai = new BigNumber(10).multipliedBy(exchangeFeeEth); // await convertETHtoDAI(network, exchangeFeeEth); TODO: uncomment
```
valiafetisov, line 152: Please use actual conversion

core/src/auctions.ts
valiafetisov: Agreed in the daily: let's drop this requirement from this PR

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,161 @@
...
+import { getCollateralConfigBySymbol } from '../../constants/COLLATERALS';
+import { getErc20SymbolByAddress } from '../../contracts';
+import { getChainIdByNetworkType, getNetworkConfigByType } from '../../network';
+import { CollateralConfig } from '../../types';
+import BigNumber from '../../bignumber';
+import { getTokenAddressByNetworkAndSymbol } from '../../tokens';
+import { WAD_NUMBER_OF_DIGITS } from '../../constants/UNITS';
+import { RateLimiter } from 'limiter';
+
+const REQUEST_LIMITER = new RateLimiter({ tokensPerInterval: 1, interval: 'second' });
```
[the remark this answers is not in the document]
KirillDogadin-std: > This will lead to the situation when as soon as we have more than 1 auction, only the first in the list get enriched with 1inch data, all others throw the error.

could you reference the source of this information?

because from the [source](https://github.com/jhurliman/node-rate-limiter/blob/9bbd3b036c33e974d51d7b40ddeee13e71bb69b0/src/RateLimiter.ts#L51) and description on npm i can't find any support of this claim. It only throws if we requrest more than max amount of tokens it seems. I.e. if we request 2 or more tokens to be deducted.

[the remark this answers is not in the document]
KirillDogadin-std: replaced with the suggested package.

core/src/calleeFunctions/helpers/oneInch.ts
KirillDogadin-std: i agree, i would love to have the response typed anyway for convenience

core/src/types.ts
```diff
@@ -407,6 +426,10 @@ export declare interface VaultTransactionLiquidated extends VaultBase {
     state: 'liquidated';
     pastLiquidations: LiquidationEvent[];
 }
+export declare type GetCalleeDataParams =
+    | { pools: Pool[] }
+    | { oneInchParams?: { txData: string; to: string } }
+    | Record<string, never>;
```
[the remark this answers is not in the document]
KirillDogadin-std: > What is Record<string, never>

recommended way (ts error itself told me) to type `{}`

adjusted this typing and the corresponding logic.

core/src/calleeFunctions/helpers/oneInch.ts
KirillDogadin-std: removed due to prev comment adjustment

core/src/calleeFunctions/helpers/oneInch.ts
KirillDogadin-std: i don't get the question.
see api https://docs.1inch.io/docs/aggregation-protocol/api/swap-params swap parameters explainer for `amount` field.

core/src/calleeFunctions/helpers/oneInch.ts
KirillDogadin-std: hehehehe, this will result into recursion actually. 

here's the call tree:

- `convertEthToDai`
  - `getMarketPrice`
    - `getMarketDataRecords`
      - `getMarketDataById`
        - `if OneInchCallee`: `getOneInchMarketData`
          - `convertEthToDai`
            - ...

KirillDogadin-std:
- so due to https://github.com/sidestream-tech/unified-auctions-ui/pull/587#discussion_r1127695139 it seems we should reevaluate the process of how the conversion happens in order to avoid the recursion.

Suggestion from my side:

  - go adjust the functions for market info extraction with additional argument (with default value): `onlyUsePredefinedRoutes`
  - Add the logic to the function `getMarketDataRecords` to skip the market data extraction if the callee config does not contain a `route` key.

This way:

  1. we call the `getMarketDataRecords` without skipping enabled
  2. it reaches the point where conversion from eth to dai is required
  3. the conversion is performed based on the market data records that are generated via calee configs with predefined routes.

core/src/calleeFunctions/helpers/oneInch.ts
valiafetisov: The collateral amount that you `.shiftedBy` here can have different different number of decimals depending on the collateral. WAD is always 18, but if your collateral is USDC it should be 6. We keep the amount of decimals inside `collateralConfig`
```suggestion
        amount.shiftedBy(collateral.decimals).toFixed(0),
```

core/src/calleeFunctions/helpers/oneInch.ts
valiafetisov: As discussed in the daily, the suggestion is to use `convertSymbolToDai` (from `uniswapV3` helpers) for gas conversions (since gas prices doesn't need to be exchanged using the best market)

You can even further optimise it
- By calling it without amount parameter like `convertETHtoDAI` does (to not send request to uniswap for each specific gas amount, but rather do it once for 1 ETH, and cache it for all other gas prices)
- Update `convertETHtoDAI` to use the same structure, avoiding `getMarketPrice`

valiafetisov:
> it seems we should reevaluate the process of how the conversion happens in order to avoid the recursion.

Thanks for discovering and raising it! Answered with another suggestion [above](https://github.com/sidestream-tech/unified-auctions-ui/pull/587#discussion_r1128142422)

core/src/calleeFunctions/helpers/oneInch.ts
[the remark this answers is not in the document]
valiafetisov: > could you reference the source of this information?

First example in the readme states that ["This call will throw if we request more than the maximum number of requests"](https://github.com/jhurliman/node-rate-limiter/blob/9bbd3b036c33e974d51d7b40ddeee13e71bb69b0/README.md?plain=1#L30)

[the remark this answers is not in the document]
KirillDogadin-std:
```js

const limiter = new RateLimiter(maxAmount);

for (const i of Array(maxAmount + 1).keys()) {
  await  limiter.removeTokens(1)
} // this works

await limiter.removeTokens(maxAmount + 1) // this throws
```

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -0,0 +1,174 @@
...
+import { getDecimalChainIdByNetworkType, getNetworkConfigByType } from '../../network';
+import { CollateralConfig } from '../../types';
+import BigNumber from '../../bignumber';
+import { getTokenAddressByNetworkAndSymbol } from '../../tokens';
+import { WAD_NUMBER_OF_DIGITS } from '../../constants/UNITS';
+import { Queue } from 'async-await-queue';
+import memoizee from 'memoizee';
+import { convertETHtoDAI } from '../../fees';
+
+const REQUEST_QUEUE = new Queue(1, 1000);
```
valiafetisov, line 14: - Nit: It's advised to put numbers into self-describing variables
- Let's use a number closer disclosed to their limit? Or did you have any problems with that?
```suggestion
const MAX_DELAY_BETWEEN_REQUESTS_MS = 600;
const REQUEST_QUEUE = new Queue(1, MAX_DELAY_BETWEEN_REQUESTS_MS);
```

core/src/calleeFunctions/helpers/oneInch.ts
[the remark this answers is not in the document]
valiafetisov: Not sure why do you bring this code snippet ,since it doesn't contradict the overall point: the package you've initially used throws, if limit is reached instead of waiting for the previous calls to finish. Thanks for switching to `async-await-queue` 🙂

KirillDogadin-std:
[here](https://github.com/sidestream-tech/unified-auctions-ui/pull/587/files#diff-4fdab105d1fea69252b610a9f416c159aa27064474cd3acddbb3ee3fd565e7e7R34) you can see for which collaterals i've enabled this. TLDR for all that already had autorouter. Lmk if this is incorrect call in your opinion

valiafetisov (approved):

LukSteib:
Live testing the 1inch calle on an active auction shows that something needs to be off with the integration.
To reproduce:
- mainnet block height `16847741`
- auction: `GUSD-A:5`

- [ ] The indicated price yielded by 1inch cannot be correct (see s1 below). Trying to execute the auction resulted in an execution reverted error.

Screenshot 2023-03-17 at 14 13 53

KirillDogadin-std:
test again

LukSteib:
- Just executed the live auction on mainnet based on this PR (see: https://etherscan.io/tx/0x51398396ddcfca37603a4a567fea689837337b3c8cf7bb5663290f95c6f8be51)
- Learnings:
   - The 1inch callee works as I was able to execute the tx via 1inch 🎊
   - The Gross Profit and Tx fees indicators (and net profit consequently) are pretty accurate, although not 100% as overall the tx ended in a loss of ~$14

--- 
@KirillDogadin-std: Do we need to apply the change you made in your last commit to the `OneInchCallee.ts` file, to other callees as well?

KirillDogadin-std:
> Do we need to apply the change you made in your last commit to the OneInchCallee.ts file, to other callees as well?

i dont think so

core/src/calleeFunctions/OneInchCallee.ts
```diff
@@ -40,16 +40,16 @@ const getMarketPrice = async function (
     collateralAmount: BigNumber
 ): Promise<{ price: BigNumber; pools: undefined }> {
     // convert collateral into DAI
-    const collateralAmountWad = collateralAmount.shiftedBy(WAD_NUMBER_OF_DIGITS).toFixed(0);
+    const collateralAmountDecimals = collateralAmount.shiftedBy(collateral.decimals).toFixed(0);
```
valiafetisov, line 43: This is typically called `*IntegerAmount` in our codebase, cause it holds something opposite of decimals

```diff
- const collateralAmountDecimals = collateralAmount.shiftedBy(collateral.decimals).toFixed(0);
+ const collateralIntegerAmount = collateralAmount.shiftedBy(collateral.decimals).toFixed(0);
```

core/src/calleeFunctions/helpers/oneInch.ts
```diff
@@ -94,7 +94,7 @@ export async function getOneinchSwapParameters(
     collateralSymbol: string,
     amount: string,
     marketId: string,
-    slippage = '10'
+    slippage = '1'
```
valiafetisov, line 97: Why was it set to 10 initially?

KirillDogadin-std: there was no evidence that slippage `1` does not work out and has to be replaced. It's probably just a typo, which i noticed and adjusted to the original script's value.

valiafetisov (approved):

LukSteib:
Rested on mainnet fork the live situation from last week: 
- mainnet block height `16847741`
- auction: `GUSD-A:5`

- [ ] When trying to execute the auction with 1inch callee selected I get the following errors:

```
MetaMask - RPC Error: Internal JSON-RPC error. 
{code: -32603, message: 'Internal JSON-RPC error.', data: {…}}
code: -32603 data: {code: -32603, message: 'Error: VM Exception while processing transaction: reverted with an unrecognized custom error', data: {…}}
message: "Internal JSON-RPC error."

```

```
auctions.ts?b334:274 Bidding error {code: -32603, message: 'Internal JSON-RPC error.', data: {…}, stack: '{\n  "code": -32603,\n  "message": "Internal JSON-RP…ogaeaoehlefnkodbefgpgknn/background-3.js:10:40024'}
```

KirillDogadin-std:
i did the same with `16848000` block and succeeded.

possible reason why your case fails:

- oneinch request is the source of information for swap
- oneinch request returns you the data given the _current_ market conditions. It might suggest you a path that was not profitable at the previous block height.
- if you use the non-profitable path/calldata for the oneinch smart contract: it will fail.

LukSteib (approved):

PR #588, opened 2023-02-09
Add caching to periodically called auction update function

KirillDogadin-std:
Closes # .

Checklist:

### DIFF
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -32,9 +32,6 @@ import { enrichAuctionWithTransactionFees, getApproximateTransactionFees, getDef
 import parseAuctionId from './helpers/parseAuctionId';
 import { EventFilter } from 'ethers';
 import getNetworkDate, { fetchDateByBlockNumber } from './date';
-import memoizee from 'memoizee';
-
-const MARKET_DATA_CACHE_TIME = 29 * 1000;
 
 const enrichAuctionWithActualNumbers = async function (
     network: string,
@@ -213,7 +210,7 @@ export const enrichAuctionWithPriceDrop = async function (auction: Auction): Pro
     }
 };
 
-const _enrichAuctionWithPriceDropAndMarketDataRecords = async function (
+const enrichAuctionWithPriceDropAndMarketDataRecords = async function (
     auction: Auction,
     network: string,
     useAutoRouter?: boolean
@@ -223,15 +220,6 @@ const _enrichAuctionWithPriceDropAndMarketDataRecords = async function (
     const auctionWithFees = await enrichAuctionWithTransactionFees(enrichedAuctionWithNewPriceDrop, fees, network);
     return await enrichAuctionWithMarketDataRecords(auctionWithFees, network, useAutoRouter);
 };
-
-export const enrichAuctionWithPriceDropAndMarketDataRecords = memoizee(
-    _enrichAuctionWithPriceDropAndMarketDataRecords,
-    {
-        maxAge: MARKET_DATA_CACHE_TIME,
-        promise: true,
-        length: 3,
-    }
-);
 
 export const fetchSingleAuctionById = async function (
     network: string,
--- a/core/src/auctions.ts
+++ b/core/src/auctions.ts
@@ -32,6 +32,9 @@ import { enrichAuctionWithTransactionFees, getApproximateTransactionFees, getDef
 import parseAuctionId from './helpers/parseAuctionId';
 import { EventFilter } from 'ethers';
 import getNetworkDate, { fetchDateByBlockNumber } from './date';
+import memoizee from 'memoizee';
+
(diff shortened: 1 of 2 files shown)


KirillDogadin-std:
I would cache it on high level, lmk if you would like to break this down for some reason.

core/src/auctions.ts
```diff
@@ -32,6 +32,9 @@ import { enrichAuctionWithTransactionFees, getApproximateTransactionFees, getDef
 import parseAuctionId from './helpers/parseAuctionId';
 import { EventFilter } from 'ethers';
 import getNetworkDate, { fetchDateByBlockNumber } from './date';
+import memoizee from 'memoizee';
+
+const MARKET_DATA_CACHE_TIME = 10 * 1000;
```
valiafetisov, line 37:
```suggestion
const MARKET_DATA_CACHE_TIME = 29 * 1000;
```

core/src/auctions.ts
```diff
@@ -221,6 +224,15 @@ export const enrichAuctionWithPriceDropAndMarketDataRecords = async function (
...
 };
 
+export const enrichAuctionWithPriceDropAndMarketDataRecords = memoizee(
+    _enrichAuctionWithPriceDropAndMarketDataRecords,
+    {
+        maxAge: MARKET_DATA_CACHE_TIME,
+        promise: true,
+        length: 3,
+    }
+);
```
valiafetisov, line 234: Why `_getMarketPrice` doesn't work? Why did you have to add another layer of caching here? How does caching of the `auction` work here? By the object values (serialised) or by the object reference (so no matter if the values change or not)?

KirillDogadin-std: > Why _getMarketPrice doesn't work? 

essentially it narrows down to having `getMarketDataRecords` in the calltree of `enrichAuctionWithPriceDropAndMarketDataRecords` which is responsible for firing a lot of requests to update the info.

> How does caching of the auction work here? By the object values (serialised) or by the object reference (so no matter if the values change or not)?

i sense a rhetorical question that is designed to send me on a journey of past memories. I recall a claim about object reference being memorized and changes to the object not being tracked.

with both points above it seems like caching `getMarketDataRecords` instead would be a better solution that the current one.

valiafetisov (commented):
I would suggest instead of caching `getMarketDataRecords` (which I believe still contains some values that are calculated every second) we would actually fix `getMarketPrice` using `normalizer` (which should properly take care of the BigNumber)

core/src/calleeFunctions/index.ts
```diff
@@ -151,6 +152,15 @@ export const getMarketDataRecords = async function (
     return marketDataRecords;
 };
 
+export const getMarketDataRecords = memoizee(_getMarketDataRecords, {
+    promise: true,
+    maxAge: MARKET_DATA_RECORDS_CACHE_MS,
+    length: 3,
```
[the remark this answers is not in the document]
KirillDogadin-std: noted, but does not make sense to adjust it now since i would like first to resolve the review-comment: https://github.com/sidestream-tech/unified-auctions-ui/pull/588#pullrequestreview-1298029401

KirillDogadin-std:
> I would suggest instead of caching getMarketDataRecords (which I believe still contains some values that are calculated every second) we would actually fix getMarketPrice using normalizer (which should properly take care of the BigNumber)

i dont think this would solve the problem because there're 2 functions that use `getMarketDataRecords`:

  - `getMarketPrice`
  - `enrichAuctionWithMarketDataRecords`

and `enrichAuctionWithMarketDataRecords` is called every second. so if there's no caching on the lower level, the change will be ineffective.

valiafetisov:
Based on the discussion we had in the daily: you've indicated that `getMarketPrice` actually uses `getMarketDataRecords` under the hood. My bad, had a different intuition. To be precise I suggest to cache it on the lower level. In this case:

- Please cache `getMarketDataById` function (where `getMarketPrice` of the actual callee is executed)
  - Don't forget to use `normalizer`
  - Don't forget to use correct `length` number
- Let's then remove all other caching inside `calleeFunctions/index.ts`

Will this work? Can you please actual caching afterwards?

KirillDogadin-std:
> Will this work?

just by code estimation: it should

> If it will work and you will implement it,

implemented

> can please directly validate that cache is activated and there is no calls to the blockchain more often than every 30 seconds, but after 30 seconds the cache is invalidated?

validated by:

- introducing this also to one-inch callee pr:
  - i don't see requests to be fired there more than once per 30 seconds in _Network_ tab
  - the app does not get timed out by rate limit
- adding `console.log('bonk')` to the beginning of the function and not observing this log batch more than once per 30 seconds

valiafetisov (approved):
Code looks valid, waiting for the CI to pass

PR #629, opened 2025-07-10
Update UniswapV2LockstakeCallee address

valiafetisov:
This PR updates `UniswapV2LockstakeCallee` address. The callee had to be updated as the previous callee depended on the legacy `MKR_SKY` converter that no longer works


LukSteib (approved):
- Was able to simulate and participate in `UNIV2DAIUSDC-A` auction via UI ✔️ 
- Was able to simulate and participate in `LSE-MKR-A` auction via UI ✔️ 

Based on these two checks approving

PR #630, opened 2025-10-06
feat: onboard LSEV2-SKY-A

DaeunYoon:
Closes N/A

This PR onboards `LSEV2-SKY-A` and offboards `LSE-MKR-A`.
`LSEV2-SKY-A` implementation can be tested via simulation, UI, and bot.

Checklist:
- ~~[ ] issue number linked above after pound (`#`)~~
- ~~[ ] issue checkboxes are all addressed~~

image
image


LukSteib (changes requested):
Review checklist:
- Able to participate in standard `ETH-A` auction leveraging callee ✔️ 
- Able to execute new `LSEV2-SKY-A` auction leveraging callee ✔️ 

---
- Q: Does our icon pack not include proper SKY token icon?
- [ ] Since the callee relies on USDS instead of DAI I think we have to make this clearer in the UI
   - [x] On the first page of the auction we should rely on USDS indicators instead of DAI (see s1) similar to how we do it on the 2nd page (see s2)
   - [ ] We have the step to authorise DAI transactions (see s3) - everything works as expected in the end but are we really authorising DAI or doe we need to change that to USDS as well? 
- [ ] (we can put it out of scope for now but still mentioning it) AFAI understand the functionality to bid with DAI in this case leads to SKY directly being transferred to the wallet instead of being stored in the VAT. We should make this more obvious by hiding the withdrawal step for these kind of auctions (see s4)

s1
image

s2
image

s3
Screenshot 2025-10-08 at 12 15 31

s4
image

DaeunYoon:
> Q: Does our icon pack not include proper SKY token icon?

It doesn't look like the library we use has the correct sky icon. (on library [sky icon](https://github.com/man15h/vue-cryptoicon/blob/9cf0ac5c73f6933520253699a34fab70074ec8d8/src/icons/sky.js) was added 4 ys ago / lib last updated 2 ys ago) Do you want me to render sky token icon separately?


>  We have the step to authorise DAI transactions (see s3) - everything works as expected in the end but are we really authorising DAI or doe we need to change that to USDS as well?

This authorizes DAI so should be fine to keep imo.

>  Since the callee relies on USDS instead of DAI I think we have to make this clearer in the UI

Updated (s1)

For debt unit, i didn't update `DAI` units on `Bid with DAI` subtab (s2)

s1.
image


s2.
image

LukSteib:
> Do you want me to render sky token icon separately?

- [x] Yes I think it makes sense as current cloud icon might be missleading

> This authorizes DAI so should be fine to keep imo.

Ok

> For debt unit, i didn't update DAI units on Bid with DAI subtab (s2)

Yes that's fine. There denomination in DAI makes sense

DaeunYoon:
>  Yes I think it makes sense as current cloud icon might be missleading

image

LukSteib (approved):

